PR #4, opened 2016-06-21
Fix bug that could prevent cubio from saving the princess

neikeq:
Just think of how many Toads weren't saved because of this...


akien-mga:
:open_mouth:

PR #20, opened 2016-09-16
VsPlatformer bugfix: bullets were unable to deal damage

kubecz3k:
Needed to change  VS "call instance" node, function parameter to `callv` (it was `call`)


akien-mga:
We really need a text-based format for VisualScript...

kubecz3k:
Yeah definitely guess it will come in post 3.0 times..

PR #29, opened 2016-10-07
Fix simple multiplayer demo for 3+ players

Faless:
Properly send newly connected dudes to old players.


networking/simple_multiplayer/gamestate.gd
```diff
@@ -62,6 +62,7 @@ remote func register_player(id, name):
 		rpc_id(id, "register_player", 1, player_name) # Send myself to new dude
 		for p_id in players: # Then, for each remote player
 			rpc_id(id, "register_player", p_id, players[p_id]) # Send player to new dude
+			rpc_id(p_id, "register_player", id, name) # Send new dude to player
```
akien-mga: So I did remove one call too much in 3e16c84 then... I don't remember what was my reasoning but I had come to the conclusion that the call would be redundant... Apparently I was wrong :D

PR #31, opened 2016-11-02
Free coin in 3D platformer

williamd1k0:
Hi, I noticed (showing collision shapes) that the coins are still on the scene after "take" animation.
So I connected the anim to use `queue_free()`.


akien-mga:
Free coins... \o/

PR #43, opened 2017-07-05
gui acts properly for window size change

kubecz3k:
UI anchoring and margins are set to support screen changes
screen_change_ui


kubecz3k:
also the scene is in scn so it probably wont be mergable if @reduz had done anything with that scene :P

PR #48, opened 2017-07-28
Fix observer yaw and pitch

pcvonz:
There were two issues with the observer script:

- Pitch and yaw were reversed (fixed by setting `r_pos` to `-event.relative`
- Yaw fixed by using the `rotate` method rather than setting transform directly. Setting the transform directly made the kinematic character rotate relative to the origin of the scene (I think). Would be nice get a clear reason why this wasn't working.


akien-mga:
Sorry for the delay. Merged manually as 8933484 to remove trailing space ;)

PR #51, opened 2017-08-07
Rigid character 3d platformer demo

kubecz3k:
Player code is ported from previous demos (2.1)


akien-mga:
Sorry for the delay. Is it still up to date with current physics in the master branch?

kubecz3k:
@akien-mga I guess it wont work because of api breakage (like fixed_process rename or others probably). Not sure how much sense its now to merge. Seen reduz had new models for platformer demo + it wont work after bullet merge. When reduz will update current demo with new assets, and we will have new physics I will do new pr.

akien-mga:
Ok, let's close this one for now then? Or do you prefer to keep it open as a reminder? I don't mind either way.

kubecz3k:
Yes I think we can close

PR #63, opened 2017-09-04
[2.1] 3D/platformer: Adjust Player's collision capsule

ISylvox:
- Issue:
Robot's head overlapped with the 'ceiling' when our robot entered a tunnel and then jumping. This happens because collision capsule for the player.scn (robot scene) does not fully cover its head.

- Expected:
The head should collide and never overlap with the ceiling.

g21_-_3dplatformer_-_fix_overlapped_head

- What the PR does:
Adjust collision capsule for player.scn to fully covering robot's head, so that the head can collide with the ceiling.

g21_-_3dplatformer_-_adjust_collisioncapsule


akien-mga:
I'm not sure to understand what's the change?

ISylvox:
The collision capsule did not cover the whole body, e.g the head. When entering a tunnel and the robot do jumping, the head overlapping with the 'ceiling'.

PR #70, opened 2017-09-30
updated viewport clearing

Valentactive:
This change is needed in order to make use of the new introduced
set_clear_mode() function that was introduced here:
https://github.com/godotengine/godot/commit/5653477a30ca36c59a0d9c8bcb4b3a27a0feca21#diff-b0e269beca0e7df95d035998c40ecb64  

I think the choosen enum value is ok since its needed only once anyway.

With the old clear() function we get a invalid call error and demo is not started:
viewport_clear_error_before

While we get no problem with this change afterwards: 
viewport_clear_error_after


Valentactive:
Update: made the changes also in the remaining Viewport demos that made use of the viewport.clear() function

TwistedTwigleg:
Looks good to me! 👍

Valentactive:
I updated also the formatting (especially intentations) according to: 
 http://docs.godotengine.org/en/latest/learning/scripting/gdscript/gdscript_styleguide.html.

PR #74, opened 2017-10-13
Remove non-static physics objects to match tutorial beginning.

jakimfett:
In the [tutorial documentation](http://docs.godotengine.org/en/latest/learning/step_by_step/instancing.html), the container scene does not have any spheres in it.


akien-mga:
I think this tutorial expects you to use the zip it links, where the base scene does not have the instanced scenes. The version in this repo is the "final" version of the demo, showing instanced scenes, so it's expected to be this way AFAIK.

akien-mga:
Closing as per the above, this demo is meant to be this way (final state).

PR #75, opened 2017-10-14
Ported over the plugin demos to Godot 3

TwistedTwigleg:
I ported over two of the plugin demos from Godot 2. The custom node and custom dock plugins didn't need any changes, but the custom importer didn't work so I'm in the process of remaking it.

Currently the importer isn't working. I'm not 100% sure on how to fix the various problems but I'll keep working on it, I'm just posting to reduce duplicate work.

If anyone who has more experience with editor plugins wants to help I'd be more than happy for the assistance as I'm fairly in the dark on how the import/export process works for plugins. I couldn't figure out how exporting a resource works so I made my own exporter but I think that's causing the issue. I'll have to look into it later.


TwistedTwigleg:
Okay, I got the custom importer working.

The problem turned out to be you cannot export custom resources from GDScript, which means you can only import already made resources. So instead of making a plugin that imports existing resources, I instead used JSON to save and load the custom material. I also made a editor dock so you can create, save, and load materials all from the editor.


___

Probably not the best long term solution since it really doesn't show how to use `EditorImportPlugin` and `EditorExportPlugin`. Until `EditorExportPlugin` and `ResourceSaver.save()` fully work from GDScript it's really the only way to save and load custom resources that I know of.

There may be some other way to save/load custom resources that I'm unaware of, but after playing around for a couple days, searching the Internet and poking my head around the source code, it was the best solution I could find.

vnen:
I don't have the time to review this right now, but my Tiled importer plugin was partially ported to 3.0 (https://github.com/vnen/godot-tiled-importer/tree/godot-3). I'm not sure if it still works, since things may have changed, but it might give you some insight.

TwistedTwigleg:
Neat, I’ll take a look. Thanks!

PR #78, opened 2017-10-26
Add logs folder to .gitignore

santouits:


akien-mga:
It's actually a bug that those logs are dumped in the local folder in the first place, see https://github.com/godotengine/godot/issues/12277
So I don't know if it's worth adding to the .gitignore, fixing the bug would be better :)

santouits:
Alright, but if you don't know it, this is another. It puts a logs folder in your game folder "res://" too. Open a project and run it and then see your game folder.

akien-mga:
I've disabled the creation of the `logs/log.txt` RotatedFileLogger for now, so this is not necessary.

PR #86, opened 2017-10-31
Code enhancements for the mono pong demo.

Hinsbart:
- Use pattern matching instead of check-and-cast.
- Rename constants to fit the usual c# code-style.


mono/monkey_pong/CeilingFloor.cs
```diff
@@ -8,9 +8,8 @@ public class CeilingFloor : Area2D
 
     public void OnAreaEntered(Area2D area)
     {
-        if (area is Ball)
+        if (area is Ball ball)
```
akien-mga: That's a pretty weird syntax :|

PR #89, opened 2017-11-03
Added a simple drawing demo (GD paint)

TwistedTwigleg:
I made a MS paint like example to help show how to use the drawing functions in CanvasItem, capturing and cropping the Viewport to get the final image, and saving the image to a png (which I actually didn't know was possible until I made this).

It has a pencil, eraser, rectangle shape, and circle shape. The pencil and eraser can use either a square shape or a circle shape. All of the code is (hopefully) neatly commented and easy to understand.

In theory this could be extended to allow extra features, like layers or unlimited undos, but I thought I'd leave it where so it (hopefully) serves as a decent starting point for anyone wanting to make a drawing program in Godot.


2d/gd_paint/PaintControl.gd
```diff
@@ -0,0 +1,261 @@
...
+	# Otherwise we're removing a either a pencil stroke or a eraser stroke.
+	else:
+		# Figure out how many elements/brushes we've added in the last stroke
+		var elements_to_remove = draw_elements_list.size() - undo_element_list_num
+		# Remove all of the elements we've added this in the last stroke
+		for elment_num in range(0, elements_to_remove):
+			draw_elements_list.pop_back()
+
+		# Now that we've undone a stoke, we cannot undo again until another stoke is added
+		undo_element_list_num = UNDO_NONE
```
aaronfranke: Would it be possible to allow many undos?

2d/gd_paint/ToolsPanel.gd
```diff
@@ -0,0 +1,107 @@
...
+	label_stats.text = "Brush objects: " + String(paint_control.draw_elements_list.size())
+
+
+func button_pressed(button_name):
+
+	# If a brush mode button is pressed
+	if button_name == "mode_pencil":
+		paint_control.brush_mode = "pencil"
+	elif button_name == "mode_eraser":
+		paint_control.brush_mode = "eraser"
```
aaronfranke: Why not use enums here?

2d/gd_paint/PaintControl.gd
```diff
@@ -0,0 +1,261 @@
...
+
+func _process(delta):
+
+	var mouse_pos = get_viewport().get_mouse_position()
+
+	# Check if the mouse is currently inside the canvas/drawing-area
+	is_mouse_in_drawing_area = false
+	if mouse_pos.x > TL_node.global_position.x:
+		if mouse_pos.y > TL_node.global_position.y:
+			is_mouse_in_drawing_area = true
```
aaronfranke: var global_pos = TL_node.global_position
    is_mouse_in_drawing_area = mouse_pos.x > global_pos.x && mouse_pos.y > global_pos.y

2d/gd_paint/PaintControl.gd
TwistedTwigleg: It would be possible, as I mentioned in the initial post.

It has been awhile so I don't remember exactly how the demo works, but if I recall correctly the elements drawn are stored in a list. The current undo functionality works because it stores the index of the element before drawing. When the undo button is pressed, it just removes all of the elements going forward after the stored index.

While this works for a single undo, now I can think of better ways to go about doing something like this. If there is interest and the demo is still desired, then I could see about rewriting the code to allow for multiple undos.

2d/gd_paint/ToolsPanel.gd
TwistedTwigleg: If I remember correctly, enums were not available in GDScript at the time. Enums would be better now that it is supported.

No doubt most of the code could/should be rewritten to use more modern GDScript practices. It is just a matter of finding the time/energy to rewrite the code.

2d/gd_paint/PaintControl.gd
TwistedTwigleg: I'd say this change is debatable.

First, I feel the name `global_pos` isn't very descriptive of what the variable actually contains. The TL_Node is the top left corner of the drawing area. Using the name `global_pos` to me sounds like it is the global position of the node that PaintControl.gd is attached to. I don't know if defining a variable makes sense in this case, as it is only used once, but regardless I think the variable should be named something different, like `topleft_global_pos` or `tl_global_pos`, as that better defines what the variable stores.

Also, I'm not the biggest fan of assigning variables directly through a conditional, as it makes the code harder to read and understand (in my opinion), which is kinda opposite of the goal of a demo.

I do think that merging both of the if statements together makes sense. So the code would be the following:

~~~
is_mouse_in_drawing_area = false
if mouse_pos.x > TL_node.global_position.x and mouse_pos.y > TL_node.global_position.y:
  is_mouse_in_drawing_area = true
~~~

or if using a variable for the global position:

~~~
is_mouse_in_drawing_area = false
var TL_global_pos = TL_node.global_position
if mouse_pos.x > TL_global_pos.x and mouse_pos.y > TL_global_pos.y:
  is_mouse_in_drawing_area = true
~~~

TwistedTwigleg:
Updated this pull request so that it works with Godot 3.1. I also made some changes to make the code clearer and updated the comments accordingly.

2d/gd_paint/PaintControl.gd
```diff
@@ -0,0 +1,260 @@
...
+const IMAGE_SIZE = Vector2(930, 720)
+
+
+func _ready():
+	# Get the top left position node. We need this to find out whether or not the mouse is inside the canvas
+	TL_node = get_node("TLPos")
+	set_process(true)
+
+
+func _process(delta):
```
aaronfranke: To suppress Godot 3.1's warnings, prefix unused arguments with `_`.

2d/gd_paint/PaintControl.gd
```diff
@@ -0,0 +1,260 @@
...
+		undo_element_list_num = UNDO_NONE
+		# NOTE: if we only had shape brushes, then we could remove the above line and could let the user
+		# undo until we have a empty element list
+	
+	# Otherwise we're removing a either a pencil stroke or a eraser stroke.
+	else:
+		# Figure out how many elements/brushes we've added in the last stroke
+		var elements_to_remove = brush_data_list.size() - undo_element_list_num
+		# Remove all of the elements we've added this in the last stroke
+		for elment_num in range(0, elements_to_remove):
```
aaronfranke: Above the line with the `for` loop, put `#warning-ignore:unused_variable`

2d/gd_paint/ToolsPanel.gd
```diff
@@ -0,0 +1,119 @@
...
+var save_dialog
+
+
+func _ready():
+	# Get PaintControl and SaveFileDialog
+	paint_control = get_parent().get_node("PaintControl")
+	save_dialog = get_parent().get_node("SaveFileDialog")
+
+	# Assign all of the needed signals for the oppersation buttons
+	get_node("ButtonUndo").connect("pressed", self, "button_pressed", ["undo_stroke"])
```
aaronfranke: Add `#warning-ignore-all:return_value_discarded` somewhere, probably above the `# Assign` comment.

2d/gd_paint/Paint_root.tscn
```diff
@@ -0,0 +1,221 @@
...
+text = "Clear picture"
+
+[node name="SaveFileDialog" type="FileDialog" parent="."]
+margin_right = 600.0
+margin_bottom = 400.0
+resizable = true
+access = 2
+filters = PoolStringArray( "*.png" )
+current_dir = "/home/ubuntu_noah/Documents/New_2019_GitHub_Repositories/godot-demo-projects/2d/gd_paint"
+current_path = "/home/ubuntu_noah/Documents/New_2019_GitHub_Repositories/godot-demo-projects/2d/gd_paint/"
```
aaronfranke: This isn't really a problem with this demo, as it's more of a Godot bug, but I'm making a quick note here. This information shouldn't be saved or committed. You can leave this as-is in the meantime.

TwistedTwigleg:
Made changes based on feedback (thanks @aaronfranke!).

The blurry appearance of the icons is due to the `.import` file for the icons not being stored, so it automatically applies a filter, which makes them blurry. Unfortunately, there is no good way to fix this without increasing the size of the file itself dramatically, since the image is 32x32 pixels in size, which (in my opinion) adds unnecessary bloat to the git repository.

As for the circle looking like an octagon, it is the best I can do in a 16x16 grid. Feel free to make a PR changing the image if you want, but personally I think for the purpose of this demo, showing off how to use the `_draw` function, it works fine.

2d/gd_paint/PaintControl.gd
```diff
@@ -0,0 +1,261 @@
...
+	var img = get_viewport().get_texture().get_data()
+	# Crop the image so we only have canvas area.
+	var cropped_image = img.get_rect(Rect2(TL_node.global_position, IMAGE_SIZE))
+	# Flip the image on the Y-axis (it's flipped upside down by default)
+	cropped_image.flip_y()
+	
+	# Save the image with the passed in path we got from the save dialog
+	cropped_image.save_png(path)
+	
+	return
```
aaronfranke: Pointless return at the end of a function.

TwistedTwigleg:
Thanks for the new tool sheet @aaronfranke, it looks great!
I added it to the project and removed the pointless return statement in `PaintControl.gd`.

2d/gd_paint/Paint_root.tscn
```diff
@@ -0,0 +1,221 @@
...
+[node name="DrawingAreaBG" type="Panel" parent="."]
+margin_left = 350.0
+margin_right = 1280.0
+margin_bottom = 720.0
+custom_styles/panelf = SubResource( 1 )
+custom_styles/panel = SubResource( 1 )
+custom_styles/panelnc = SubResource( 1 )
+
+[node name="PaintControl" type="Control" parent="."]
+editor/display_folded = true
```
aaronfranke: Please remove all `editor/display_folded` lines. Here's a bash alias that you can add to `~/.bash_aliases`:

    alias tscnnoeditor="find . -name '*.tscn' -exec sed -i '/^editor\//d' {} \;" # Remove "editor/display_folded"

PR #93, opened 2017-11-13
Updated SDF demo to 3.0

RoyBerube:
Signed Distance Field Font Demo updated for 3.0


LikeLakers2:
Wait, why are we putting this in the 2D folder, instead of its own folder inside the 2D folder...?

akien-mga:
Ah nice catch, I had overlooked that. It should be `git mv` to a properly folder.

PR #94, opened 2017-11-13
Move SDF demo to proper folder

RoyBerube:
Signed distance field font demo was not created in its own folder. Moved files to their own folder.


reduz:
this is cool because we have no way to import a font as SDF :P

PR #99, opened 2017-11-25
3D Dynamic Controller Platformer demo

kubecz3k:
3D platformer demo based on current platformer 3d kinematic demo.
Player logic was written from scratch (not using code from 2.x demo directly), but it's based on platformer 2.x code ideas. 
I hoped for this demo to be a lot easier to understand than what we had before. Physics code, game logic and animation logic are untangled.  (game logic and physics are in `player.gd` animation logic is in `player_armature.gd` file)
I had strong desire to modify file structure as well (create player and enemy directories), but I resigned since that way it will be a lot easier to keep in sync with kinematic platformer (only `player.gd` logic was touched, other elements can be always copied from kinematic platformer demo). 
Also I have added moving platform in the second area.


kubecz3k:
Hmm worth to mention this demo has a lot simpler logic than previous one, maybe even to simple since it doesn't really try to preserve outside forces which are influencing the player. But I need to say our previous demo was **not** doing good job in this area either and the logic was a lot more complicated (with no practical benefit in my opinion).  
Because of that this demo probably should be reviewed first.

kubecz3k:
After some consideration I think I should do a better job. This demo is quite readable and while it's simple and it works in platformer3d scenario it doesn't really shows how dynamic controller should look like, because of what I mentioned in previous comment.

PR #100, opened 2017-11-27
3D Dynamic Controller Platformer demo

kubecz3k:
3D platformer demo based on current platformer 3d kinematic demo.

Player logic was written from scratch (not using code from 2.x demo directly). The core of the physics processing is not based on 2.x demo. I hope it should be simpler to understand and in my opinion the forces are preserved in a better way than in 2.x demo 
The control scheme is based around the idea of separating calculations for forces applied by the player, and forces which are influencing the character due to 'external' physics interactions (like being pushed by some other body). 

I had strong desire to modify file structure as well (create player and enemy directories), but I resigned since this way it will be a lot easier to keep in sync with kinematic platformer (only player.gd logic was touched, other elements can be always copied from kinematic platformer demo).

Also this is just a proposition I can do a port of the 2.x code if that's more desired.


LikeLakers2:
If you don't mind me asking, why `.SCN` files instead of `.TSCN`? TSCN's better for version control if my memory serves me right.

kubecz3k:
Sure it's better. As I mentioned I wanted to leave easy possibility to easy sync with kinematic version of this project (by copy pasting everything except player scene and script).

kubecz3k:
Also some background: Those demos were created at the time when there was no tscn file format yet. We probably should convert all demos to tscn, but I think it's out of scope for this pr.

3d/platformer_rigid/enemy.gd
```diff
@@ -0,0 +1,78 @@
...
+		
+		if (cc):
+			if (cc is preload("res://bullet.gd") and not cc.disabled):
+				set_mode(MODE_RIGID)
+				dying = true
+				#lv = s.get_contact_local_normal(i)*400
+				state.set_angular_velocity(-dp.cross(up).normalized()*33.0)
+				get_node("AnimationPlayer").play("impact")
+				get_node("AnimationPlayer").queue("explode")
+				set_friction(1)
```
aaronfranke: Deprecated in 3.1

3d/platformer_rigid/player_armature.gd
```diff
@@ -0,0 +1,44 @@
+extends Spatial
+
+onready var animation_tree_player = get_node("../AnimationTreePlayer");
```
aaronfranke: AnimationTreePlayer has been deprecated in Godot 3.1.

kubecz3k:
@aaronfranke I don't think this pr its really desired at current time since, there is new platformer demo in the works and as far as I remember there are some plans on creating new dynamic character controller demo at some point.

PR #103, opened 2017-12-01
Fixed the joypad demo to work with Godot 3 Beta 1

### LINKED ISSUE #102: joypad demo needs more axes nodes because of JOY_AXIS_MAX=10
The constant JOY_AXIS_MAX is set to 10, which is used to iterate over the axes nodes. But there are only 7 nodes for the axes in the scene.

When setting 7 instead of JOY_AXIS_MAX in the for loop, the demo works.

TwistedTwigleg:
Closes #102

I added the three missing axis nodes needed to make the joypad demo work with Godot 3 Beta 1.
I also changed the window size to fit the new additional nodes and some of the files were changed automatically when I opened the project.


akien-mga:
Thanks!

PR #104, opened 2017-12-02
fixed font rotation animation in sdf demo

jaggygames:
The signed distance field font didn't rotate during the demo and produced the following errors instead.

2017-12-02-101721_1366x768_scrot

The font now should now properly rotate. Tested on Godot 3.0 Beta 1.


jaggygames:
#170 fixed this issue!

PR #105, opened 2017-12-02
added C# version of 2D nav mesh

brogan89:
Hey just learning to use Godot through doing the demo projects and converting them into C#. Thought I may as well help out too :)


mono/Navigation Polygon (2D) (C#)/Navigation.cs
```diff
@@ -0,0 +1,78 @@
...
+        SetProcess(true);
+    }
+
+    public override void _Input(InputEvent _event)
+    {
+        if (_event is InputEventMouseButton mouse)
+        {
+            if (mouse.Pressed && mouse.ButtonIndex == 1)
+            {
+				GD.Print($"Click Pos: {mouse.Position}");
```
LikeLakers2: Did you mean to use spaces here instead of tabs? It seems a little inconsistent from all the other lines having spaces.

brogan89: I noticed that, it doesn't look like that in vscode. I have it set to [Tabsize: 4]. I'm not sure what I did wrong sorry. Even after formatting it didn't show any changes in git.

brogan89: This is it in source tree [img](https://www.dropbox.com/s/gde3bgm0ln99uds/Capture.PNG?dl=0) so you know I'm not lying :)

mono/Navigation Polygon (2D) (C#)/Navigation.cs
```diff
@@ -0,0 +1,78 @@
...
+    public override void _Ready()
+    {
+        agent = GetNode("Agent") as Sprite;
+    }
+
+    public override void _Process(float delta)
+    {
+        if (path.Count > 1)
+        {
+            var toWalk = delta * SPEED;
```
TwistedTwigleg: I think we should avoid using `var` as C# is a static typed language, which means using `var` forces the compiler to figure out the type. Though this works in most cases, it's best to be implicit when defining variables (also defining the type follows the style set in the C# pong demo).

So in this case: `var` -> `float` or `double`

mono/Navigation Polygon (2D) (C#)/Navigation.cs
```diff
@@ -0,0 +1,78 @@
...
+            while (toWalk > 0 && path.Count >= 2)
+            {
+                var pFrom = path.Last();
```
TwistedTwigleg: See above:
`var` -> `Vector2`

mono/Navigation Polygon (2D) (C#)/Navigation.cs
```diff
@@ -0,0 +1,78 @@
...
+                var pTo = path[path.Count - 2];
```
TwistedTwigleg: See above:
`var` -> `Vector2`

mono/Navigation Polygon (2D) (C#)/Navigation.cs
```diff
@@ -0,0 +1,78 @@
...
+                var d = pFrom.DistanceTo(pTo);
```
TwistedTwigleg: See above:
`var` -> `Vector2`

mono/Navigation Polygon (2D) (C#)/Navigation.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    public override void _Input(InputEvent _event)
+    {
+        if (_event is InputEventMouseButton mouse)
+        {
+            if (mouse.Pressed && mouse.ButtonIndex == 1)
+            {
+				GD.Print($"Click Pos: {mouse.Position}");
+                begin = agent.Position;
+                end = mouse.Position - this.Position;
```
TwistedTwigleg: To keep this demo as a complete conversion of the GDScript demo, please add the comment the following comment (taken from the GDScript version): `# Mouse to local navigation coordinates`

brogan89: Ah yeap fair enough, I'll remember that for when I do more. Wanted to start with a small easy script. I'm just do used to using `var`. Is it all good using  the `Linq` namespace as well? Im only using it for `Last()` but to be consistent with the GDscript should I use `[length - 1]` instead?

TwistedTwigleg: I don't think it matters, as long as `Linq` is in Mono.
I suppose `[length - 1]` would be a more 'complete' conversion, but `Last` is more clear on what's happening, so whatever you want to do works for me :smile:

brogan89: cool made those changes. I realised I was using Linq to convert array to list anyway so left it. Also I just seen its pushing the `.proj` and '.sln' shouldn't these should be ignored?

brogan89:
Also removed Vector2 decorations as they're structs so will never be null.

TwistedTwigleg:
Looks good to me!

(I have no idea on the `.csproj` and `.sln` files, sorry)

brogan89:
I'm not 100% sure what they do but they are for IDE/Text Editor information. I'm guessing they link Godot to the `.cs` files but I could be totally wrong too. In all my Unity repos we ignore them, but I see the other mono project has them in the repo so ill leave it. - Does someone do the merge in the main repo? I've never done a pull request before lol

TwistedTwigleg:
Good to know what they’re for, as I had no idea.

As for your question: one of the developers will merge the PR, assuming they’re cool with the addition and it’s ready to be merged. You may need to squish your commits into one commit, but I’d wait for now to make sure everyone’s okay with the changes.

mono/Navigation Polygon (2D) (C#)/Navigation.cs
```diff
@@ -0,0 +1,81 @@
...
+                GD.Print($"Click Pos: {mouse.Position}");
+                begin = agent.Position;
+
+                // Mouse to local navigation coordinates
+                end = mouse.Position - this.Position;
+                UpdatePath();
+            }
+        }
+    }
+}
```
aaronfranke: Please ensure that there are newlines at the end of each file.

brogan89:
oh man I completely forgot about this. It can be discarded if you want. It doesn't really bring much new to the table.

aaronfranke:
It's perfectly fine if you want to close this. We do indeed have many demos already.

PR #108, opened 2017-12-03
Redid the particles 2D demo

### LINKED ISSUE #82: Particles 2D demo is broken
Current master branch for Godot 3.x.

Nothing is displayed, some of the particles have been converted to plain Nodes when converting from 2.1. It should likely be redone from scratch.

TwistedTwigleg:
I redid the particles 2D demo, staying as close as possible to the Godot 2 version as I could. closes #82 

*  I added all of the particles that were in the Godot 2 version, trying my best to keep them similar. I did change the fire particles texture to a simple gray scale texture (that I made), and I also made the magic particles have a trail because I thought it looked cool, and would show off the new trails functionality.
*  I added two additional particle examples for the new additional mask emitting modes.
*  I changed the icon to a picture of the fire particles

I'm fairly happy with out it turned out, though the explosion and smoke particles could probably use some polish:
particles_editor_picture

I noticed that running the example gives an entirely different result with the fire particles:
particles_runtime_picture
~~I'm not sure why it looks so dramatically different from the editor. I tried changing some things around, and the only conclusion I could gather is that I'm using a canvas material (to make it blend with addition instead of mixing), so maybe there's a bug there? Or maybe it's just a linux thing?~~

EDIT: Seems there is already an open issue for the canvas material bug (https://github.com/godotengine/godot/issues/13443)


akien-mga:
Sorry for the huge delay :/

There are now some merge conflicts on .import files to fix. There's also an Image resource embedded directly in the TSCN file, which makes it huge. It would be better to keep it as a PNG resource.

TwistedTwigleg:
No worries! I imagine the amount of work you guys put in maintaining Godot's repositories is tremendous. I'm totally cool with having to wait :smile: 

I'll get these fixed soon!

TwistedTwigleg:
I found out that the three embedded textures are the point textures for the outlines. I'm not sure on what I should do though. Should I save them to their own .PNG files? should I remove the outline particles entirely?

I also fixed an issue with the fire particles. Now it looks like the original picture again.

(Also, I should squish these commits into one, right?)

akien-mga:
>  Should I save them to their own .PNG files? should I remove the outline particles entirely?

I think it should be possible to save them to their own PNG files from the inspector.

> (Also, I should squish these commits into one, right?)

That would be great :)

TwistedTwigleg:
I tried exporting the point images into .PNG files, but it doesn't work. The particles just emit from a single point as if it doesn't have a texture. I also tried exporting the point images to .tex and .tres files, but it has the same issue when running the project. Strangely when it's exported into .tex or .tres files, it does work in the editor, just not when you launch it.

I'll see if I can find a way to get it to work.

TwistedTwigleg:
Alright, I found a way to get them out of the scene file.

For the particles that emit from the pixels, you can save the imagetexture as a .tscn file and then load that. Everything works in both the editor and when you play the project.

For the outline and directional outline particles, you have to save the image inside the imagetexture, and then everything works.

I'm not sure why it works that way, but now I can safely say the images are outside of the main .tscn file (and now inside their own .tscn files)

TwistedTwigleg:
I think I successfully squashed all of the commits into one. :tada:

akien-mga:
Thanks!

PR #109, opened 2017-12-03
Fix animation property in pause demo.

RaXaR:
Renamed rotation_deg to rotation_degrees.


akien-mga:
Thanks! Sorry for the delay.

Issue #179, opened 2018-01-02
demo projects should be updated and tidied up ready for 3.0 release


sambler:
Currently most of the demo projects fail to run using the current (v3.0) runtime. If they are opened in the editor, the *.import files get updated with [deps] entries and then they can be started with the runtime. Maybe that step should be handled in the runtime.

There are also some small issues that should be cleaned up in some of them, eg. the hexagonal map appears to have tiles out of place and visual pong doesn't work properly. Maybe anyone familiar with the demos can have a look through and see that each one works as expected.

sambler:
Actually the contents of the .import folders also need to exist for the runtime to work. Maybe there needs to be a code change to fix this so that the .import contents are always rebuilt by the runtime.

As the runtime fails to work without the .import folder, should these be listed in .gitignore? Should it remain in a '.' folder or be made more visible?

vnen:
The `.import` folder is built if missing, but only in the editor. Import functionality is not present in a release build. For the demos, we could add the `.import` folder to the repository so it's ready to play, but I'm not sure if it is worth it. After all, they're all meant to be opened in editor to see how to use the features, not to mention it would increase the download size a lot.

Calinou:
> For the demos, we could add the .import folder to the repository so it's ready to play, but I'm not sure if it is worth it.

I don't think it's a good idea to commit files which could be considered build artifacts to a repository, but uploading a ZIP archive of demos with pre-imported assets as a release (using GitHub Releases) makes sense.

PR #180, opened 2018-01-02
Automatic .import updates made by opening projects in the godot editor

sambler:
Related to issue #179 this pull request is the result of opening each project in the editor, the only changes are what godot automatically saves.

These changes to the *.import files allow the demo projects to be opened using the godot runtime and prevent reimporting every time they are opened in the editor.

There were also changes saved to gui/translation/text.xx.translation but as I am unable to verify the changes and they don't prevent the runtime starting, so I didn't include those changes here.


akien-mga:
Merged manually as abb08c9 to fix a build conflict.

PR #181, opened 2018-01-04
Fixed some issues in your Mono Pong demo

LuckySvenn:
The previous code would only allow the ball to change its y direction off the second paddle, keeping its x direction, so it would never bounce off the second paddle. Also, it would only bounce off the ceiling, and not the floor.

### DIFF
--- a/mono/monkey_pong/CeilingFloor.cs
+++ b/mono/monkey_pong/CeilingFloor.cs
@@ -3,14 +3,11 @@ using System;
 
 public class CeilingFloor : Area2D
 {
-    [Export]
-    private int yDirection = 1;
-
     public void OnAreaEntered(Area2D area)
     {
         if (area is Ball ball)
         {
-            ball.direction += new Vector2(0, yDirection);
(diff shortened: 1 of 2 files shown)


LikeLakers2:
You'll want to make these changes in the GDscript version too! -> https://github.com/godotengine/godot-demo-projects/tree/master/2d/pong

akien-mga:
That's not the right fix, if there's a bug in the first place. Those are *export* variables, they're meant to be changed manually in the editor to adapt the instanced scenes. 1 is for left paddle, for the right paddle you'd set -1.

LuckySvenn:
Yeah, I was wondering why the GDScript version was working correctly. It looks like the C# project doesn't show any Exported variables to the editor, so I suppose this is a Beta issue. Sorry, I downloaded these projects to learn Godot and how to use C# with it, didn't realize I was bumping into a different kind of issue.

akien-mga:
Yeah I think C# only shows export vars after you've compiled once, that's a known bug but it does make it harder to understand this demo :D I'll see with other devs what's the best fix in the meantime.

Issue #174, opened 2017-12-07
Very low FPS in Platformer 3D demo under some circumstances


Calinou:
When running the demo and moving to some areas of the demo level, FPS will become very low, even on high-end hardware (tested on an i7-6700K and a GTX 1080, I get 10-15 FPS when it happens).

This may not be due to the renderer, but rather due to a scripting error.

See the console output sent by an user on Discord below:

![godot_platformer](https://user-images.githubusercontent.com/180032/33729463-07bb58e2-db7e-11e7-8233-749c69cd0ef6.png)

**Edit:** Someone on Discord commented that this may be due to the coin animation.

> looks like the low fps issue with the 3d platformer is the coin animation uses 
> `rotation_deg`
> which changed to 
> `rotation_degrees`

TwistedTwigleg:
@Calinou is this fixed now?

I fixed the animation property before running the project and it seems to have fixed the low FPS (for me at least), but I don’t have anything to compare it against. Do you mind testing?

NathanLovato:
I don't have the issue. As the bug dates back from 3.0 beta and there's no update I'm closing this. You can open another issue if need be

PR #182, opened 2018-01-06
Platformer3D fixes

### LINKED ISSUE #81: Platformer 3D has broken dependency in tiles.scn
The `tiles.scn` scene (source for the MeshLibrary) has a dependency on `texture.tex`, which is no longer there in 3.0. It should be `texture.png` directly.

TwistedTwigleg:
I fixed the following in the platformer 3D demo:

- Renamed rotation_deg to rotation_degress in the coin animation.
- Changed the dependency in tiles.scn from texture.tex to texture.png
- (The import files were updated when I opened the project)

This should close #81 and maybe #174. I didn't find any FPS drops after fixing the coin animation, but more testing is probably needed just to make sure everything is running smoothly.


akien-mga:
Merged manually as 36923a2 to remove the conflicting *.import changes.

PR #184, opened 2018-01-06
Demo cleanup for godot 3

### LINKED ISSUE #179: demo projects should be updated and tidied up ready for 3.0 release

TwistedTwigleg:
NOTE: this is a repost of a previous PR, just without merge conflicts. I couldn't find a way to rebase without the potential of ruining everything. I'm still fairly new to git, and I felt it would be easier this way and would have less chance of me screwing up the repo. My apologies if this (making another PR) causes issues :cold_sweat: 

___

I went through all of the demos and tidied them up for Godot 3. Here's what I fixed:
- Both platformer 2D demos now also use the Z key to shoot, so that it follows the same input as the platformer 3D demo
- Both platformer 2D demos bullets now have proper particle effects, designed to look as similar as possible to the platformer3D demo
- The rich_text_bbcode demo link was changed to link to the documentation page, as opposed to the github wiki page (is there a github wiki for Godot still?)
- Redid the some of the map for the hexagon demo, as there was some tiles out of place.
- Fixed a bug in the input_mapping demo that would cause an error if you pressed a key before pressing any buttons after just launching the demo.
- Fixed the rotation property in the tween demo (was set_rotation_in_degress, now it's set_rotation_degress)
- Moved the troll in Isometric so that it's not in the wall by default
- Added a default environment to sdf_font

I did not try to fix any of the multiplayer demos, the visual scripting demo(s), and the C# demo(s), as I feel I'm not the best person to make changes to any of those demos.

This somewhat closes #179. The multiplayer, C#, and visual script demos may still need some tidying, so this probably shouldn't fully close #179.

I can squash these commits down if you guys want, or make separate PRs.


akien-mga:
Great work, thanks!

PR #187, opened 2018-01-09
Fixed ball direction in visual pong. randf instead of randi

sudoguy:
randi replaced on randf, because after _on_area_entered ball always moved down.
screen shot 2018-01-09 at 23 44 03


akien-mga:
Thanks! I went ahead with a different fix (5116684) as it needs to be `2 * randf() - 1` to be within [-1; 1], otherwise the ball always goes down.

PR #190, opened 2018-01-13
Multiplayer Bomber fix parser error

### LINKED ISSUE #189: Multiplayer Bomber Parser Error
OS: Ubuntu Linux 16.04
Godot version: latest (master branch)
Godot Demo Projects version: latest (master branch)

The Multiplayer bomber doesn't run.

Output:
```
Parser Error: Name for argument 'name' can't shadow class property of the same name.
0 - res://gamestate.gd:59
```

It ran a week ago with the same code so I don't know if the demo should be updated or if it's an Engine's issue.

LinuxUserGD:
This is my first pr so it isn't perfect.
Fixes #189
I changed 'name' to 'namespace' because 'name' got a property.


akien-mga:
Using `namespace` everywhere seems to be overkill when the issue was just some shadowing of `name` in a few function parameter, which could be replaced by `new_name`.

akien-mga:
I went ahead with a slightly different solution, and fixed a couple other issues while at it in e182dd7. But thanks a lot for giving it a try, and don't hesitate to make other PRs :)

Just make sure to use more meaningful commit logs than "test" ;)
See http://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html for a guide on Git usage and the workflow we're using for Godot.

LinuxUserGD:
Ok, thank you for reviewing. Now I know how to do it better next time :)

PR #194, opened 2018-01-15
Fix ball bounce only working for left paddle

noidexe:


akien-mga:
This is the same as #181, it's just hiding a bug with export variables being lost. CC @neikeq

akien-mga:
See https://github.com/godotengine/godot-demo-projects/blob/master/mono/monkey_pong/pong.tscn#L67, `ballDir` is properly defined for the right paddle. But this value is lost when opening the project due to https://github.com/godotengine/godot/issues/15371

noidexe:
Sorry, was about to close it myself. I just noticed the bug with exported variables.

akien-mga:
Well we might need to merge such a fix if we can't get https://github.com/godotengine/godot/issues/15371 fixed in time for 3.0 though. But yeah ideally we'd fix that issue and unbreak the demo.

PR #195, opened 2018-01-15
Fix ball bounce only working for ceiling

noidexe:


akien-mga:
Same as #194, this works but it's hiding the real bug.

BTW as far as a Pong implementation is concerned this is also a simpler code, but the aim of this demo is to show how to reimplement the GDScript pong in C# using the same editor features, and namely here showcasing export variables.

PR #218, opened 2018-02-12
bugfix/android-iap-signal-instead-of-timer: Update the IAP project demo

xsellier:
Related to: https://github.com/godotengine/godot/pull/16647

### DIFF
--- a/misc/android_iap/iap.gd
+++ b/misc/android_iap/iap.gd
@@ -1,6 +1,8 @@
 
 extends Node
 
+const IAP_PLATFORM = "Android"
+
 signal purchase_success(item_name)
 signal purchase_fail
 signal purchase_cancel
@@ -15,7 +17,13 @@ signal consume_not_required
 signal sku_details_complete
 signal sku_details_error
 
+signal iap_disconnected
+signal iap_connected
+
 onready var payment = Globals.get_singleton("GodotPayments")
+
+# Used to be able to stub the android behavior
+onready var is_debug_build = OS.is_debug_build() and OS.get_name() != IAP_PLATFORM
 
 func _ready():
 	if payment:
@@ -26,7 +34,6 @@ func _ready():
 func set_auto_consume(auto):
 	if payment:
 		payment.setAutoConsume(auto)
-
 
 # request user owned item, callback : has_purchased
 func request_purchased():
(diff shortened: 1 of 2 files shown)


misc/android_iap/iap.gd
```diff
@@ -15,80 +17,92 @@ signal consume_not_required
...
+# Used to be able to stub the android behavior
+onready var is_debug_build = OS.is_debug_build() and OS.get_name() != IAP_PLATFORM
 
 func _ready():
-	if payment:
-		# set callback with this script instance
-		payment.setPurchaseCallbackId(get_instance_ID())
+  if payment:
+    # set callback with this script instance
+    payment.setPurchaseCallbackId(get_instance_ID())
```
bojidar-bg: Use tabs, not spaces 😃

xsellier: Done :)

akien-mga: You forgot to push I guess, it's still the same commit :)

xsellier:
Indeed, since I amended my commit I had to force the push.

misc/android_iap/iap.gd
```diff
@@ -27,82 +35,88 @@ func set_auto_consume(auto):
...
 		payment.setAutoConsume(auto)
 
-
 # request user owned item, callback : has_purchased
 func request_purchased():
 	if payment:
 		payment.requestPurchased()
 
 func has_purchased(receipt, signature, sku):
-	if sku == "":
```
akien-mga: Those changes add a lot of noise to the diff for no practical reason, and even though it's now consistent in this file, it's still inconsistent with many of the other demos where `"` is used.
So I prefer we avoid sneaking in such changes in unrelated commits, and if we decide on a given style guide for this, then we can do the change. Personally I think `"` is fine for string quotes, it's much rarer to have to use (and then escape) `"` in a string than `'`.

xsellier: Sure.

aaronfranke:
What is the status of this PR? Is this fix still desired to be added to the 2.1 branch?

misc/android_iap/iap.gd
```diff
@@ -27,7 +35,6 @@ func set_auto_consume(auto):
 	if payment:
 		payment.setAutoConsume(auto)
 
-
```
aaronfranke: Per the [style guide](https://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html), functions should be separated by two empty lines.

xsellier:
Since no one is interested in this PR, I'm closing it.

misc/android_iap/iap.gd
xsellier: This PR is almost 2 years old. There is no need for further review.

PR #220, opened 2018-02-19
Increase sky radiance size to 128 (from 64) in the Material Tester demo

Calinou:
This makes reflections look nicer, especially on materials with low roughness values. The performance impact is probably negligible.

**Before:**

material_tester_sky_radiance_size_64

**After:**

material_tester_sky_radiance_size_256


reduz:
And makes it crash on mobile, but I guess should be fine now thst es2
backend is coming along

aaronfranke:
Now that 3.1 is released and we have GLES 2, this should be re-based and re-tested.

Calinou:
Rebased. I also increased the value to 512 as it's the default value in Environment.

clayjohn:
Value should be set to 128 as that is the new default as of https://github.com/godotengine/godot/pull/29182

Not sure what it is currently set at. Given the new changes 512 is a little too high. The visual difference between 128 and 512 is very slight.

Calinou:
@clayjohn It seems loading this PR automatically sets the radiance size to 128 now, as 512 was the prior default value. Godot doesn't save default properties in scene files (even in binary format, it seems), so it automtaically uses the new default value. I'll update the commit message accordingly.

akien-mga:
Thanks!

PR #223, opened 2018-03-14
Code cleanup

cbscribe:
I'm working on a tweaked autopep8 that can find these automatically. It's still a bit flaky, but getting there.

* Remove unneeded parentheses
* Add spaces around operators

I'm also tempted to go through and replace some of the `get_node()` calls with `$`, but then I noticed a lot of inconsistent node naming, too, so that might be a bigger undertaking.


pwab:
Nice work! Looks good to me.

**Regarding the `get_node2$`-conversion:**
I know it is not common (and can be circumvented by using groups or signals) but how would you handle string concatenated node paths with the `$`-notation?

**Example:**
```gdscript
func turn_around():
	var list_of_enemies = ["enemy1", "enemy2", "enemy3"]
	for enemy in list_of_enemies:
		get_node("enemy_group/" + enemy).rotation_degrees += 180
```

bojidar-bg:
@pwab For concatenated / computed strings, you still have to use `get_node`.

NathanLovato:
@cbscribe We can already merge this PR if you're okay with it, and see later for other changes? Just give the word and I'll merge

cbscribe:
I'm good!

PR #225, opened 2018-03-16
3D IK demo

### LINKED ISSUE #3: Demos for tool mode usage

### LINKED ISSUE #211: New Demo: IK Demo
I have been working on a relatively simple IK (inverse kinematics) system for Godot 3, intended to be used in the FPS tutorial I'm making. Now that the system is fairly stable, I thinking about maybe sharing the system here on the demo repository.

The IK system works both in the editor and at runtime, the performance is pretty good (though it could use some touching up in a few places), and the demo would add IK implementation for others to use and build on. The code could also help as a reference on how to make a EditorPlugin that reacts when exported variables are set and can show how to add nodes to the scene tree (at runtime and in the editor).

You can see the system in action [here](https://www.youtube.com/watch?v=TfAO2W6lO8E), [here](https://www.youtube.com/watch?v=RwouEFDMelM&feature=youtu.be), and [here](https://www.youtube.com/watch?v=mGRH9M3FLis).

I can add a few examples to show how to use the system (like dynamically placing feet on terrain. Another potential use case that I'd like to look into is making ragdolls)

I can also write a tutorial in the docs explaining how the system works. It's fairly simple and should be relatively easy for intermediate Godot users to understand.

What do you guys think?

TwistedTwigleg:
Adds a 3D IK demo with the following IK solvers: Look at and FABRIK.
Three scenes are included, one of which shows a example use case using both solvers.

Please give it a try and let me know if anything is not working correctly!

Closes #211 (maybe closes #3 too, as SADE uses `tool` heavily)

I've decided to call the plugin SADE, or *Skeleton Additions and Enhancements*. (NOTE: I know SADE implies there is a word starting with D. I decided "additions" should contribute both the A and the D so the name was not SAE, which looks rather strange in my opinion).

The plugin code contains lots of parenthesis and semicolons. I'll remove these if you guys want (I forgot about them until I started writing this, whoops 😅 ). I think the other code not related to SADE is written without semicolons and parenthesis, I'll double check when I remove them in SADE (if you guys want me to remove them in SADE).

Also, in the FABRIK system I have all of the code for constraints commented out. This is because I could not get it working, but I thought I'd leave it in case someone else wants to work on it as they may benefit from my (not working) constraint system.

There are a one annoying error with the plugin I could not fix: whenever you save a scene with a `IK_FABRIK` node it prints several errors to the console. Thankfully it seems that's all that happens and it only happens when saving in the editor. If anyone knows how to fix it, please let me know!


3d/IK/ExamplePlayer.gd
```diff
@@ -0,0 +1,244 @@
...
+const ACCEL= 8.5
+
+# A vector for storing the direction the player intends to walk towards.
+var dir = Vector3()
+
+# Sprinting variables. Similar to the varibles above, just allowing for quicker movement
+const MAX_SPRINT_SPEED = 34
+const SPRINT_ACCEL = 18
+# A boolean to track whether or not we are spriting
+var is_spriting = false
```
pwab: Did you mean spri**n**ting? (Also in the comment above)

pwab:
The icon at `3d/IK/icon.png` could be changed to something individual.

pwab:
**Regarding the errors/warnings:**

1. `ERROR: No loader found for resource: res://Blender/Game_stuff/Godot_FPS/GodotBattleBot_Colors.png`
2. `ERROR: No loader found for resource: res://Blender/Game_stuff/Godot_FPS/GunTextures.png`
3. `WARNING: Collada::_parse_mesh_geometry: Primitive type "polygons" is not well supported (concave shapes may fail). To ensure taht the geometry is properly imported, please re-export using "triangles" or "polylist".`

But I wasn't able to reproduce them. Maybe some issues with import files?

akien-mga:
I haven't checked the demo yet (though from what I see it seems great), but I'm concerned about the file names which are capitalized, as it's not consistent with all other demos using lowercase snake_case for folder and file names.

3d/IK/ExamplePlayer.gd
TwistedTwigleg: Whoops! Yeah, I meant **sprinting**, not spriting 😅

TwistedTwigleg:
Thanks @pwab, @akien-mga!

I'll start fixing everything soon!

______

> akien-mga:
> ... I'm concerned about the file names which are capitalized, as it's not consistent with all other demos using lowercase snake_case for folder and file names.

I will definitely change the file names. I'm hesitant to change the `SADE` and `IK` folders because they are both acronyms. Thoughts? (I guess it's not a huge issue either way. For now I'll change them to lowercase)

> pwab:
> The icon at `3d/IK/icon.png` could be changed to something individual.

I totally forgot about the icon! Whoops!

> pwab:
> **Things I encountered:** ...

I'll remove the `;` and `if (a == b)` stuff too, along with setting the main scene (I totally forgot, again. I guess I was a bit absent minded yesterday 😅 ), increasing the mouse look speed, and fixing any/all spelling errors I can find.

> pwab:
> **Regarding the errors/warnings:** ...

I think I know what the problem is, and it should be easy to fix. I think the imported materials are looking for the textures, which only exist on my computer. Removing the images from the `.dae` file's materials should fix it.

> pwab:
> By the way: You are saying that the right hand has 3 bones and the left only 2. But the view position is in front of the Bot and therefore I'm not sure if you mean the right hand or the hand on the right side. 😅

I'll add a label showing which side is which! (I think it's relative to the player's view, but honestly I'm not 100% sure!)

______

Thanks again guys!

TwistedTwigleg:
Well, I'm going to have to make another PR, because Github is doing strange things with the renamed folder. For some reason it is not letting me change the `IK` folder to `ik` and now it is including both (while on my machine there is only one folder).

Sorry guys! I'll make a new (working) PR and link it to this one.

PR #226, opened 2018-03-17
3D IK demo (fixed)

### LINKED ISSUE #211: New Demo: IK Demo

TwistedTwigleg:
Alright, hopefully everything is working now. This PR is the same as PR #225. with the fixes suggested there applied. I had to make a new PR because Github had issues with the `IK` folder being renamed to `ik`.

Sorry if making another PR causes issues / extra work. I could not (after several attempts) figure out a way to remove the `IK` folder (despite git saying it was not there... Not sure what was going on).

See #225 for details. Closes #211 (and maybe #3)


TwistedTwigleg:
Thanks pwab!

NathanLovato:
Tried the project, it runs well and that's already a sizeable demo! Thanks for your work. I'm merging as it runs smoothly and the code is well-commented, but I think the code style could use some extra review and work after that if anyone is up for the task. 

I'm talking about things like using abbreviations for variable names while Godot 3's api moved to more readable properties. Leaving some code that's not in use as well. As this is a public demo for learning purposes and the code is quite complex, the more self-documented and organized it can be the better.

TwistedTwigleg:
> Has the IK addon been released in the asset library?

No, I haven’t uploaded the IK addon to the asset library.

I can see about getting it up there later if you guys want, I’ll just need to figure out what’s needed to add it to the asset library.

PR #234, opened 2018-03-27
Rewrite the 2d/navigation demo so it's easier to understand

NathanLovato:
The original code could be made much simpler. I also added some comments for learning purposes.


akien-mga:
It's fine that you renamed the `agent.png` and `path.png` files, but you should remove the old ones (with `git rm` - or alternatively you could have renamed them with `git mv` -- but no need to redo it, manual rename + `git rm` is equivalent for git).

akien-mga:
There's also `navigation.tscn` that should be removed. And of course the `.import` files for the two renamed textures.

NathanLovato:
My bad, forgot to do that. Should be fixed

akien-mga:
Thanks!

PR #236, opened 2018-03-28
Add 2d/navigation_astar demo for grid-based pathfinding

### LINKED ISSUE #235: Astar demo project
I've seen a few questions about pathfinding, got a few requests for tutorials already. That's why I updated the nav2d demo yesterday. 

For any kind of grid-based pathfinding you'd better use Astar. Godot's implementation is very flexible but could use a complete example.

If that's okay I'll work on this today, in a demo that uses Astar and a tilemap node for 2d navigation.

Should it be a new Astar demo or part of the existing navigation one?

NathanLovato:
This includes a commented version of the code for learning purposes
There's a bug in the console every time you call AStar.get_point_path(): Condition p_elem->_root is true
The demo still runs no problem but some help and feedback would be much appreciated.

closes #235


2d/navigation_astar/pathfind_astar_comments.gd
```diff
@@ -0,0 +1,177 @@
...
+export(Vector2) var map_size = Vector2(16, 16)
+
+# The path start and end variables use setter methods
+# You can find them at the bottom of the script
+var path_start_position = Vector2() setget _set_path_start_position
+var path_end_position = Vector2() setget _set_path_end_position
+
+var obstacles = []
+var _point_path = []
+var _half_cell_size = Vector2()
```
LikeLakers2: `onready var obstacles = get_used_cells_by_id(0)`?
`onready var _half_cell_size = cell_size / 2`?

onready stuff runs before _ready(), if you didn't know

LikeLakers2:
Also, why have a uncommented script if you're going to have a second version that is commented? Having two versions is not only a little bloat, but makes code harder to maintain.

2d/navigation_astar/pathfind_astar_comments.gd
```diff
@@ -0,0 +1,177 @@
...
+const DRAW_COLOR = Color('#fff')
+
+# Click and Shift force the start and end position of the path to update
+# and the node to redraw everything
+func _input(event):
+	if event.is_action_pressed('click') and Input.is_key_pressed(KEY_SHIFT):
+		# To call the setter method from this script we have to use the explicit self.
+		self.path_start_position = world_to_map(get_global_mouse_position())
+	elif event.is_action_pressed('click'):
+		self.path_end_position = world_to_map(get_global_mouse_position())
```
LikeLakers2: I'd say try to avoid checking the same condition twice if you can help it.

```gdscript
func _input(event):
	if event.is_action_pressed('click'):
		if Input.is_key_pressed(KEY_SHIFT):
			# To call the setter method from this script we have to use the explicit self.
			self.path_start_position = world_to_map(get_global_mouse_position())
		else:
			self.path_end_position = world_to_map(get_global_mouse_position())
```

2d/navigation_astar/pathfind_astar_comments.gd
```diff
@@ -0,0 +1,177 @@
...
+	# here the id 0 corresponds to the grey tile, the obstacles
+	obstacles = get_used_cells_by_id(0)
+	# The code is split into two methods so it's self-documenting
+	var walkable_cells_list = astar_add_walkable_cells(obstacles)
+	astar_connect_walkable_cells(walkable_cells_list)
+
+
+# Loops through all cells within the map's bounds and
+# adds all points to the astar_node, except the obstacles
+func astar_add_walkable_cells(obstacles = []):
```
LikeLakers2: Not sure I understand the need for a default value for the `obstacles` parameter, since there would be no point to running this function if we're not going to give it any walkable cells.

2d/navigation_astar/pathfind_astar_comments.gd
NathanLovato: Sure, can use onready in this case

2d/navigation_astar/pathfind_astar_comments.gd
NathanLovato: I don't mind either but avoiding nested indent is also a "good practice", at least in Python. That's favoring code readability.

NathanLovato:
```
Also, why have a uncommented script if you're going to have a second version that is the same, just commented? Having two versions is not only a little bloat, but makes code harder to maintain. Just keep the commented one and toss out the uncommented one.
```

I'll wait for someone else's view on the matter, but I don't agree. The two files are for two different persons. A developer doesn't need any comments to read the code, and I'd argue comments make the code harder to read. But beginners often feel like they can't and won't read code, so they need comments (a mini tutorial in this case) to ease them in the project.

Again I don't mind having one or the other, but e.g. if it was for me or a developer I'd remove the commented one. For beginners I'd keep it instead.

NathanLovato:
After giving it some second though, I removed the non-commented version. As it's just a demo, it's better this way. I've also recorded a video to run people through the code yesterday so it'll be more than enough

2d/navigation_astar/pathfind_astar_comments.gd
LikeLakers2: I disagree, because I feel the nested indent actually makes the code more readable than what you have right now. Perhaps we could compromise with this?

```gdscript
func _input(event):
	if not event.is_action_pressed('click'):
		return
	
	if Input.is_key_pressed(KEY_SHIFT):
		# To call the setter method from this script we have to use the explicit self.
		self.path_start_position = world_to_map(get_global_mouse_position())
	else:
		self.path_end_position = world_to_map(get_global_mouse_position())
```

NathanLovato:
**Please don't merge yet**: I'm adding a path-following character with steering. Just heading out to do the groceries now and I'll wrap it up in 2-3 hours

2d/navigation_astar/pathfind_astar_comments.gd
NathanLovato: As I said I don't mind either, it's really a minor detail. Pushing the update in a few, it removes this code

NathanLovato:
It's now ready, with a character that follows the path using steering.

2d/navigation_astar/character.gd
```diff
@@ -0,0 +1,58 @@
...
+var velocity = Vector2()
+
+func _ready():
+	_change_state(IDLE)
+
+
+func _change_state(new_state):
+	if new_state == FOLLOW:
+		path = get_parent().get_node('TileMap').get_path(position, target_position)
+		if not path:
```
0ttoman: The first instance of path contains the centre of the tile the character is currently in. It could be removed it with `path.remove(0)` to avoid sending the character in the wrong direction at the start (keep clicking the target_position repeatedly to see an effect)
Only then I can see a use for the following if statement.

2d/navigation_astar/character.gd
```diff
@@ -0,0 +1,58 @@
+extends Position2D
+
+export(float) var SPEED = 200.0
+
+enum STATES { IDLE, FOLLOW }
+var _state = null
```
0ttoman: It might be easier to read `_state=IDLE` and to leave out the `_ready()` function

NathanLovato: Doing it like that can lead beginners to write code that's wrong. You should pretty much always use a method like `_change_state` to initialize and clean up the object's state. Right now the code reads as: at the start of the program the object has no state, and when it's ready it changes state to IDLE. As soon as you add animation or access any other node you'll need that.

2d/navigation_astar/character.gd
NathanLovato: Thanks fixing it now!

NathanLovato:
Just fixed the remaining bug thanks to someone's help on YT. Should be good to merge now 😄

akien-mga:
Thanks!

0ttoman:
there is still a minor bug in this commit. 
line 26 in 2d/navigation_astar/character.gd :
before accessing the array at index 1 we have to confirm the array has a length of two.
if you click the same tile the character is currently on, the path array in _change_state will have a length of one. you then cant access the second vector2

PR #238, opened 2018-04-06
Fix vehicle vibration and rolling issues.

### LINKED ISSUE #219: Truck Town: Car doesn't stay still (Bullet)
https://github.com/godotengine/godot/issues/12521

Vega-KH:
I think these issues are due to different calculations in the Bullet Physics system.  This fixes the vibration and roll, but additional settings on the suspension also should be looked into.

Fixes #219


LinuxUserGD:
This reduces the vibrations by a huge amount, nice!
With Physics FPS set to 9 I can notice *minimal* vibrations while braking (although not relevant in standard gameplay).

Vega-KH:
Thanks for testing. I'm glad it worked as well on Linux, I only tested on Win10. 

I think there are still some issues with the physics numbers so it's not perfect yet.  I will play around more when I get a chance, but this seemed like the most pressing concern.

akien-mga:
Thanks!

PR #252, opened 2018-04-24
Add State pattern, HFSM and pushdown automaton demo

### LINKED ISSUE #250: Add a State/HFSM/Pushdown demo
The State pattern is super useful in a typical indie game. I made a demo for my Godot course, but as it could be useful to anyone - especially as Lars, an experienced gamedev, reviewed it - we agreed it'd be good to comment and open source it. Adding it ASAP

NathanLovato:
Currently in 2d folder as it's a 2d project. There's no folder that really fits atm.

closes #250


2d/finite_state_machine/debug/keys.gd
```diff
@@ -0,0 +1,6 @@
+tool
+extends Panel
+
+
+func _ready():
+	set_as_toplevel(true)
```
bojidar-bg: This file is the same as explanations.gd, differing only in what they extend.
Consider making just one script which extends `Control` (where `set_as_toplevel` is defined).

bojidar-bg:
@NathanLovato Well, you should get notified by `git` whenever you try to add such a file.

NathanLovato:
I simplified the code: it uses signals instead of static delegates. The character has more states as well now, showing attacks that combine with the jump.

NathanLovato:
Reworked and simplified the code once more. It's much cleaner now, with better separation of concerns. Still works exactly the same as before. Could someone check it and approve for merge if you find it good enough?

akien-mga:
I've only had a very cursory look but it seems good to me :)

PR #257, opened 2018-05-01
Fixed the materials using upper case names in the IK demo.

TwistedTwigleg:
Fixed the materials using upper case names in the IK demo.
I was able to remove the upper case materials by having the materials imported from the `.dae` files stored in the mesh, and then overriding them in each scene.

Now all of the files in the IK demo use snake_case. :+1:


NathanLovato:
Seems all good, thank you!

PR #258, opened 2018-05-01
checking array length beforehand

0ttoman:
This fixes a minor bug when the astar target_position is the same as the staring point. In that case the path-array has the length of one. The bug is already fixed in @NathanLovato's repository but hasn't been pushed yet.


NathanLovato:
Ah I thought I did update here too but I forgot to force push the change. Thanks!

PR #259, opened 2018-05-07
Added "Wheely" C# Mono Project

DavidJVitale:
The "mono" folder could use some more high quality examples. I have created a simple example called "Wheely". Wheely is a wheel-based platformer with a KinematicBody2D Wheel rolling around some StaticBody2D polygon platforms. It has very intuitive physics, and is coded very cleanly in all C#.

wheely

-----

It has only been tested on my Ubuntu system with Godot 3.X, but I don't see why it wouldn't work on other systems. It is licensed MIT so you may do with it whatever you please.


mono/Wheely/LICENSE.md
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2018 David Vitale
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
```
aaronfranke: I don't think we are doing this. See #314

DavidJVitale:
@aaronfranke I have addressed your comments on and submitted a new PR, please review here: #334 

(Since I accidentally deleted my fork there is nothing I can do about this PR, it will remain an orphan)

aaronfranke:
Closing as a duplicate of #334 (which might also be closed, not sure yet).

PR #260, opened 2018-05-18
Use an higher FOV value in demos

Calinou:
A vertical FOV of 74 degrees is roughly equivalent to a 90 degree horizontal FOV on a 4:3 display (~106.26 degrees on 16:9), which is the typical default FOV used in PC games.

Some camera distances were decreased to account for the higher FOV.


NathanLovato:
Thanks :)

PR #262, opened 2018-06-22
Fix finite state demo warning

goodyttoor:
Not effect runtime, just remove some warning spam in editor output.

- Remove autoload non existence script
- Not running state stack display in editor


akien-mga:
CC @NathanLovato

NathanLovato:
Thanks!

Issue #263, opened 2018-07-02
Turn-based combat/RPG demo


NathanLovato:
We're working on a turn-based RPG tutorial series with @henriiquecampos.

Henrique made a JRPG demo already, just wanted to give everyone a heads-up! 

![discord_2018-07-02_10-38-59](https://user-images.githubusercontent.com/12694995/42141438-305f725c-7de4-11e8-90dc-96c32fd7420d.png)

I think it was requested on Patreon, so that would answer the patrons' request. Also, it's different from #61 as the other issue mentions it's not about combat.

Right now the little demo includes:

1. Top-down grid-based movement (a little like RPG maker)
2. A simple dialogue system
3. Transition in and out of combat
4. Turn-based combat

henriiquecampos:
Hey, thanks a lot for the heads-up! 

Just passing by to leave a small showcase .gif of it :blue_heart:

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/10171059/42184075-8aad870e-7e1a-11e8-9548-30c6b39f3c02.gif)

It shows quite useful approaches to these system, specially for combat as it uses `PackedScene` to instantiate the combat actors, so is very simple to design a combat :smile:

jaggygames:
Wow this is amazing! Thank you!

Zeph-dev:
That looks great! Really looking forward to it, do you think this could be modified to act like a Pokémon-like game? That would be really cool!

henriiquecampos:
@Zeph-dev Pokémon was the reference actually, so yeah, totally :smile:

jaggygames:
@Zeph-dev Any ETA on this? :heart_eyes:

henriiquecampos:
@jaggygames well it's up to the guys managing the repo :thinking: it is as shown in the gif, but the code may need tweaks to keep it as standard as the other demos. So it will depend on:

1. If and when it will be accepted
1. The time between the PR and reviewing it
1. The amount of changes it will take to make it acceptable (lots lol)
1. Me applying said changes :stuck_out_tongue:

Zeph-dev:
@henriiquecampos awesome!

Zeph-dev:
@henriiquecampos so can you have more than one character to fight with, or be able to switch between characters? It would be awesome for that to happen because then people can make Pokémon fan-games, or their own JRPGs. It's awesome to see what you've done already, hope you continue it.

henriiquecampos:
[...]

PR #267, opened 2018-07-27
Add JRPGish demo

henriiquecampos:
Well, as suggested in #263 I'm making the PR for a small JRPG demo currently it works as follow:

- The game has 2 states: **Exploration** and **Combat** which are two screens that are removed/added to the SceneTree depending on which state the game is currently
- Exploration is a basic topdown grid movement system, character can walk, bump on obstacles and if it bumps on an opponent, a dialog will show up, when dialog finishes the combat is built.
- **Actor** (you will see this a lot in the demo) is an agent of a system.
- The Actors for the exploration also carry a **Combat Actor** `PackedScene` variable which is exposed in the inspector, so to build a combat you can design a new enemy, drag and drop into the Exploration `Actor.combat_actor` slot in the inspector and you are done
- As for the Combat, it just takes an array of **Combat Actors** to build the combat it is currently not very scalable, since the interface has space for just 2 **Actors**, but nothing actually prevents from having more, it will only seem very ugly
- The combat is take on turns with targets being fixed and automated
- When someone dies a signal is emitted, the Combat finishes, going back to Exploration and a Dialog is shown depending on who won the combat.

Most of the systems were made with mentoring of @NathanLovato but since I ducktaped them together by myself, it will probably be a mess and will need some hard reviewing, so don't take easy on pointing bad practices.

That said, thanks a lot for the opportunity, hope this helps some peeps :blue_heart:


aztecsensei:
this is great! godot need more game demos.
base on your experience,how much difficult would it be to make a demo that plays like Final fantasy 7?
im talking about the programming side. of course. :)

NathanLovato:
I'll review it after I pushed/scheduled the next Godot course update and batch of tuts.

NathanLovato:
Some general issues to fix:
- missing last lines on some script files
- when you talk to the opponent multiple times, you try to connect the signals again and get an error

image

- Deffend -> defend (and it feels this button + method should be `heal` or something?)

image

Zeph-dev:
Could be a link to the items, or where healing, catching items are stored perhaps? Or do you already have a button for that?

henriiquecampos:
Hey there finally was able to update this :stuck_out_tongue: 

> and it feels this button + method should be heal or something?)

As I'm using [Health.gd](https://github.com/godotengine/godot-demo-projects/pull/267/files#diff-288b61e17803ca41107d4b6c67c188d8R8) to calc the amount of damage the character takes, when using **Defend** it increases the `Health.armor`, decreasing the total damage, no healing there. :stuck_out_tongue_closed_eyes:

NathanLovato:
Okay there's some typos, and the inventory button does nothing. Aside from that it works well. Let me merge and fix that

PR #270, opened 2018-08-08
Split screen platformer for godot 3

### LINKED ISSUE #224: Need to port split screen demo
It looks like the split screen demo has not been written for 3.0.
Related QA thread: https://godotengine.org/qa/25696/split-screen-in-godot-3

markopolojorgensen:
There are quite a few questions in the QA about splitting the screen in godot 3. Hopefully this demo will answer some of them.

I assume this is the same approach used to create this demo for 2.1: copy the platformer demo, add split screen.

If the platformer demo changes significantly, this demo may fall behind. We could instead have a dedicated demo that just shows split screen functionality that didn't depend on the platformer...?

fixes #224 

screenshot from 2018-08-08 11-09-32


cbscribe:
I was planning on adapting this: http://kidscancode.org/blog/2018/07/godot3_splitscreen/

It keeps the game very simple and focuses on the splitscreen setup.

aaronfranke:
@markopolojorgensen Perhaps it would be better to add a second scene to the platformer demo for splitscreen. It shouldn't take too many modifications to allow it to work with either mode.

aaronfranke:
I'm going to close this PR because I think there are better approaches to solve this. We could either add a new scene to the platformer demo, or have a simpler demo dedicated to splitscreen. Feel free to continue the discussion below.

PR #284, opened 2018-09-24
Fix the git tracking issue with Grid.gd and Pawn.gd on Windows

LikeLakers2:
This fixes an issue introduced with 2c260e010399e045f11f7fa485ca850319226dce, where git tries to track two capitalizations of the Grid.gd and Pawn.gd files. Git allows tracking of multiple files with different capitalizations but otherwise the same name -- however, Windows doesn't like that. It only wants one file with a name, regardless of capitalization.

Since I imagine there's other users using Windows on this repository, I've gone ahead and done the small bit of work to get the conflicting names removed.

Unfortunately I'm not able to merge these two commits, as I'm on Windows. I tried that already before making this PR, and it didn't work, as git once again thought it should try to track both the Pawn.gd and pawn.gd files. Sorry!


NathanLovato:
Thanks :)

NathanLovato:
Thanks

PR #286, opened 2018-10-02
Improve Godette sprite in sprite_shaders demo

williamd1k0:
The current sprite is quite aliased and with some artifacts (see the rich diff).

*Actually it would be better if there were an official source with transparent background :/*


fire:
[godotea.zip](https://github.com/godotengine/godot-demo-projects/files/2454466/godotea.zip)

Did a quick and dirty vector convert. It's very rough.

akien-mga:
Thanks!

PR #291, opened 2018-12-02
Improved the custom_dock plugin by making the button work

rhulha:
I think supplying a custom_dock plugin demo where one of the buttons actually does something is more informative and helpful for beginners like me.

Changes:
Added a simple little script for the custom_dock button.
Implemented the button pressed event.
The button press now executes: print("Plugin Button Pressed!")


bojidar-bg:
I think you should remove the default comments in the script.

rhulha:
> I think you should remove the default comments in the script.

I think they are helpful to beginners.

rhulha:
> If this were trying to demonstrate Godot's scripting language as a whole, then I would agree. But this isn't serving that purpose.
Fair enough.
> 
> Also, you should save that scene as a `.tscn` file for the purposes of this repo.
I did not change the extension of the scene.
I just changed the already existing scene file.

plugins/custom_dock/button.gd
```diff
@@ -0,0 +1,20 @@
...
+
+func _ready():
+	# Called when the node is added to the scene for the first time.
+	# Initialization here
+	pass
+
+#func _process(delta):
+#	# Called every frame. Delta is time since last frame.
+#	# Update game logic here.
+#	pass
```
aaronfranke: Remove all empty functions and default comments.

PR #292, opened 2018-12-04
Fix viewport_quad orientation

noidexe:
Rotated the viewport quad so the pong game is rendered correctly. It was rotated 180 degrees.


aaronfranke:
This will need to be [rebased](https://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html) before it can be merged.

PR #298, opened 2019-01-05
Fixed Finite State Machine error with enums

Rubonnek:
Not sure if this is wanted, but this pull request fixes the FSM demo

I was not able to run this demo using the latest Godot changes.


NathanLovato:
Yup enum members don't register as individual constants anymore in 3.1, thanks for the fix!

PR #301, opened 2019-01-28
Updated "Grid-based pathfinding with Astar" to version 3.1

Toshiwoz:
I wanted to see how path finding work and I noticed that didn't work with 3.1, and it didn't work as it was.
The changes I made here allowed me to correctly test the project.
Not sure if I had to exclude some files in the commit. Neither if you're actually going to merge this untill 3.1 is out.


akien-mga:
Thanks!

Toshiwoz:
I also tried with bake navmesh, and after some tweaking it worked, still, it doesn't look 100% right, it actually looks like it's flipped upside-down, even if it works.
image
View from beneath.

image
Climbing too much.

So I think I shouldn't add that untill I figure out how to configure correctly the navmesh parameters.

PR #302, opened 2019-02-06
Automatic brake/reverse

kkmzero:
Value fwd_mps converts global velocity into a velocity vector which is rotated to cars local coordinate system (in this example x-axis).
Then if key "ui_down" is pressed and the fwd_mps value is greater or equal to -1 (which seems to be very good spot and represents car still moving forward but nearly standing still) the car reverses (negative engine_force_value). If this condition isn't met (which means the car is moving forward) then the car brakes.

Tested and working on Godot 3.1 Beta 3.

Suggestions:
In this case, if you want something more realistic, the -engine_force_value can be much lower because no car reverses in same speed as it goes forward but for the sake of demo project I left it as it is.

Created for my own project with help of user wombatstampede from godotdevelopers.org/forum

### DIFF
--- a/3d/truck_town/vehicle.gd
+++ b/3d/truck_town/vehicle.gd
@@ -10,6 +10,8 @@ var steer_target = 0
 export var engine_force_value = 40
 
 func _physics_process(delta):
+	var fwd_mps = transform.basis.xform_inv(linear_velocity).x
+	
 	if Input.is_action_pressed("ui_left"):
 		steer_target = STEER_LIMIT
 	elif Input.is_action_pressed("ui_right"):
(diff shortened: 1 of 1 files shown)


aaronfranke:
While there are no conflicts, the project has been updated since, so you should rebase and test.

However, the change works quite well in my tests, reversing indeed works.

kkmzero:
> While there are no conflicts, the project has been updated since, so you should rebase and test.
> 
> However, the change works quite well in my tests, reversing indeed works.

I tested it in Godot Engine 3.1.1 stable (current latest official version) and the current (master branch) version of godot-demo-projects - it is working same as before. There are no differences I can see between current version of the script and my commit, even though I'm aware of code cleanup done on Mar 14, 2018. If I edit my commit now and replace it with the code of the current script I get "There are no changes to show.", so they are obviously identical. I don't quite understand how that happened.

aaronfranke:
@akien-mga This should be good to merge. This is for the Truck Town demo.

akien-mga:
Thanks, and sorry for the delay :)

PR #312, opened 2019-02-28
2d/physics_platformer works flawlessly on Godot 3.1 beta 8

TestDotCom:
* fix warnings for inferred node type
* every script has a class_name
* formatted code for better clarity
* each files has its own subfolder
* bullet collisions and shooting have to be executed from call_deferred


akien-mga:
Thanks!

PR #319, opened 2019-03-23
Updated rotation of Quad to be right side up

wscalf:
I pulled down the samples and started playing with the 3D GUI and..noticed the GUI was upside down. It was pretty easy to flip it upright, so I thought I'd go ahead and do that and make a pull request. If this isn't the appropriate fix, please advise.


akien-mga:
Thanks!

PR #322, opened 2019-04-08
Fix navigation demo not reaching the final point

### LINKED ISSUE #321: navigation demo character doesn't quite reach destination
in the [navigation demo](https://github.com/godotengine/godot-demo-projects/tree/master/2d/navigation) the character doesn't quite reach its destination.

This can be seen by adding

`print('$Character.position ', $Character.position, ' last_point ', last_point)`

to the `move_along_path()` function, **within the `for loop` and after** `path.remove(0)`

You'll see the character is always a few pixels short of the destination.
I would expect `$Character.position` to match `last_point` exactly, or am I missing something?

bojidar-bg:
Fixes #321


akien-mga:
Thanks!

PR #323, opened 2019-04-22
Make the Platformer 2D demo work with ultrawide displays

Calinou:
This makes it possible to fill the whole viewport when playing on a 21:9 monitor or a 19:9 smartphone.

## Preview

image


2d/platformer/player/Player.tscn
```diff
@@ -247,17 +247,24 @@ passby_press = true
 action = "move_right"
 visibility_mode = 1
 
-[node name="Jump" type="TouchScreenButton" parent="UI"]
-position = Vector2( 666.224, 359.02 )
+[node name="BottomRightAnchor" type="Control" parent="UI"]
```
Calinou: I had to make them child of a Control so the relevant TouchScreenButtons can be anchored to the bottom-right corner.

aaronfranke:
The sky and background need to be extended. Right now it's easy to see the edge of the sky.

Screenshot from 2019-06-06 00-25-04

Calinou:
@aaronfranke Should be fixed now. To do this, I had to tweak how the parallax platforms were drawn though; instead of relying on ParallaxLayer's Mirroring property (which can only mirror something once), I used TextureRegion to make the texture repeat several times. I don't know if this has an adverse effect on performance, especially on mobile.

Calinou:
@aaronfranke This would require the Mirroring property to support repeating a texture more than once, I guess. This way, I could drop the repeating texture hack.

PR #326, opened 2019-05-05
Adding OpenSimplexNoise Viewer demo

DevinPentecost:
Adding a simple demo for users to look at and play with the OpenSimplexNoise parameters.
It has a few simple controls that will manipulate the noise parameters, which is then rendered to the screen.

There is also a Min/Max clip option, which will force colors above/below to be White or Black, depending on if they should be clipped. This functionality is implemented in a simple shader to significantly reduce draw time.


DevinPentecost:
Doesn't the `NoiseTexture` object take an `OpenSimplexNoise` object anyways? I'm guessing it does the same thing?
I wanted to stick with just one object type, the `OpenSimplexNoise`, since that's the focus of the example. However, if there are significant performance improvements with the `NoiseTexture` instead of feeding the `Image` into the `ImageTexture` that could be trivially implemented. I also like that it acts as a 'mini tutorial' of the difference between an `Image` and an `ImageTexture` and how to use the two together.

JFonS:
@DevinPentecost Yes, `NoiseTexture` takes the reference to an `OpenSimplexNoise` resource and updates the texture data automatically when any of the noise parameters changes. That would reduce the amount of code in this demo, but I don't mind leaving it as is. 

My only concern is that people taking this tutorial as an example might overlook the existance of `NoiseTexture`. So we could add a comment mentioning it as an alternative maybe?

DevinPentecost:
I've added the NoiseTexture as an example so users can see how to utilize both.
A flag can be set in the script to switch from one to the other.

Thanks,
-Devin

akien-mga:
Thanks!

PR #330, opened 2019-06-03
Update and optimize 3D IK Demo

aaronfranke:
A lot of small things changed, but here's the basics of it:

* Finish update to 3.1 and fix warnings
* Fix being able to lean through walls in FPS example
* Move FPS example into a subfolder for organization
* Some code cleanup and modernization in `ik_fabrik.gd` (see commit)


aaronfranke:
Note: This demo doesn't work properly on Godot 3.2 (with or without this pull request), so this will have to be fixed after 3.2 is released.

PR #333, opened 2019-06-06
Remove "background load" demo from the 3.0 branch

### LINKED ISSUE #266: background_load failed to load
Godot version: 3.0.5 official stable

I just downloaded the "godot-demo-projects" zip and the editor doesn't want to load the "misc/background_load" project.

Here's the console's error:
```
ERROR: _load_settings_text: Condition ' config_version > 3 ' is true. returned: ERR_FILE_CANT_OPEN
   At: core/project_settings.cpp:495
```

I managed to load the project by manually editing the "project.godot" file and replace the
`config_version=4`
with
`config_version=3`

However, you might want to fix the problem in the official release of the demos.
Cheers,
Radu

aaronfranke:
This was created by reduz in 3.1 dev before alpha. The simplest solution is to require using Godot 3.1 to use this demo (and therefore removing it from the 3.0 branch). I don't think it's worth trying to backport this demo just because it was added to the repo before, it might not behave as expected. 

Closes #266 @akien-mga


akien-mga:
Thanks!

PR #334, opened 2019-06-10
Added "Wheely" C# Mono Project

DavidJVitale:
Continuing from #259 

The "mono" folder could use some more high quality examples. I have created a simple example called "Wheely". Wheely is a wheel-based platformer with a KinematicBody2D Wheel rolling around some StaticBody2D polygon platforms. It has very intuitive physics, and is coded very cleanly in all C#.

wheely

----

I have addressed @aaronfranke 's comments from the previous PR. This PR uses Godot 3.1 Mono, the LICENSE.md file was removed, etc. Let me know how else I can help!


mono/Wheely/src/Player.cs
```diff
@@ -0,0 +1,106 @@
...
+    ///     - The Player will follow the curve when it should "fly away" from it
+    /// The solution to this is to make the angle a bit "above" the traditional "forward".
+    ///     - "above" is defined as the direction towards "Normal" (away from the platform)
+    ///     - "above" is dependant the direction player is moving (AKA the angle of velocity)
+    private Vector2 calculateForwardAngle(){
+        //Calculate the angle above foreward
+        const float angleAbovePercent = 0.20f;
+        var normalAngle = this.currentNormal.Normalized();
+        var currentVelocityAngle = this.currentTravel.Normalized();
+        var angleAbove = angleAbovePercent * currentVelocityAngle.AngleTo(normalAngle);
```
aaronfranke: Excessive variable declarations. Instead of `var normalAngle = this.currentNormal.Normalized();` what you should do is change `this.currentNormal = collision.Normal;` so that you can just call `currentNormal` in place of `normalAngle`.

mono/Wheely/src/Player.cs
```diff
@@ -0,0 +1,106 @@
...
+        //Apply the previously calculate "angleAbove" to the forward angle 
+        var unadjustedForwardAngle = normalAngle.Rotated((float)Math.PI / 2f).Normalized();
+        var adjustedForwardAngle = unadjustedForwardAngle.Rotated(angleAbove);
```
aaronfranke: Excessive use of `var`.

mono/Wheely/src/Player.cs
```diff
@@ -0,0 +1,106 @@
...
+        else if (Input.IsActionPressed("ui_right") &&
+        (this.forwardAccell <= MAX_FORWARD_ACCEL)) {       
+            this.forwardAccell += FORWARD_ACCEL_UNIT;
+        }
+        else{
+            this.forwardAccell *= DECELL_EFFECT;
+        }
+    }
+
+    private void processPhysics(float delta){
```
aaronfranke: Method names should be PascalCase, as per the [style guide](https://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html).

mono/Wheely/src/VisiblePlatform.cs
```diff
@@ -0,0 +1,18 @@
...
+using System;
+
+public class VisiblePlatform : CollisionPolygon2D{
+    /// Using this class makes the Polygon2D child visible in the game
+    
+    public override void _Ready(){
+        for(int i=0; i<this.GetChildCount(); i++){
+            var child = this.GetChild(i);
+            if(child is Polygon2D){
+                Polygon2D visiblePolygon = (Polygon2D)GetChild(0);
```
aaronfranke: I don't think this is intentional. You are iterating through the children and then you don't do anything with the results of what you've found, you just always manipulate the 0-th child.

aaronfranke:
I'm going to close this PR because:

* It has been over 6 months with no update after the changes were requested.

* This demo doesn't really showcase any particular Godot features or how a specific type of game would be implemented in Godot.

* I don't think if it's worth having demos exclusively in C# when we could instead port (a few) GDScript demos which would provide a comparison between C# and GDScript.

Feel free to continue the discussion below, but I highly doubt this will be added to the repo.

PR #335, opened 2019-06-14
Add two WebRTC demo

Faless:
`webrtc_minimal` contains two examples of a minimal peer connection, and local signaling server.

`webrtc_signaling` contains a reference signaling server with lobby support for both GDScript and Node.js , a "raw" GDScript client, an helper to get the `WebRTCMultiplayer` setup (godotengine/godot#29755), and a usage demo.


akien-mga:
Thanks!

PR #337, opened 2019-07-05
Extract shader code from 2D shader demos

robspsj:
I was looking at the shader examples to explain them to a friend and we had some dificulty navigating and comparing the different shaders in the examples, and having each shader in it's individual file instead of internal to the scene helped a lot.

### DIFF
--- a/2d/screen_space_shaders/shaders/blur.shader
+++ b/2d/screen_space_shaders/shaders/blur.shader
@@ -1,0 +1,8 @@
+shader_type canvas_item;
+
+uniform float amount : hint_range(0.0, 5.0);
+
+void fragment() {
+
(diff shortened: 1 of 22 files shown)


2d/screen_space_shaders/shaders/blur.shader
```diff
@@ -0,0 +1,8 @@
+shader_type canvas_item;
+
+uniform float amount : hint_range(0.0, 5.0);
+
+void fragment() {
+
+	COLOR.rgb = textureLod(SCREEN_TEXTURE, SCREEN_UV, amount).rgb;
+}
```
aaronfranke: Please ensure there are newlines at the ends of each file. Many other files in your PR are like this, you can see which by the red error symbol in GitHub.

robspsj: sorry for the time it took to correct this, the files should be ok now :)

PR #339, opened 2019-07-10
Use p2p in multiplayer bomber player exchange.

Faless:
We used to broadcast player info on connect and have the server relay it
to other clients.
With this approach, each peer (including server) sends its own info once
to other peers as soon as they connects.
When a new player connects, it is notified of all the already connected
peers by the `network_peer_connected` signal.
Any already connected peer is also notified of the newly connected peer
by the same signal.

References godotengine/godot#18436


akien-mga:
Thanks!

PR #348, opened 2019-08-21
Fix seesaw falling plank from 2d demo

texaco:
Seesaw did stand up in 2d Demo - Physics Platformer.

Add a pilar as static body and reasign PinJoin2D nodes


bojidar-bg:
Typo: Pillar is with double-L.

Is the pillar necessary, or does reassigning the PinJoint already fix the problem?

texaco:
@bojidar-bg It is needed. Thanks for spotting the typo

bojidar-bg:
@texaco Would be nice to remove the collision shape of the pillar or make it lower, since currently enemies are getting stuck on it (going to do it myself in some minutes, in case you aren't around).

PR #350, opened 2019-09-01
Update Audio demos for Godot 3.1.1

aaronfranke:
A lot of this is just style and typo fixes, but there are also some minor optimizations and renamings, as well as fixing warnings. Also, 3 of the 4 projects were lacking a `config/name`, so I added them.

Note: The BPM Sync demo requires Godot 3.2. I still made some improvements to it, but it doesn't actually work in 3.1.1. We can remove the BPM Sync demo in the 3.1 branch after 3.2's release.

### DIFF
--- a/audio/bpm_sync/Control.gd
+++ b/audio/bpm_sync/Control.gd
@@ -12,48 +12,46 @@ const SYNC_SOURCE_SOUND_CLOCK = 1
 
 var sync_source = SYNC_SOURCE_SYSTEM_CLOCK
 
+# Used by system clock.
+var time_begin
+var time_delay
+
 
 func strsec(secs):
 	var s = str(secs)
-	if (s.length()==1):
-		s="0"+s
+	if (secs < 10):
+		s = "0" + s
 	return s
-	
-# used by system clock
-var time_begin 
-var time_delay
 
-func _process(delta):
+
+func _process(_delta):
 	if (!playing or !$Player.playing):
 		return
 	
-	var time 
+	var time = 0.0
 	if (sync_source == SYNC_SOURCE_SYSTEM_CLOCK): 
-		# obtain from ticks
+		# Obtain from ticks.
 		time = (OS.get_ticks_usec() - time_begin) / 1000000.0
-		# compensate
-		time -= time_delay		
-	elif (sync_source == SYNC_SOURCE_SOUND_CLOCK):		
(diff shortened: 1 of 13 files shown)


audio/bpm_sync/Control.gd
```diff
@@ -3,48 +3,46 @@ extends Panel
 const BPM = 116
 const BARS = 4
 
-var playing = false
-const COMPENSATE_FRAMES = 2
-const COMPENSATE_HZ = 60.0
+var playing := false
```
akien-mga: I'm don't think we reached a consensus that optional typing should be used in official demos or the documentation. Last we discussed it we preferred not to use it, unless strongly relevant, IIRC.

aaronfranke: I've removed these then (except for one place in spectrum where it's very relevant), but it would be great to have a discussion about this; I don't see anything if I search for "is:issue typing".

audio/generator/generator_demo.gd
```diff
@@ -1,28 +1,28 @@
 extends Node
 
-
-var hz = 22050.0 # less samples to mix, GDScript is not super fast for this
+var sample_hz = 22050.0 # Keep the samples to mix low, GDScript is not super fast.
```
follower: Suggestion: "Keep the samples..." -> "Keep the number of samples..." 

This wording seems clearer to me and more accurately represents original wording. ("mix low" could be misunderstood as meaning low in audible volume.)

audio/generator/generator_demo.gd
```diff
@@ -1,28 +1,28 @@
...
 
-
-var hz = 22050.0 # less samples to mix, GDScript is not super fast for this
+var sample_hz = 22050.0 # Keep the number of samples to mix low, GDScript is not super fast.
+var pulse_hz = 440.0
 var phase = 0.0
 
-var pulse_hz = 440.0
-var playback = null #object that does the actual playback
+var playback: AudioStreamPlayback = null # Actual playback stream, assigned in _ready().
```
yuvallanger: Could it be changed to `onready var playback: AudioStreamPlayback = $Player.get_stream_playback() # Actual playback stream, assigned in _ready().` and then remove the same line in `_ready()`?

aaronfranke: It couldn't, because this has to be assigned after `$Player.stream.mix_rate = sample_hz` or else it won't work.

PR #353, opened 2019-09-23
Fixed casting from Basis to Quat

hbina:
Noticed a lot of error spat out because of non-normalized basis.
I simply followed the suggestion to use `get_rotation_quat()` instead of using constructor of Quat(x).


akien-mga:
Thanks!

PR #356, opened 2019-10-08
Add Ball Factory

henriiquecampos:
So, since this demo project showcases how instancing works I thought that it could be a good idea to also show how we can create instances in runtime. I also took rid of the `container.png` and its respective sprite node because we can pretty much achieve it with polygon2d, which is cool.


henriiquecampos:
Force pushed to squash a minor change

misc/instancing/container.tscn
```diff
@@ -35,16 +35,18 @@ bounce = 0.4
 
 [node name="container" type="Node"]
 
-[node name="static" type="StaticBody2D" parent="."]
+[node name="ball_factory" type="Position2D" parent="."]
```
aaronfranke: Why is it of type `Position2D`? It should either be `Node` or `Node2D`. Or, even better, put the script on the scene's root.

misc/instancing/ball_factory.gd
```diff
@@ -0,0 +1,16 @@
...
+
+func _unhandled_input(event):
+	if event.is_echo():
+		return
+	if event is InputEventMouseButton and event.is_pressed():
+		if event.button_index == BUTTON_LEFT:
+			spawn(get_global_mouse_position())
+
+
+func spawn(spawn_global_position = global_position):
```
aaronfranke: Assuming the script is placed on a `Node`, you can just make this `Vector2.ZERO` if we want a default value, but really it doesn't need to have a default value.

aaronfranke:
This is overall an improvement, so I'll merge this, even if it's not perfect. Thanks!

PR #357, opened 2019-10-08
Improve Pause demo

henriiquecampos:
In this demo, the focus is on the pausing/unpausing of the SceneTree, so I removed the signal connections and popup thingies to ensure that the only thing that should pick the attention is how to pause/unpause a game in Godot. I also explain how to ensure that there is a safe node that will still process and allow for the scene to unpause.

### DIFF
--- a/misc/pause/pause_button.gd
+++ b/misc/pause/pause_button.gd
@@ -1,0 +1,18 @@
+extends Button
+
+func _ready():
+	#This ensures that this Node won't be paused, allowing it to
+	#process even when the SceneTree is paused. Without that it would
+	#not be able to unpause the game. Note that you can set this through
+	#the inspector as well
+	pause_mode = Node.PAUSE_MODE_PROCESS
+
+
+func _toggled(button_pressed):
+	#Pause or unpause the SceneTree based on whether the button is
+	#toggled on or off
+	get_tree().paused = button_pressed
+	if button_pressed:
+		text = "Unpause"
+	else:
+		text = "Pause"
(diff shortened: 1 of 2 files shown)


misc/pause/pause_button.gd
```diff
@@ -0,0 +1,18 @@
+extends Button
+
+func _ready():
+	#This ensures that this Node won't be paused, allowing it to
+	#process even when the SceneTree is paused. Without that it would
+	#not be able to unpause the game. Note that you can set this through
+	#the inspector as well
+	pause_mode = Node.PAUSE_MODE_PROCESS
```
aaronfranke: This isn't necessary to have in code, it's already set as a property of the node. The project still works fine if `_ready` is deleted.

misc/pause/pause_button.gd
```diff
@@ -0,0 +1,18 @@
...
+
+
+func _toggled(button_pressed):
+	#Pause or unpause the SceneTree based on whether the button is
+	#toggled on or off
```
aaronfranke: Comments should start with spaces

misc/pause/spinpause.tscn
```diff
@@ -1,6 +1,6 @@
 [gd_scene load_steps=4 format=2]
 
-[ext_resource path="res://spinpause.gd" type="Script" id=1]
```
aaronfranke: This script isn't being used anywhere anymore, it should be deleted.

aaronfranke:
This is overall an improvement, so I'll merge this, even if it's not perfect. Thanks!

PR #358, opened 2019-10-08
Revamp Input Remap demo simplifying it

henriiquecampos:
So, this demo looked over-complicated, working with things that sounded like beyond its scope. I tried to simplify it a lot by focusing strictly on Input Remapping, removing saving and loading of defaults within other things that, seemed to me, unnecessary for the purpose of the demo, focusing the goal of the demo in a simpler method, the [`remap_action_to(event)`](https://github.com/godotengine/godot-demo-projects/compare/master...pigdevstudio:input-remap-revamp?expand=1#diff-a2b0e71b69a604a9c7521eb290a61103R25) method


gui/input_mapping/InputRemapMenu.tscn
```diff
@@ -0,0 +1,151 @@
...
+anchor_right = 1.0
+anchor_bottom = 1.0
+
+[node name="Instructions" type="Label" parent="."]
+margin_left = 20.0
+margin_top = 20.0
+margin_right = 620.0
+margin_bottom = 34.0
+text = "Click on a button to reassign its action key.
+ Press \"Return\" or click on the button again to confirm."
```
aaronfranke: This is worse behavior than the demo previously had. Now it's easy to accidentally click on multiple things at once, and it's not possible to bind the Enter/Return key. It would be much better if it had the old behavior where clicking a key assigns *and* finishes the assigning.

aaronfranke:
@henriiquecampos Any update? This isn't good to merge as-is due to the issues above.

henriiquecampos:
Ohh, I completely lost track of these.

Yeah I will fix that, the way buttons toggle based on ui_accept makes this implementation less effective. 

I'll fix it

henriiquecampos:
@aaronfranke can you test it now? Now to toggle the remapping you use <kbd>RMB</kbd>

Enter, Escape etc... can be assigned correctly now :grin:

gui/input_mapping/ActionRemapButton.tscn
```diff
@@ -0,0 +1,11 @@
...
+
+[ext_resource path="res://ActionRemapButton.gd" type="Script" id=1]
+
+[node name="ActionRemapButton" type="Button"]
+margin_right = 90.0
+margin_bottom = 30.0
+focus_mode = 1
+toggle_mode = true
+enabled_focus_mode = 1
+script = ExtResource( 1 )
```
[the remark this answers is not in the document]
henriiquecampos: Basically it makes it easy to reuse. The ActionRemapButton can be instanced easier and makes it explicitly self contained. When things are not independent scenes I usually read them as being dependencies of the scene `owner` and relying on it for their implementation.

It is a good practice I follow, but if you'd rather take rid of it, I can delete the scene and make the instance just straight Buttons with a `ActionRemapButton.gd` attached to them.

aaronfranke:
Sorry, but I just don't think this PR is an improvement in any way except the sizing of buttons.

* I much prefer the existing behavior of pressing finishes assigning. Any button for finishing assignment does not feel intuitive to me as someone who's played PC games for a decade.

* [Too easy to get into unintended situations of rebinding multiple keys by accident.](https://user-images.githubusercontent.com/1646875/74295249-7e2d4b80-4d0e-11ea-98dc-7d7c56f38169.png)

* I don't think making a scene just for increasing reusability of one node is a good idea.

* The rewrite of the script doesn't contain the save/load functionality that the current script does. Key bindings are lost when the project restarts.

* There is a style issue in the script anyway (comments should start with spaces).

* You didn't even remove the old files...

I don't think this will ever be merged due to the top issue, so I'll close this.

henriiquecampos:
I disagree, this PR is a good improvement for this demo, especially because it provides a much cleaner, simple and easy to understand code.

- Godot Input Mapping behavior itself doesn't comply with your suggestion. In Godot the input prompt keep registering the new input until the user either click on OK or Cancel. This is consistent in many other tools and games. In either case, it is fixed in the latest commit
- This is fixed in the latest commit
- There is no decent reason why save/load the settings is relevant for an **input remap** demo, the only relevant piece of code is the remapping itself, present in [ActionRemapButton.gd L20-L29](https://github.com/pigdevstudio/godot-demo-projects/blob/ded3a089616f1ae8cd555be0cda53bba529c52c3/gui/input_mapping/ActionRemapButton.gd#L20-L29). Following your reasoning it be could argued that save/load is something you would want to have in literally every other demo in this repository, but I think it would only be relevant in a Save/Load specific demo
- Fixed in latest commit
- Fixed in latest commit, they were present because a conflict due to 84c8d6ef693efbd8e071ed641ff14fffb61e123e

bojidar-bg:
> There is no decent reason why save/load the settings is relevant for an input remap demo [...], but I think it would only be relevant in a Save/Load specific demo

Well, a demo showing saving and loading of settings should be added in such a case.

I would still argue that saving and loading of input map actions is pretty complicated, as you have to get the proper InputEvent and somehow serialize/deserialize it.
Perhaps it should be made it a simpler and less complicated manner than the current demo, however.

Just my 2 cents.

gui/input_mapping/ActionRemapButton.gd
```diff
@@ -0,0 +1,35 @@
...
+	if button_pressed:
+		text = "... Key"
+		release_focus()
+	else:
+		display_current_key()
+
+
+func _unhandled_key_input(event):
+	# Note that you can use the _input callback instead, especially if
+	#you want to work with gamepads
```
aaronfranke: Missing space and period.

gui/input_mapping/InputRemapMenu.tscn
```diff
@@ -0,0 +1,174 @@
...
+[node name="InputRemapMenu" type="Control"]
+anchor_right = 1.0
+anchor_bottom = 1.0
+
+[node name="Instructions" type="Label" parent="."]
+anchor_right = 1.0
+margin_top = 24.0
+margin_bottom = 55.0
+text = "Click on a button to reassign its action key.
+Click on the button again to confirm the remapping."
```
aaronfranke: The 2nd line of this text is not relevant anymore.

henriiquecampos: Fixed.

gui/input_mapping/ActionRemapButton.gd
henriiquecampos: Fixed.

aaronfranke:
Alright, I think I'll just go ahead and merge this then. Perhaps saving and loading isn't super important for this demo, you've done a lot of work on this demo, and I don't want this to be in limbo forever.

PR #369, opened 2020-01-01
Add Look towards moving direction

nonunknown:
the look_system itself was developed by me, but lerp_angle and short_angle i've got on the internet forums

I think this will help a lot of people to get their job done fast!!!

needs testing - and the character need to have a kind of "nose" to its model, just to simplify visualization

### DIFF
--- a/3d/kinematic_character/cubio.gd
+++ b/3d/kinematic_character/cubio.gd
@@ -10,9 +10,9 @@ const ACCEL= 2
 const DEACCEL= 4
 const MAX_SLOPE_ANGLE = 30
 
+var dir = Vector3() # Where does the player intend to walk to
 
 func _physics_process(delta):
-	var dir = Vector3() # Where does the player intend to walk to
 	var cam_xform = $target/camera.get_global_transform()
 
 	if Input.is_action_pressed("move_forward"):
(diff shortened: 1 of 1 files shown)


3d/kinematic_character/cubio.gd
```diff
@@ -48,7 +48,26 @@ func _physics_process(delta):
...
+		
+	look_system() # System to look at direction that player is moving
 
+var new_dir:Vector3 = Vector3.ZERO
+var new_angle:float = 0
+func look_system():
+	if (dir != Vector3.ZERO):
+		new_dir = dir
+	var angle:float = atan2(new_dir.x,new_dir.z) #get the angle from the player's input ie: 1,1.5,3
+	new_angle = lerp_angle(new_angle,angle,.25) # smooth lerp to the new angle (using custom method to avoid weirdo rotations)
```
aaronfranke: As per the [style guide](https://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html), use spaces after commas, and start comments with spaces.

3d/kinematic_character/cubio.gd
```diff
@@ -48,7 +48,26 @@ func _physics_process(delta):
...
+	var deg = rad2deg(new_angle) #convert the angle to degrees ie: 360,180,0,270
+	rotation_degrees.y = deg #apply to kinecnatic player's rotation
+
+func lerp_angle(from, to, weight):
+	return from + short_angle_dist(from, to) * weight
+
+func short_angle_dist(from, to):
+	var max_angle = PI * 2
+	var difference = fmod(to - from, max_angle)
+	return fmod(2 * difference, max_angle) - difference
```
aaronfranke: This function can be replaced with `wrapf(to - from, -PI, PI)` if I'm not mistaken.

But also, `lerp_angle` was added in core in 3.2, so this will not be necessary at all in the near future.

aaronfranke:
@nonunknown Any update? Godot 3.2 has been released, so we should use the built-in lerp angle function.

aaronfranke:
Actually, I don't think this behavior is desired anyway. It looks very out-of-place.

If people want to see a demo of this behavior, they could check out the 3D platformer demo.

nonunknown:
I disagree, since how any game can have a player who doesnt rotates? I would agree with you if  3D platformer demo's code was better described and better variable naming, for advanced users its fine, but for newbies like me, I had much difficult to get it to understand.

PR #370, opened 2020-01-03
Disable ETC2 texture compression in the Material Testers demo

Calinou:
Importing ETC2 textures is slow and requires a lot of RAM, so it makes sense to disable it. Those who would like to run the demo on mobile platforms can re-enable it in the Project Settings.


clayjohn:
Good thinking!

Feniks-Gaming:
surly this can be just merged it's a project setting that is simple to implement and there isn't much to debate here?

PR #315, opened 2019-03-06
Add 2.5D Demo Projects

aaronfranke:
A few users [have asked](https://github.com/godotengine/godot/issues/25980) how to create 2.5D games in Godot, so I've created two demo projects that demonstrate how such a thing would be implemented. Actually, they are both the same project, the difference is that one is written entirely in GDScript and one is written mostly in C#.

Though currently in a functional state, this is a work-in-progress. A few things on the to-do list:

- [x] ~~The platform texture used in the demo is just the first result I found on Google images. It needs to be replaced. I'm not an artist, so contributions welcome.~~ It has been replaced!

- [x] ~~The demo character could also use some improvements. Right now it's just the Godot logo.~~ Thanks to @Steph155's excellent work we have a character now!

- [x] ~~There is currently no project icon, which would be nice to have.~~ Icon is now character's face.

- [ ] The code should be reviewed by people.

ZIP download link: https://github.com/aaronfranke/godot-demo-projects/archive/2point5d.zip

PR #376, opened 2020-01-23
Update and improve plugin demos

aaronfranke:
I removed the custom dock demo, since the material creator demo already has a custom dock. Therefore, this PR supersedes and closes #291.

In the material creator demo, I added an "Apply material" button, and abstracted out the logic of creating a resource from the panel's values, so that it can be used for both apply and save. I also changed the UI so that it uses a VBoxContainer instead of just having Control nodes at specified positions.

All of the scripts have been updated to reflect the GDScript style guide. I also added README files to each (and while I was at it, I tweaked the root folder's README too).

CC @vnen @TwistedTwigleg since y'all created the original material plugin demos.


vnen:
Note that those plugin demos are references to the tutorial in the documentation: https://docs.godotengine.org/en/3.1/tutorials/plugins/editor/making_plugins.html#a-custom-dock. So if we don't want the reference project it's fine. If we do (as I prefer), we should keep both in sync and update the docs too>

IMO the material creator can go away. It is a reminiscent from pre-Godot 3.0 when we didn't have the import system. This one isn't in the docs anyway, and we already have the import demo to replace it.

aaronfranke:
Good point, I've updated this PR to link to the documentation. Do you think the documentation should link to these demo projects as well?

It's worth noting that the only existing plugin here that mirrors the documentation closely is the import plugin demo. As for the other two:

* The custom node in the documentation is "MyButton" while the custom node in this repo is "Heart". I'm not sure which should be changed to which, but anyway this PR doesn't modify the custom node demo all that much, it's just superficial changes like empty lines and of course the README.

* The custom dock plugin in this repo is fairly different from the documentation. The custom dock plugin in this repo does not showcase much activity. I think it makes sense for the material creator to be kept and to serve as a demo for custom docks, since the material creator demo also contains other useful things (like how to get and modify selected nodes).

I also checked the documentation for things that we don't have demos for. I've been working on #315 recently and if all goes well it will have a main screen in it. That leaves spatial gizmo plugins, inspector plugins, and visual shader plugins. I could possibly make demos for the first two myself. Adding these is out-of-scope for this PR, but I thought it's worth mentioning.

TwistedTwigleg:
I'm cool with the changes. 👍

aaronfranke:
I'll merge this now then. It is possible there are ways this can be improved further, if so that can be done in another PR.

PR #382, opened 2020-01-29
Fixed GDScript IK demo so it works with Godot 3.2

### LINKED ISSUE #375: IK demo is out of date

https://github.com/godotengine/godot-demo-projects/tree/master/3d/ik

Windows 10 N/a

**Issue description:**
IK is not using the skeleton ik system.

https://godotengine.org/article/skeleton-inverse-kinematic

As shown here:

https://downloads.tuxfamily.org/godotengine/demos/inverse_kinematics.zip

TwistedTwigleg:
This PR fixes the Godot IK demo so that it works with the recently released Godot 3.2.

This should partially close #375.
I could add a scene using the SkeletonIK node if desired.

______________________

The first issue was that the `set_bone_global_pose` function was renamed to `set_bone_global_pose_override` and a few new arguments were added. This was a simple fix. The only thing to note is that now the `set_bone_global_pose_override` function has a weight/interpolation value, which I set to `1` in this demo.

The second issue is a slight change in how the `set_bone_global_pose_override` function works. Prior to Godot 3.2, when you changed the transform of a bone in a Skeleton, all of the children bones would update and *stay connected* to each other. However, in Godot 3.2 with a weight value of `1`, this no longer happens. Instead, you now have to pass the correct positions for each bone, even if it is a child bone, in the transform passed to `set_bone_global_pose_override`.
This was the biggest cause of the issues with the IK demo. To work around this, the FABRIK IK script now assigns all of the transform positions for each of the bones in the bone chain, which makes it work as expected.

~~There is a small bug in the `fabrik_ik` scene that sometimes occurs with the 3-bone IK arm. I've tried to find a fix, and am still investigating what is causing the issue~~ Fixed!

Another minor change I needed to make was adding a bit of code to the `ik_look_at` script, so it can (optionally) position itself using a different bone in the Skeleton. Without this change, the right hand in the `fabrik_ik` scene wouldn't be connected to the arm.

Other minor changes:

* I adjusted the WorldEnvironment nodes so the glow effect is not quite as overpowering.
* I changed the plugin details to better reflect the goal of the plugin, and so that it shows the Godot community is the author of the plugin.
* Removed the code that straightens the FABRIK chain when the target is out of reach because it was not working. Now the FABRIK chain executes even when the target is out of reach, which gives the desired result and simplifies the code, at the cost of a little bit of performance.

### DIFF
--- a/3d/ik/addons/sade/ik_look_at.gd
+++ b/3d/ik/addons/sade/ik_look_at.gd
@@ -10,14 +10,14 @@ export(bool) var use_our_rotation_y = false
 export(bool) var use_our_rotation_z = false
 export(bool) var use_negative_our_rot = false
 export(Vector3) var additional_rotation = Vector3()
+export (bool) var position_using_additional_bone = false
+export (String) var additional_bone_name = ""
+export (float) var additional_bone_length = 1
 export(bool) var debug_messages = false
 
 var skeleton_to_use
 var first_call = true
-
-export (bool) var position_using_additional_bone = false
-export (String) var additional_bone_name = ""
-export (float) var additional_bone_length = 1
+var _editor_indicator = null
 
 
 func _ready():
@@ -131,9 +131,9 @@ func update_skeleton():
 func _setup_for_editor():
 	# So we can see the target in the editor, let's create a mesh instance,
 	# Add it as our child, and name it
-	var indicator = MeshInstance.new()
-	add_child(indicator)
-	indicator.name = "(EditorOnly) Visual indicator"
+	_editor_indicator = MeshInstance.new()
+	add_child(_editor_indicator)
+	_editor_indicator.name = "(EditorOnly) Visual indicator"
 
 	# We need to make a mesh for the mesh instance.
 	# The code below makes a small sphere mesh
@@ -150,7 +150,7 @@ func _setup_for_editor():
 	indicator_material.albedo_texture = preload("editor_gizmo_texture.png")
 	indicator_material.albedo_color = Color(1, 0.5, 0, 1)
 	indicator_mesh.material = indicator_material
-	indicator.mesh = indicator_mesh
+	_editor_indicator.mesh = indicator_mesh
 
 
 func _set_update(new_value):
--- a/3d/ik/addons/sade/ik_look_at.gd
+++ b/3d/ik/addons/sade/ik_look_at.gd
@@ -67,42 +67,42 @@ func update_skeleton():
 	if update_mode >= 3:
 		return
 	
-	# Get the bone
-	var bone : int = skeleton_to_use.find_bone(bone_name)
+	# Get the bone index.
+	var bone: int = skeleton_to_use.find_bone(bone_name)
 	
-	# If no bone is found (-1), then return (and optionally print an error)
+	# If no bone is found (-1), then return and optionally print an error.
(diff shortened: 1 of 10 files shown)


TwistedTwigleg:
I added a SkeletonIK example, which should help fully close #375. I'm not 100% sure what the magnet property in the SkeletonIK node does and if it would help or not, so I just left it empty.

I also added proper anchors to all of the UI nodes so it can scale according to the screen size. Because of this, I removed the project settings that locked the screen ratio, since now it should work with any aspect ratio.

aaronfranke:
Just so you know, I'm going to continue updating the demos to 3.1, then we'll split off a separate 3.1 branch and update all the demos to Godot 3.2. So it might be several days before I review this.

TwistedTwigleg:
No problem! I just wanted to tackle the issue while I was thinking about it. Take all the time you need 👍

TwistedTwigleg:
I’ll take a look at the errors and see if I can figure out what is going on. As for the shaky hands, it might be the look_ak IK script is updating out of sync with the FABRIK IK.

With the Skeleton IK, I did encounter that issue but I was not sure how to fix it. I _think_ I need to find a way to set the bend if the middle joint. That said, I definitely need to do some more research into the proper way to use the SkeletonIK node and make adjustments.

I’ll try to start looking at this PR (and the FPS project on the docs) once I’m able. Right now I am dealing with the flu, but I am hoping to be recovered enough to start tackling these issues soon.

3d/ik/addons/sade/ik_look_at.gd
```diff
@@ -10,10 +10,14 @@ export(bool) var use_our_rotation_y = false
 export(bool) var use_our_rotation_z = false
 export(bool) var use_negative_our_rot = false
 export(Vector3) var additional_rotation = Vector3()
+export (bool) var position_using_additional_bone = false
+export (String) var additional_bone_name = ""
+export (float) var additional_bone_length = 1
```
aaronfranke: These export statements shouldn't have the space at the beginning.

TwistedTwigleg:
I fixed the issue with the export statements and added a few static type hints to the lookat IK so it better fits the style in the FRABRIK IK code. I'm still not totally sure on what is causing the hand shake. I thought it was something being out of sync, but changing the update order didn't fix the issue.
Maybe the issue is something to do with how I'm calculating the positions in the FABRIK chain... I dunno. Regardless, if you think it is a minor issue then I'm fine with merging it as is, and hopefully later finding the solution to the shaky hand issue.

3d/ik/addons/sade/ik_look_at.gd
```diff
@@ -64,7 +68,7 @@ func update_skeleton():
 		return
 	
 	# Get the bone
-	var bone = skeleton_to_use.find_bone(bone_name)
+	var bone : int = skeleton_to_use.find_bone(bone_name)
```
aaronfranke: This doesn't follow the style guide, there should not be a space before the colon.

Also, could you make the comment `# Get the bone index.` ?

TwistedTwigleg: Done! Thanks for catching it, I totally missed the extra space.
I also tweaked and redid some of the comments in ik_look_at_.gd to better fit what the code actually does.

PR #388, opened 2020-01-31
Significantly improve the material tester demo

aaronfranke:
As of this PR, I have updated, tested, and improved every single `/3d` demo with Godot 3.1.2.

* All SCN files have been converted to TSCN.

* Organize into subfolders for test materials, models, and backgrounds.

* Significant overhaul of the script. Conform to style guide, use `$` and cache nodes, don't use quaternions anywhere, all interpolations are done with single numbers before being applied as transformations, and otherwise simplified logic overall. The old script was very overcomplicated.

* The test bed model has been modified to be centered in itself so that we don't need to offset everything in the engine. It's now exported as a GLB file instead of a DAE Collada file. The source Blender file is now saved here too (it's compressed).

* The scene has been rotated so that the balls are facing +Z, which is a more natural orientation than it was using previously (facing +X). This means we don't need to apply extra rotations. This is also why the camera is now made of 3 Spatial nodes instead of 4.

* Fix camera's starting position being on the right side of the balls.

* Allow rotating the camera with the middle and right mouse buttons.

* Change SSAO to High quality and change the blur to 3x3 (thanks @clayjohn)


clayjohn:
I was working in the demo today and noticed that the SSAO was at the lowest quality and it appeared very bad. We should likely change the SSAO to high quality with a 3x3 blur as well.

aaronfranke:
@clayjohn Thanks, I didn't even notice it until I changed the setting and saw how much better that is.

PR #392, opened 2020-01-31
Tweak 3D settings in various demos

Calinou:
- Use 4× MSAA for a better visual appearance.
- Use "3D Without Effects" framebuffer allocation when possible for better performance.
- Disable Glow in the 3D Kinematic Character demo for better performance. Increase the DirectionalLight strength to compensate.
- Use a non-linear attenuation for the OmniLight in the Window Management demo (and increase its range to compensate).

I left the Material Testers demo untouched since it's being edited in #388, but it may also be able to use the "3D Without Effects" framebuffer allocation mode.


aaronfranke:
Feel free to rebase then change the material tester demo, I just merged #388.

Calinou:
@aaronfranke The Material Testers demo already uses 4× MSAA. Also, since the demo uses SSAO, I can't make it use the 3D Without Effects framebuffer allocation mode. We can leave it as-is :slightly_smiling_face:

aaronfranke:
Since this uses 3.2, it will have to wait a few days before merging, I'm still updating everything to 3.1.2.

Issue #390, opened 2020-01-31
Setting a clear design goal for the demo projects


NathanLovato:
The code and quality of the demos in this repository vary a lot. Some are done with quick and dirty code, some have nice graphics, some have programmer art, and a lot teach poor programming practices when working with Godot.

At the last GodotCon, we talked with @akien-mga and @reduz about setting a clear design target for these demos, in the hope to bring them to a higher quality standard. We'd like to contribute (or keep contributing) towards that with @johnnygossdev, @razcore-art, and @henriiquecampos 

To me, the goals of the demos are to:

- **Showcase** engine features, that is to say, give a sense for what Godot can do. In that sense, the demos are a communication asset and may contribute to giving a positive or a negative image of the engine.
- **Help people learn** these features. So the demos have an educational purpose.

Users can get a feel for Godot's power with these demos. If an experienced developer may see the value of a feature showcased through code, a casual user or hobbyist may wrongly feel that Godot sucks because some of the demos are really bare-bones.

Teaching-wise, I'd argue that a lot of demos offer counter-productive examples of how you're meant to use Godot. That is due in particular to the fact that a lot of demos were coded to "just work".

Below is a proposal to set some guidelines for this repository, start to frame a target to collectively move towards, and help increase the quality of the demos here. We would also like to contribute demos as part of the GDQuest project, with the team.

### The demos' design ###

Each demo should **clearly focus on one feature**. The scope of the _feature_ can be _one node, one game mechanic, or one system_. 

For example:

1. `KinematicBody2D`, make a simple game with a side-view platform character, and another scene with top-down movement, and obstacles, including moving ones.
2. Touch gestures, a demo that shows how to detect a gesture, generate an input event from that, and consume it in another node.
3. User interface anchors and margins, showing the effect of various layout and anchor menu options.

To make this clearer, the idea is to have a clearer scope for the demos. We could avoid having tiny demos with several lines of code, that leave some more insights to be desired, or entire game projects that should rather be in a dedicated repository.

### The target ###

Update existing demos to:

- Show good practices when it comes to Godot and game creation. Starting with following the [official GDScript guidelines](http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html)
- Make the demos feel and look both better and more consistent.

### Having an art direction, or a visual language ###

We'd like to suggest using a simple, yet consistent visual language for the demos that don't have a game world and art of their own. An art style that:

[...]

PR #405, opened 2020-02-05
Remake the 2d platformer demo

NathanLovato:
This Godot project is a complete rewrite of the official Platformer 2D demo. This is part of a proposal to improve and harmonize the official demo projects. We want to:

- Teach good programming practices with Godot.
- Bring demos to a higher quality standard.
- Keep unifying the code in the demos, a big task that @aaronfranke has been working on lately.

See the [open issue](https://github.com/godotengine/godot-demo-projects/issues/390) for more information.

This new demo showcases features from the original and works with Godot 3.2.
This is the same demo from https://github.com/GDQuest/godot-demo-projects-remake, coded by @johnnygossdev, marked as a co-author of the commit, first [reviewed by me](https://github.com/GDQuest/godot-kickstarter-2019/pull/58), and then by @aaronfranke.

We hope you'll like it!


NathanLovato:
I can write a README for the demo if you'd like, I just need to know if there's a format or anything, in particular, that you need there. The source code contains quite a lot of comments for learning purposes.

2d/platformer/src/UserInterface/PauseMenu.gd
```diff
@@ -0,0 +1,26 @@
...
+func open():
+	visible = true
+	resume_button.grab_focus()
+
+
+func close():
+	visible = false
+
+
+func _on_ResumeButton_pressed() -> void:
```
aaronfranke: Akien said he doesn't want static typing used in most of the demo projects. I personally like static typing and I'm not against using it in some places if it improves readability. However, `-> void` is one of the least useful things in static typing, and it doesn't seem necessary here, so I don't think this should be here.

NathanLovato: I had removed the type hints, but it's probably because I forgot to remove the static type hints when switching to the project at some point. I've got a script to remove them all.

In the previous discussion about code order and the style guide, I think we chose to use dynamic only unless the project or code snippets showcase static typing. And it's in the docs iirc. So I'll just remove them all.

NathanLovato:
Thanks, somehow I didn't get or missed the notification! I'll address the requested changes ASAP.

NathanLovato:
I've just addressed everything except for the background for now. I'm trying to get the game window to stay within supported screen ratios:

- I'd rather not have black bars on the game window
- You don't want the user to be able to resize to any window sizes with the expand stretch mode as they can see a lot more of the level, thus break the game/level design.

Do you know how to do that well? I have some code that does keep the game within size ratios, but there are bugs as I'm forcing a window resize every time the viewport resizes.

Here's the current code that's failing:

```gdscript
# Only allow window size ratios between 4:3 and 2:1, which are common display ratios
export var screen_ratio_range = Vector2(4.0 / 3.0, 2.0 / 1.0)

[...]

func _on_SceneTree_screen_resized():
	var window_size = OS.get_window_size()
	var ratio = window_size.x / window_size.y
	var ratio_clamped = clamp(ratio, screen_ratio_range.x, screen_ratio_range.y)
	# Use the is_equal_approx function to compare two decimal values, as computers 
	# don't store floats with absolute accuracy.
	var is_ratio_in_range = is_equal_approx(ratio_clamped, ratio)
	
	if not is_ratio_in_range:
		OS.window_size = (window_size / ratio * window_size.x).round()
```

2d/platformer/README.md
```diff
@@ -0,0 +1,25 @@
+# 2D Platformer #
```
aaronfranke: It's interesting that this works (using # at the end, TIL), but it's not common practice in Markdown, so I'd suggest avoiding it. Should look like:

    # 2D Platformer

2d/platformer/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This demo is a pixel art 2D platformer with graphics and sound.
+
+It shows you how to code characters and physics-based objects in a real game context. This is a relatively complete demo where the player can jump, walk on slopes, fire bullets, interact with enemies, and more. It contains one closed level, and the player is invincible, unlike the enemies.
+
+You will find most of the demo’s content in the `Level.tscn`. You can open it from the default `Game.tscn` scene, or double click on `Level.tscn` in the `src/Level/` directory.
```
aaronfranke: "in the `Level.tscn` [scene/file]", or maybe drop the article ("the") if you think it's too many words: "in `Level.tscn`"

2d/platformer/README.md
NathanLovato: This works everywhere, that's how emacs inserts markdown headings. But I'll remove it if it's a problem.

NathanLovato:
I copied the settings from the original demo. So now there are black bars at the top and the bottom of the window when you make it taller than the base ratio. The last commit also prevents the background from blinking, and I extended it so you don't see the clear color when the background is wide.

I also tweaked the markdown.

2d/platformer/src/Main/Game.gd
```diff
@@ -0,0 +1,31 @@
...
+# the game or toggling the window full-screen.
+extends Node
+
+
+# The "_" prefix is a convention to indicate that variables are private,
+# that is to say, another node or script should not access them.
+onready var _pause_menu = $InterfaceLayer/PauseMenu
+
+
+func _init() -> void:
```
aaronfranke: 🤔

johnnygossdev: Hi! I'm just jumping in. What's the issue here? Is it the use of `-> void:`?

I've just tested and the platforms in the background are visible for me but the clouds are indeed to the right. This should be a simple fix of offsetting the `ParallaxBackground`.

I've made some tweaks and I can't find issue with wide resolutions.
platformer-wide

I can make other touch ups and commit if that's all good @NathanLovato

NathanLovato: Yeah sure, by the way @aaronfranke you have write access to my branch and PR, so when you find a typo or something, you can remove it right away - it'll take you less time than asking us to remove one.

I'm jumping between many projects every day, all but this one use type hints and the related completion. I don't even notice them to be honest. I'll check if I can't add a local pre-commit commit hook to run the script that cleans type hints for future contributions. Or at least I should be able to do that on save in Emacs.

NathanLovato: Also, I can confirm the platforms do show up here - and yes the clouds are offset, I'll move them to the left. I'll also try to see what could make the platforms disappear.

Screenshot from 2020-02-12 07-40-31

NathanLovato: Is there a way to change editor settings per-project? Type hints seems like something that could be overriden per file or per project.

NathanLovato:
Here, like yesterday, the platforms show at every window size. I fixed the clouds, but also tried to change the parallax layer setup so they don't get culled - they had their origin in the top-left of the game window, maybe this can create some edge cases?

I also removed the type hint. 

Screenshot from 2020-02-12 07-50-38

@johnnygossdev I took care of this one because you don't have write access to my branch and it just took a few minutes. It's a matter of adding the final touches at this point.

johnnygossdev:
Sure no problem! :+1:

NathanLovato:
Just added some clouds for when making the window too long

Screenshot from 2020-02-12 07-54-17

NathanLovato:
Ah, I had forgotten to change the code order, thanks for cleaning this up!

NathanLovato:
> Once concern is that it might be over-documented in a few places, but I think that's fine since I would expect this demo to be viewed by beginners who need this information.

Yes, so that was the idea. We assumed that this would be a go-to demo for beginners as it's the kind of thing a newcomer may want to do with Godot.

Also, we've designed it with the idea of not writing and having to maintain a long text-based, step-by-step tutorial, which is difficult to keep up to date. So all the explanations are inline.

Thanks for taking the time to review and make fixes!

2d/platformer/src/UserInterface/PauseMenu.gd
```diff
@@ -0,0 +1,26 @@
+extends Control
```
Sslaxx: Seems odd that this file is entirely uncommented.

2d/platformer/src/Main/Game.gd
Calinou: > Is there a way to change editor settings per-project? Type hints seems like something that could be overriden per file or per project.

Not yet, although I suppose we could move the type hint setting from the Editor Settings to the Project Settings. If we really need it, we could keep the editor setting and make it an enum with 3 values: "Project Default", "Always", "Never".

2d/platformer/src/UserInterface/PauseMenu.gd
NathanLovato: It's not the core of the demo, that's about 2D platforming. I'd recommend following a tutorial dedicated to pausing, as half of the work is happening in the inspector, with the pause mode on the nodes.

Sslaxx: I meant within the context of the demos, not tutorials or other documentation. If the actual pause demo is going to be updated as well, at least giving a pointer to it might here be useful.

In general, if something is going to be demo-ed somewhere else in more detail, pointing to that in another demo might be helpful.

NathanLovato: Sorry, I may not get what you mean. You mean we could add a comment telling people to go look at a dedicated demo about pausing? If so sure, anyone can commit something like that anytime.

Sslaxx: Telling people which other demo would demonstrate a feature in more detail, yes.

PR #412, opened 2020-02-15
Changing the art assets for the 2D platformers to "real" pixel art (instead of 2x upscaled)

### LINKED ISSUE #409: 2D Platformer Demo: All art assets are 2x the necessary size.
[2D Platformer Demo (Kinematicbody)](https://godotengine.org/asset-library/asset/120)
[2D Physics Platformer Demo (Rigidbody)](https://godotengine.org/asset-library/asset/119) 

Affects any OS/device

**Issue description:**

What is the current behavior?
If I would like to use this demo as a starting point for my own pixelart platformer, I can't, because simply alter or replacing art assets results in all kinds of unexpected behaviour, because every visual pixel in the assets is 4 actual pixels (all art is 2x scaled up)

What is the expected behavior?
Expected behavior would be to be able to simply swap the art asset with some other and start poking around in the code changing values here and there, to create a new enemy or slightly different player behavior ... beginner stuff.

Tell us the steps to reproduce the bug, and if possible share a minimal demo of the problem.

1. Download the and run the project, see something you would like to experiment with,
2. open one of the art assets in your graphics editing software
3. realize the art is not pixel art
4. try to make it pixelart so you can work with it, by scaling it down
5. run the game
6. realize this breaks everything

or

1. Download the and run the project, see something you would like to experiment with,
2. swap out one of the art assets with an existing one of your own
3. run the game
4. realize this breaks everything

**Screenshots of issue:**
It does make simple replacing with other pixelart or amending it impossible for beginners. even for experienced users it's a lot of work and no fun at all. None of the scales in the game matches anymore.

platformer_demo_pixelsize

Fixing that is a lot of work even and you already have to know the engine to do this. Which defeats the purpose of this demo as a learning device for newcomers quite substantially.

platformer_demo_pixelsize2

If anyone would want to add another action for the robot dude, they would have to use a 2x2 pixel brush. That would less of a problem if graphic software could also be set to snap the brushtip to a 2x2 pixel grid. But they don't. They snap to single pixels. If I were to set the grid to every 2 pixels and set my brush to snap to the grid, the 2x2 pixel is set in the center, bet the top left corner like it is in the art assets of this project. So you work without snap but constantly have to check and correct your brush strokes to make the 2x2 pixels line up again.

NetroScript:
This should close #409 
All sprite textures were scaled by 0.5 with nearest neighbor
Additionally all values considering positions, sizes should have been adjusted

Directly dividing the player accelerations in the physics demo by 2 did not work, so I used some custom values, if you have better ones, feel free to tell me


NetroScript:
Did the changes you requested, I additionally forgot the hitbox of the bullet in the physics demo, and the pixel art texture import for the bullet in the non physics demo. Additionally I adjusted the spawn positions for the bullets.
With the changed gravity the halved forces also worked.

While testing the platformer demo I saw that the current stretching mode will cause the parallax background to disappear (well the gray background showing because it is not not wide enough) if the user has a monitor with more1920px width. (And max window width is set to the monitor width, I for example have a 2560px monitor) should I open an issue for that or is that not important for the scope of the demo?

PR #413, opened 2020-02-15
Add a viewport 3D scaling demo

Calinou:
This demo shows how to scale the 3D viewport rendering without affecting 2D elements such as the HUD. It also demonstrates how to toggle filtering on a viewport by using TextureRect to display the ViewportTexture delivered by the Viewport node.


viewport/3d_scaling/hud.tscn
```diff
@@ -0,0 +1,105 @@
...
+__meta__ = {
+"_edit_use_anchors_": false,
+"_editor_description_": ""
+}
+
+[node name="Viewport" type="Viewport" parent="TextureRect"]
+size = Vector2( 1024, 600 )
+handle_input_locally = false
+usage = 3
+render_direct_to_screen = true
```
clayjohn: This only applies when using GLES2. It's better to leave off for a demo so that users don't get confused.

Calinou: I added it accidentally while prototyping, so I just removed it.

viewport/3d_scaling/hud.gd
```diff
@@ -0,0 +1,41 @@
...
+var scale_factor = 1.0
+
+onready var texture_rect = $TextureRect
+onready var viewport = $TextureRect/Viewport
+onready var scale_label = $VBoxContainer/Scale
+onready var filter_label = $VBoxContainer/Filter
+
+func _ready():
+	# Required to change the 3D viewport's size when the window is resized.
+	get_viewport().connect("size_changed", self, "_root_viewport_size_changed")
```
aaronfranke: This line of code produces a warning that the return value is discarded. The warning should be silenced.

Calinou: I thought that warning was now disabled by default… Anyway, fixed.

Calinou:
@aaronfranke I don't know what's causing the error, nor if there is a way to fix/silence it.

PR #419, opened 2020-02-19
Update networking and multiplayer demos

### LINKED ISSUE #88: Variable is named host instead of client
In the join_game function the variable is named host instead of client:
https://github.com/godotengine/godot-demo-projects/blob/3616322c708b02cb7380b028f3fb661d3991a7af/networking/multiplayer_bomber/gamestate.gd#L135

aaronfranke:
All demos: Style fixes, open and save in Godot 3.2.

Multiplayer bomber:

* Allow moving with WASD or controllers.

* Get the player name from the system. There might be a better way to do this.

* Change name of a variable in `join_game`, fixes #88.

All of the demos could use further improvements, especially when it comes to documentation. A few don't work and I'm not sure if I just didn't set up the demo correctly.

### DIFF
--- a/networking/multiplayer_bomber/gamestate.gd
+++ b/networking/multiplayer_bomber/gamestate.gd
@@ -1,55 +1,61 @@
 extends Node
 
-# Default game port
+# Default game port. Can be any number between 1024 and 49151.
 const DEFAULT_PORT = 10567
 
-# Max number of players
+# Max number of players.
 const MAX_PEERS = 12
 
-# Name for my player
+# Name for my player.
 var player_name = "The Warrior"
 
-# Names for remote players in id:name format
+# Names for remote players in id:name format.
 var players = {}
+var players_ready = []
 
-# Signals to let lobby GUI know what's going on
+# Signals to let lobby GUI know what's going on.
 signal player_list_changed()
 signal connection_failed()
 signal connection_succeeded()
 signal game_ended()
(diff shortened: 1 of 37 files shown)


networking/multiplayer_bomber/gamestate.gd
```diff
@@ -1,107 +1,115 @@
 extends Node
 
-# Default game port
+# Default game port. Can be any number between 1024 and 49151.
```
Calinou: This reminds me, should we add a warning in core when listening to a port greater than 49151 (as server)? Such ports are reserved for ephemeral use.

networking/multiplayer_bomber/lobby.gd
```diff
@@ -7,66 +7,78 @@ func _ready():
 	gamestate.connect("player_list_changed", self, "refresh_lobby")
 	gamestate.connect("game_ended", self, "_on_game_ended")
 	gamestate.connect("game_error", self, "_on_game_error")
+	var desktop_path = OS.get_system_dir(0).split("/")
```
Calinou: This is not immediately obvious, so there should be a comment:

```suggestion
	# Set the player name according to the current username.
	var desktop_path = OS.get_system_dir(0).split("/")
```

aaronfranke: Done and done, except I think it's better to fallback to the path.

PR #341, opened 2019-07-20
Handle multiple resolutions in most demos

Calinou:
This makes demos render correctly on hiDPI displays, while also demonstrating how to handle multiple resolutions.

The 3D in 2D demo now uses "3D No-Effects" for the 3D viewport, which is faster to render. Thanks to this, 4× MSAA is now enabled for a better result.

The background loading demo now uses mipmaps for better-looking images.

The material testers demo now samples mouse input in a resolution-independent manner when panning.

Default clear colors were also changed in some projects for visual consistency with the project's theme.

2D platformer demos aren't affected by this PR, but I could continue the work in https://github.com/godotengine/godot-demo-projects/pull/323 (or another PR) if this is desired.

As this is a large PR, it's likely to conflict with other PRs over time. Feel free to merge other PRs before and I'll solve conflicts as needed.

PR #420, opened 2020-02-19
fix 3d in 2d demo for vulkan master

### LINKED ISSUE #422: Vulkan: 3D in 2D viewport not working
**Godot version:**
Vulkan master
Arch Linux
**Issue description:**
2D in 3D viewports work (display correctly). But it doesn't work the other way around.
This becomes a problem when the whole display of a game is black because the viewport doesn't render anything (e. g. when using anaglyph effect with two cameras).
Maybe it's not implemented yet.
**Steps to reproduce:**
3D in 2D - bug
(the 3D character is not visible)

**Minimal reproduction project:**
- https://github.com/godotengine/godot-demo-projects/tree/master/viewport/3d_in_2d (godot demo 3d in 2d)
- https://www.dropbox.com/sh/mhjvlg6zot8and3/AAC7JvaNF_bVst2ILFVW6J6ha?dl=0 (anaglyph_tutorial.zip)

LinuxUserGD:
fixes #422 (thought it would be an issue with vulkan directly).
This fixes rendering of the demo on Vulkan while y is still flipped for compatibility with Godot 3.2 stable (changed rotation from 180° to 0° but x and y scale to -1).
Instead of using clear color, it now uses the gray editor color for camera environment (background). Removed Camera2D.


clayjohn:
The flipped y is an issue that was present in Vulkan but should be fixed in master now. I don't think we should update a demo to work around a bug in the rendering code.

aaronfranke:
@clayjohn Generally, if a demo can be updated to be future-proofed and still work in 3.2, then I think we should do it. But if this was a temporary bug that doesn't exist anymore in master anyway, then yeah we should not merge this. @LinuxUserGD may wish to update this PR with just the other changes (removing Camera2D) if those changes are still seen as improvements.

Calinou:
@LinuxUserGD The Camera2D node is here so the demo can support multiple resolutions, see #341.

Issue #423, opened 2020-02-20
Can you combine all into one project?


kienvn:
Combine all into one project will help beginer explore instead open each project

aaronfranke:
No, this is not possible for a few reasons (different renderers, different input settings, different project settings) and it's also very impractical to do so. It's much better the way it is now.

If you are annoyed by how long it takes to open each project, try using "Scan" in the project manager.

PR #424, opened 2020-02-20
Add instructions on importing all projects at once in the README

Calinou:
This may be helpful to newcomers; see #423.


aaronfranke:
Perhaps we could also include a link and instructions for the asset library?

PR #426, opened 2020-03-05
Convert most demos to GLES 2 and update to Godot 3.2

aaronfranke:
If a demo fully works and looks the same in GLES 2, then we should use GLES 2. This means that these demos are more compatible with older systems that use GLES 2. This also means that these demos will hopefully be more future proof since Godot 4.0 will not have GLES 3.

If a demo wasn't converted and is still using GLES 3, then it has visual regressions or doesn't look the same in GLES 2, so we'll keep using GLES 3 for them.

There are also a few misc changes, fix Viewport error in 3D scaling demo, add an icon for multiplayer bomber, and rename some files in the BPM sync demo.


clayjohn:
Great work!

PR #429, opened 2020-03-07
Navigation A star using Mono

MugenTwo:


mono/navigation_astar/Character/Character.cs
```diff
@@ -0,0 +1,103 @@
...
+    private Vector2 targetPointWorld;
+    private Vector2 targetPosition;
+    private Vector2 velocity;
+
+    public override void _Ready()
+    {
+        this.path = new List<Vector2>();
+        this.targetPointWorld = new Vector2();
+        this.targetPosition = new Vector2();
+        this.velocity = new Vector2();
```
aaronfranke: No need to use `this` here, and there's many places throughout the script where `this` is used where it isn't necessary and shouldn't be used.

mono/navigation_astar/Character/CharacterState.cs
```diff
@@ -0,0 +1,7 @@
+public enum CharacterState
+{
+
+    IDLE,
+    FOLLOW
+
+}
```
aaronfranke: Please include a newline at the end of each file (except for the csproj file, which should not end with a newline), and also don't include newlines after opening braces or before closing braces.

MugenTwo: @aaronfranke I am going to make all the changes according to Godot's C# Guideline to be consistent with the other projects in this mono-repo. But for me there is one thing that doesn't fit in right, I don't think that using underscore for private member instance variables is a good idea. I think Godot C# should be consistent with MS C# guidelines on this regard: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions

MugenTwo:
@aaronfranke I am going to make all the changes according to Godot's C# Guideline to be consistent with the other projects in this mono-repo. But for me there is one thing that doesn't fit in right, I don't think that using underscore for private member instance variables is a good idea. I think Godot C# should be consistent with MS C# guidelines on this regard: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions

MugenTwo:
@aaronfranke changes are done

mono/navigation_astar/Character/Character.cs
```diff
@@ -0,0 +1,101 @@
...
+        Vector2 steering = desiredVelocity - _velocity;
+        _velocity += steering / Mass;
+        Position += _velocity * GetProcessDeltaTime();
+        Rotation = _velocity.Angle();
+    }
+
+    private bool ArrivedTo(Vector2 destination)
+    {
+        return Position.DistanceTo(destination) < ArriveDistance;
+    }
```
aaronfranke: I'm not necessarily opposed to abstracting code like this, but if you do, then you need to keep the GDScript version in sync and make the changes there too.

mono/navigation_astar/README.md
```diff
@@ -0,0 +1,12 @@
+# Grid-based Navigation with Astar
```
aaronfranke: Title should include "with C#" or similar. Same with the project name in `project.godot`.

mono/navigation_astar/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+This is an example of using AStar for navigation in 2D,
+complete with Steering Behaviors in order to smooth the movement out.
+
+Language: [C#](https://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/index.html)    
+
+Renderer: GLES 2
+
+## Screenshots
```
aaronfranke: Missing "Note: There is a GDScript version here." like the other README files have.

mono/navigation_astar/Character/Character.cs
MugenTwo: Do you think the C# sharp version should be a literal translation of the GD script version: method to method line by line?
I think you have already seen it in this PR that I added some methods, constants, variables that weren't in the GDScript version because, well, I felt like the code is more readable that way.

MugenTwo: For example, in this case I felt like the method move_to in the GDScript equivalent was violating the Single Responsibility Principle in the sense it was moving and at the same time it was checking whether the character had arrived to the destination.

mono/navigation_astar/Tileset/PathFindingAStarTileSet.cs
```diff
@@ -0,0 +1,193 @@
...
+        Vector2[] cellPathArray = _aStarNode.GetPointPath(startCellIndex, endCellIndex);
+        for (int i = 0; i < cellPathArray.Length; i++)
+        {
+            _cellPath.Add(cellPathArray[i]);
+        }
+
+        Update();
+    }
+
+    private void ChangePathStartPosition(Vector2 newPathStartPosition)
```
MugenTwo: @aaronfranke Another notable change that I made was that in GDScript this method was being declared as a setter. I personally didn't have a very good experience with setters that do more than just setting value. 

My experience is that people that add logic in setters generally confuse other people. I've seen people delete and update DB entries using a setter, change other class properties, calculate stuffs,... So while other people who didn't know about the setter implementation were happily using the setter, they didn't realize that they were actually messing up some DB tables, slowing down the program and such. Using another another name like what I did "ChangeBlahBlahBlah" usually will get people to notice that method isn't just setting the value and it usually starts a conversation with the people that implemented it.

mono/navigation_astar/Character/Character.cs
aaronfranke: The point is that

1. If the C# code is more readable this way, perhaps the GDScript code would be too.

2. It should be easy for users to read code from both versions and follow what's going on. Being close to a direct translation helps with this.

mono/navigation_astar/Tileset/PathFindingAStarTileSet.cs
aaronfranke: Yes, I agree, this is too much logic for a setter of a property. Still, you should change the GDScript version to be the same way - it's still a bad idea in GDScript.

aaronfranke:
If you'd like this to be merged, it needs to have a high amount of parity with the GDScript version.

Feel free to submit a PR to change the GDScript version, then amend this, then this can be re-opened.

Note that you *should not force push while the PR is closed* because it breaks GitHub - if you make changes and you think it's ready to be re-opened, either push an additional commit and comment, or let me know and then I can re-open and then you can force-push.

PR #430, opened 2020-03-07
Kinematic character 3d using Mono

MugenTwo:


MugenTwo:
@aaronfranke changes are done here as well

aaronfranke:
As I said with #429, if you'd like this to be merged, it needs to have a high amount of parity with the GDScript version.

Feel free to submit a PR to change the GDScript version, then amend this, then this can be re-opened.

Note that you *should not force push while the PR is closed* because it breaks GitHub - if you make changes and you think it's ready to be re-opened, either push an additional commit and comment, or let me know and then I can re-open and then you can force-push.

PR #437, opened 2020-03-10
Navigation mono demo

MugenTwo:


mono/navigation/README.md
```diff
@@ -0,0 +1,15 @@
+# Navigation Polygon 2D
```
aaronfranke: The name should include "with C#" at the end.

Missing "Note: There is a GDScript version here" like with the other C# demos.

mono/navigation/project.godot
```diff
@@ -0,0 +1,32 @@
...
+config_version=4
+
+_global_script_classes=[  ]
+_global_script_class_icons={
+
+}
+
+[application]
+
+config/name="2d-navigation-demo-cs"
```
aaronfranke: The name also needs to be changed here.

mono/navigation/Navigation.cs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public override void _Input(InputEvent inputEvent)
+    {
+        if (!inputEvent.IsActionPressed("click"))
+        {
+            return;
+        }
+        Sprite character = GetNode("Character") as Sprite;
+        UpdateNavigationPath(character.GetPosition(), GetLocalMousePosition());
```
aaronfranke: 'Node2D.GetPosition()' is obsolete: 'GetPosition is deprecated. Use the Position property instead.'

Also, the "as" operator is inappropriate to use here because `null` is not an acceptable value for the next line. You can just use `GetNode<Sprite>` or `GetNode<Node2D>` instead and you don't need a separate line for saving this.

mono/navigation/Navigation.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    private void UpdateNavigationPath(Vector2 startPosition, Vector2 endPosition)
+    {
+        _path.AddRange(GetSimplePath(startPosition, endPosition, true));
+        SetProcess(true);
+    }
+
+    private void MoveAlongPath(float walkDistance)
+    {
+        Sprite character = GetNode("Character") as Sprite;
```
aaronfranke: Same as above.

PR #439, opened 2020-03-10
Platformer mono demo

MugenTwo:


aaronfranke:
Just an FYI, we don't want to convert every single demo to C#, that would just be bloat.

Please only submit a converted demo if you believe there is value in seeing that particular demo converted (ex: maybe the demo uses advanced languages features, or can make use of optimizations in C# that don't exist in GDScript).

MugenTwo:
That makes a lot of sense. Although, I am putting these demos up because I already had them (so it was easy for me). In fact these last 3 demo projects that I PR'd in this mono-repo were already submitted and accepted in Godot Asset Library. It's up to you guys to merge the PR if you see any value in them or no.

aaronfranke:
I'm not sure I will accept all of the PRs you made (I very well could, or I could not, really just not sure), so if you feel like any of them in particular are not useful, feel free to close them. In any case I will most likely review them all anyway and think for awhile before making a decision.

I will probably accept this one (after review of course) because the Platformer demo is one that I see as high quality, fairly complex, and does some interesting things like having Viewports for splitscreen.

aaronfranke:
Actually, I will just close this PR. The GDQuest team recently recreated the Platformer demo from scratch. They might do the same with the Physics Platformer demo since they're very similar, and I don't want to tell you to recreate the Physics Platformer demo from scratch if it would just need to be recreated from scratch again later.

PR #425, opened 2020-03-01
Multiple-threads loading

VitaZheltyakov:
Multiple-threads loading demo.
This demo shows how you can organize background loading using multiple threads.
Сompatible with the latest stable Godot version (currently 3.2).
For more information, see:
[http://docs.godotengine.org/en/3.2/tutorials/io/background_loading.html#using-multiple-threads](http://docs.godotengine.org/en/3.2/tutorials/io/background_loading.html#using-multiple-threads)

*Bugsquad edit:* Supercedes #281

PR #448, opened 2020-03-15
Style tweaks and add import files to multiple threads loading demo

aaronfranke:
Amending #425 with some minor changes.


VitaZheltyakov:
https://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html#classes-and-nodes
`queue` in preload.gd

aaronfranke:
`queue` isn't a preloaded class, it's an instance of one.

PR #450, opened 2020-03-20
Add a formatting script for local checking and for Travis CI

aaronfranke:
There have been a few pull requests which have minor file formatting issues such as missing a newline at the end of text files. It would be nice to automate checking them, and also to enforce consistent file formatting. This PR adds a simple script which can be ran locally to fix file formatting, or ran on Travis CI in the case of checking pull requests. We could possibly add more checks later.

The bottom half of the script was based on the script in the main Godot repo.

Travis CI builds won't run unless they're enabled, so pinging @akien-mga.


akien-mga:
Thanks!

PR #451, opened 2020-03-22
simplifying steering to a single line with move_toward

LaBodilsen:
Replaced:
```
	if steer_target < steer_angle:
		steer_angle -= STEER_SPEED * delta
		if steer_target > steer_angle:
			steer_angle = steer_target
	elif steer_target > steer_angle:
		steer_angle += STEER_SPEED * delta
		if steer_target < steer_angle:
			steer_angle = steer_target
	
	steering = steer_angle
```
with
`	steering = move_toward(steering, steer_target, STEER_SPEED * delta)`

### DIFF
--- a/3d/truck_town/vehicle.gd
+++ b/3d/truck_town/vehicle.gd
@@ -1,9 +1,8 @@
 extends VehicleBody
 
-const STEER_SPEED = 0.2
+const STEER_SPEED = 1
 const STEER_LIMIT = 0.4
 
-var steer_angle = 0
 var steer_target = 0
 
 export var engine_force_value = 40
@@ -27,4 +26,4 @@ func _physics_process(delta):
 	else:
 		brake = 0.0
 	
-	steering = lerp(steering, steer_target, STEER_SPEED)
+	steering = move_toward(steering, steer_target, STEER_SPEED*delta)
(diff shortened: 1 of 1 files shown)


aaronfranke:
I think it would be better to use `move_toward` instead. If you change `lerp` to `move_toward`, add `* delta` in the last part, and set `STEER_SPEED` back to 1, the behavior is the same as before.

Also, `steer_angle` can be removed since it's not used anymore.

LaBodilsen:
Clever to use the move_towards function instead. 

i agree that it would be the better solution.  should make a new pull request for this, and revoke this one?

3d/truck_town/vehicle.gd
```diff
@@ -27,13 +26,4 @@ func _physics_process(delta):
...
-		steer_angle -= STEER_SPEED * delta
-		if steer_target > steer_angle:
-			steer_angle = steer_target
-	elif steer_target > steer_angle:
-		steer_angle += STEER_SPEED * delta
-		if steer_target < steer_angle:
-			steer_angle = steer_target
-	
-	steering = steer_angle
+	steering = move_toward(steering, steer_target, STEER_SPEED*delta)
```
aaronfranke: There should be spaces around `*`

LaBodilsen: You are of course correct.  

That is now implemented in the request.

aaronfranke:
Squashing-and-merging, thanks!

PR #456, opened 2020-03-27
Prevent additional button presses from remapping actions.

tokc:
Fix a bug in gui/input_mapping/ActionRemapButton.gd which causes every previously clicked button to be remapped to the last pressed action.

This may not be the ideal way to prevent the behaviour, but it succeeds in making the buttons work as expected.

### DIFF
--- a/gui/input_mapping/ActionRemapButton.gd
+++ b/gui/input_mapping/ActionRemapButton.gd
@@ -22,6 +22,9 @@ func _unhandled_key_input(event):
 	# you want to work with gamepads.
 	remap_action_to(event)
 	pressed = false
+	
+	# Stop processing inputs once the action has been remapped.
+	set_process_unhandled_key_input(false)
 
 
 func remap_action_to(event):


aaronfranke:
I can't reproduce the bug that this PR is supposed to fix. Can you post reproduction steps?

tokc:
OK. In that case let's double-check all my settings. It is likely I am doing something wrong!

Step 1:
```$ uname-a 
Linux syringa 4.19.108-1-MANJARO #1 SMP Thu Mar 5 20:27:09 UTC 2020 x86_64 GNU/Linux

$ godot --version
3.1.2.stable.official

$ git clone https://github.com/godotengine/godot-demo-projects.git
Cloning into 'godot-demo-projects'
```

Step 2: Open the project, press play.
Step 3: Click "Left Key", press A.
Step 4: Click "Right Key", press D.
Expected behaviour: "Left Key" now reads "A Key", and "Right Key" now reads "D Key".
Observed behaviour: "Left Key" and "Right Key" both read "D Key".

Peek 2020-03-31 12-23

aaronfranke:
This repo is currently targeting Godot 3.2, please re-test with Godot 3.2.1.

tokc:
It appears to work in 3.2.1.

Thank you for your time.

Issue #454, opened 2020-03-26
Dodge The Creeps Demo does not match "Your first game" tutorial in the docs


golddotasksquestions:
**Which demo project is affected:**
Dodge The Creeps Demo

**OS/device including version:**
Win 64

**Issue description:**
The part about [removing the creeps](https://docs.godotengine.org/en/3.2/getting_started/step_by_step/your_first_game.html#removing-old-creeps) has not been added to the demo yet.

It's very easy to add, but since this demo is meant for absolute beginners, it can irritate or confuse them if it is not exactly the same as the tutorial. [Example here](https://www.reddit.com/r/godot/comments/fpdck9/tutorial/).
Imho the demo should match the tutorial 100% so beginners can use it as reference and compare their own project when they first follow the "Your first game" tutorial in the docs.

PR #459, opened 2020-04-03
Updated the dodge_the_creeps demo to include the "Removing the creeps" part in the tutorial.

EludeQ:
This is for #454.
A [small part from the tutorial](https://docs.godotengine.org/en/3.2/getting_started/step_by_step/your_first_game.html#removing-old-creeps) for the dodge_the_creeps demo was missing in the project here.

### DIFF
--- a/2d/dodge_the_creeps/Main.gd
+++ b/2d/dodge_the_creeps/Main.gd
@@ -33,6 +33,8 @@ func _on_MobTimer_timeout():
 	direction += rand_range(-PI / 4, PI / 4)
 	mob.rotation = direction
 	mob.linear_velocity = Vector2(rand_range(mob.min_speed, mob.max_speed), 0).rotated(direction)
+	# warning-ignore:return_value_discarded
+	$HUD.connect("start_game", mob, "_on_start_game")
 
 
 func _on_ScoreTimer_timeout():
--- a/2d/dodge_the_creeps/Mob.gd
+++ b/2d/dodge_the_creeps/Mob.gd
@@ -11,3 +11,7 @@ func _ready():
 
 func _on_VisibilityNotifier2D_screen_exited():
 	queue_free()
+
+
+func _on_start_game():
+	queue_free()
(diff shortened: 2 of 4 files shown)


2d/dodge_the_creeps/Mob.gd
```diff
@@ -11,3 +11,6 @@ func _ready():
 
 func _on_VisibilityNotifier2D_screen_exited():
 	queue_free()
+
+func _on_start_game():
```
aaronfranke, line 15: There should be two lines between methods to follow the [GDScript style guide](https://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_styleguide.html).

2d/dodge_the_creeps/Main.gd
```diff
@@ -33,6 +33,7 @@ func _on_MobTimer_timeout():
 	direction += rand_range(-PI / 4, PI / 4)
 	mob.rotation = direction
 	mob.linear_velocity = Vector2(rand_range(mob.min_speed, mob.max_speed), 0).rotated(direction)
+	$HUD.connect("start_game", mob, "_on_start_game")
```
aaronfranke, line 36: The function 'connect()' returns a value, but this value is never used.

Please add `# warning-ignore:return_value_discarded` above this line to silence the warning.

EludeQ: Alright! I've added the warning ignore and added that missing line. I've also added this to the C# version.

mono/dodge_the_creeps/Main.cs
```diff
@@ -3,86 +3,88 @@
 
 public class Main : Node
 {
-    [Export]
-    public PackedScene mob;
+	[Export]
+	public PackedScene mob;
```
aaronfranke, line 7: Because it's the C# standard, we use spaces for indentation exclusively in C# files.

EludeQ: Woops! Didn't know Godot changes indents automatically. Should be fixed now.

2d/dodge_the_creeps/Main.gd
aaronfranke: But you didn't actually add this line.

PR #463, opened 2020-04-09
3d Platform using Csharp

nonunknown:
3D Platformer using a Finite State Machine and C#

It contains:
* Basic Player movement
* Camera Look at Player
* Finite State Machine
* 3d Model + Animations


aaronfranke:
I think what you're looking for is the [asset library](https://godotengine.org/asset-library/asset).

PR #465, opened 2020-05-08
Add translation ignore

aaronfranke:
Now synced with GitHub's Godot.gitignore plus the bottom section.

https://github.com/github/gitignore/blob/master/Godot.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,11 @@
+
 # Godot-specific ignores
 .import/
 export.cfg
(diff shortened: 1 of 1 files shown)


Calinou:
Shouldn't we modify Godot to import translations into the `.import/` folder eventually?

aaronfranke:
@Calinou Yes, but https://github.com/godotengine/godot/issues/23368 should be done first.

PR #466, opened 2020-05-08
More realistic jump.

szmyd:
Do not modify x and y movement if we are midair, but if we have momentum carry that into the jump.


aaronfranke:
I appreciate the commitment to realism, but the existing behavior is very common for games and is also the behavior in the Platformer 2D demo. I think it should stay the way it is.

By the way, you forgot to adjust the sprite logic.

PR #480, opened 2020-06-17
Remove trailing space characters in the formatting script

aaronfranke:
Remove trailing spaces, but don't remove trailing tabs.


Calinou:
Out of curiosity, why are trailing tabs preserved?

aaronfranke:
Tabs have semantic meaning in GDScript, and they are visible inside of the Godot editor, so I think they shouldn't be removed automatically.

```
func _process(delta):
	time += delta
	
	print(time)
```

The empty line above is still part of the `_process` body, which the tab semantically signifies.

PR #484, opened 2020-06-22
Add an operating system testing demo

### LINKED ISSUE #8: OS test demo
Moved from https://github.com/godotengine/godot/issues/3142:

> The OS class provides an abstraction layer over the platform-dependent code.
> The engine now has code for something like 13 platforms.
> It would be nice to have a demo that tests all (or most) of the OS class methods.
> This way it would be much more easier to test and compare different platform implementations.
> And also find bugs in the platform-dependent code.
> It will also be very helpful when adding support for a new platform.

Calinou:
This demo showcases various OS-specific features in Godot. It can be used to test Godot while porting it to a new platform or to check for regressions.

This closes #8.

## Preview

OS Testing Demo


misc/os_test/actions.gd
```diff
@@ -0,0 +1,78 @@
...
+		# Windows-specific.
+		path = OS.get_environment("USERPROFILE")
+
+	OS.shell_open(path)
+
+
+func _on_ChangeWindowTitle_pressed():
+	OS.set_window_title(
+			"Modified window title. Unicode characters for testing: é € × Ù ¨"
+	)
```
aaronfranke, line 20: I think this should just be one line. There are already other similar lines in this script that are longer than 100 characters. The line length guideline in GDScript exists to discourage writing bad code, but this is just a long string so IMO it's 100% fine to have it all on one line.

This also applies to line 69, and `os_test.gd` line 139. Everything else seems good.

Calinou:
Rebased on top of the latest commits with suggestions applied.

PR #486, opened 2020-06-24
Add Mono C# preprocessor defines, README, and screenshots to OS testing demo

aaronfranke:
If a non-Mono version of Godot opens, the project still works fine (it will just say "Mono module enabled: No" and not have Mono-specific information).

128-bit and Haiku currently aren't possible, but I figured why not future proof it? :)

Preview the README here: https://github.com/aaronfranke/godot-demo-projects/blob/os-test/misc/os_test/README.md

The second screenshot shows the Mono information in action.


aaronfranke:
Heh, I just noticed the issue number, it's perfect for this PR :tada:

PR #493, opened 2020-07-02
Disable antialiasing on the font in the 2D platformer demo

Calinou:
Pixel fonts look best when antialiasing is disabled.

### Before

Antialiasing

### After

No antialiasing


aaronfranke:
Thanks!

PR #494, opened 2020-07-02
Progressively interrupt the jump in the 2D platformer demo

Calinou:
This feels more smooth while still giving a sense of control to the player.


aaronfranke:
Thanks!

PR #495, opened 2020-07-02
Godot 3.2 IK demo SkeletonIK error fix

TwistedTwigleg:
Fixed bug in the IK demo where the SkeletonIK scene would print an error. The issue is that the target nodes are initialized after the SkeletonIK node. Moving the nodes fixed the issue and now no more errors are printed.


aaronfranke:
Thanks!

PR #497, opened 2020-07-08
Add memory usage to the Voxel demo

aaronfranke:
I recently added this to the TPS demo's debug menu, so I figured I would add it to the Voxel demo too.


volzhs:
isn't it better to use `String.humanize_size(bytes)`? just my 2 cents.

aaronfranke:
`humanize_size` always outputs 4 digits, for this demo I think it's better to only display increments of one MiB.

But thanks for bringing that up, and I will likely use that in the future.

volzhs:
Oh i see :)

PR #498, opened 2020-07-24
Mono Android IAP Demo Project

ricardoalcantara:
Hi, I would like to add this demo project I created with C#. I translated the new GodotGooglePlayBilling plugin implementation to C# calls and objects, I also made a small flow of a purchase using the fake ids google provides for testing purpose.


aaronfranke:
@timoschwarzer Can you take a look at this?

@ricardoalcantara There is a minor formatting problem, you can fix it by running `format.sh`.

mono/android_iap/project.godot
```diff
@@ -0,0 +1,35 @@
...
+config_version=4
+
+_global_script_classes=[  ]
+_global_script_class_icons={
+
+}
+
+[application]
+
+config/name="Android_Iap_with_C#"
```
aaronfranke, line 18:
```suggestion
config/name="Android IAP with C#"
```

Changing this name also requires changing or regenerating the csproj/sln files.

mono/android_iap/GodotGooglePlayBilling/GooglePlayBilling.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            GD.Print($"GodotGooglePlayBilling PurchaseAcknowledgementError error {code}: {message}");
+            EmitSignal(nameof(PurchaseAcknowledgementError), code, message);
+        }
+        private void OnGodotGooglePlayBilling_purchase_consumed(string purchaseToken)
+        {
+            GD.Print($"GodotGooglePlayBilling PurchaseConsumed successfully: {purchaseToken}");
+            EmitSignal(nameof(PurchaseConsumed), purchaseToken);
+        }
+        private void OnGodotGooglePlayBilling_purchase_consumption_error(int code, string message, string purchaseToken)
```
aaronfranke, line 144: Place an empty line between functions.

mono/android_iap/GodotGooglePlayBilling/GooglePlayBilling.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            GD.Print($"GodotGooglePlayBilling PurchaseConsumption_error error {code}: {message}, purchase token: {purchaseToken}");
+            EmitSignal(nameof(PurchaseConsumption_error), code, message, purchaseToken);
+        }
+        #endregion
+
+    }
```
aaronfranke, line 151: Avoid empty lines before closing braces (or after opening braces).

mono/android_iap/GodotGooglePlayBilling/GooglePlayBilling.cs
```diff
@@ -0,0 +1,152 @@
+using Godot.Collections;
+using Godot;
+
+namespace Android_Iap.GodotGooglePlayBilling
+{
+    public enum PurchaseType
+    {
+        inapp,
+        subs
+    }
```
aaronfranke, line 10: Enum members should be PascalCase in C#

mono/android_iap/GodotGooglePlayBilling/BillingResult.cs
```diff
@@ -0,0 +1,67 @@
+using Godot;
+using Godot.Collections;
+
+namespace Android_Iap.GodotGooglePlayBilling
+{
+    // https://developer.android.com/reference/com/android/billingclient/api/BillingClient.BillingResponseCode
+    public enum BillingResponseCode
+    {
+        // The request has reached the maximum timeout before Google Play responds.
+        SERVICE_TIMEOUT = -3,
```
aaronfranke, line 10: Enum members should be PascalCase in C#

mono/android_iap/Android_Iap_with_C#.csproj
```diff
@@ -0,0 +1,65 @@
...
+    <Reference Include="GodotSharpEditor" Condition=" '$(Configuration)' == 'Debug' ">
+      <Private>False</Private>
+      <HintPath>$(ProjectDir)/.mono/assemblies/$(ApiConfiguration)/GodotSharpEditor.dll</HintPath>
+    </Reference>
+    <Reference Include="System" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="*.cs" />
+    <Compile Include="Properties\*.cs" />
+    <Compile Include="GodotGooglePlayBilling\*.cs" />
```
aaronfranke, line 62: For Godot 3.2, use explicit file includes for each file rather than wildcards, for improved compatibility.

When Godot 4.0 comes out, we will switch to the new csproj format which allows writing a single `Include` glob.

ricardoalcantara:
Ouch, it looks like the README.md also have that eof new line rule.
Trust me I read that link before but I missed a few things, I will fix that.

mono/android_iap/project.godot
ricardoalcantara: Usually I avoid all type of space because I have had trouble before, gonna fix it!

mono/android_iap/GodotGooglePlayBilling/GooglePlayBilling.cs
ricardoalcantara: I used this way because I use the string literal.
The plugin has a method
```Python
payment.querySkuDetails(["my_iap_item"], "inapp")
```
So I translated with enum and to avoid use if InApp to inapp I did that
```C#
_payment?.Call("querySkuDetails", querySkuDetails, $"{type}");
```
Do I really need to change that?

mono/android_iap/GodotGooglePlayBilling/BillingResult.cs
ricardoalcantara: In this case I used those enum from google docs I can change it to be like the guideline, usually I prefer to keep the reference from the source https://developer.android.com/reference/com/android/billingclient/api/BillingClient.BillingResponseCode

mono/android_iap/Android_Iap_with_C#.csproj
ricardoalcantara: Ok, I didn't know that, I will change that.

ricardoalcantara:
I've done all the suggested changes. Sorry for the inconvenience, I am not used to work with guidelines, I've been a wild dev for longer than I would like, I am a bit rusty.

aaronfranke:
When you added those empty lines, you accidentally left trailing spaces.

ricardoalcantara:
Done again. I have been doing squash and force push because I read once someone from the Godot Core Devs to do it, but if you prefer I can do normal push too, this is not my ordinary way to commit things.

aaronfranke:
Squash and force-push is the recommended workflow, you are doing it correctly. @timoschwarzer Can you take a look at this?

ricardoalcantara:
Made another commit, Another 'Enum members should be PascalCase in C#' from Purchase.cs

mono/android_iap/GodotGooglePlayBilling/GooglePlayBilling.cs
aaronfranke: A bit late for me commenting this, but... Enums are not strings, so I don't see a strong reason to use `inapp` when we can just use the PascalCase `InApp` and follow the style guide. If we needed to convert an enum to a string we could have a method with something like `case PurchaseType.InApp: return "inapp";`

mono/android_iap/GodotGooglePlayBilling/BillingResult.cs
aaronfranke: In this situation, a case conversion is trivial, there is no confusion that can arise from people expecting `SERVICE_TIMEOUT` and only finding `ServiceTimeout`, so it's best to just follow the style guide here too.

ricardoalcantara:
After those style fixes I broke the SkuDetails parse, but it's now fixed.

PR #499, opened 2020-07-24
3D NavMesh Demo - fix for navmesh (re)baking issue

### LINKED ISSUE #472: Navmesh Project doesn't work after rebaking.
For the Nav Mesh Demo on my Windows 10 device. When ever I download the demo and click the Bake NavMesh button in the NavMesh node the demo doesn't work anymore. The Agent stays in place regardless of where ever I click on the screen in game.

Those are all the steps it takes for me to reproduce this bug. I didn't touch anything else in the demo. The demo works if I don't touch the button on the other hand. I discovered this bug because when I was trying to make my own NavMesh the same thing would happen regardless of my children set up. The only thing is that replicating it in this demo is a little weird since one would imagine it baked just fine before the demo was uploaded but doesn't work for me when I re-bake it in my computer.

It gives this bug when ever it runs also "ERROR: _navmesh_unlink: Condition "!nm.linked" is true. At: scene/3d/navigation/cpp:144"

smix8:
3D NavMesh Demo - fix for navmesh (re)baking issue. Fixes #472 
- changed the level mesh to a child of the navmesh so "Bake NavMesh" can find it
- changed level mesh size/scale (made 4x times larger, was way to small to be useable on a 1m=1unit scale)
- changed navmesh agent settings for the size of the robot. (default radius was to large, slope values to small, climp value to high)
- changed camera position and movement speed value to account for the larger scale.


3d/navmesh/navmesh.gd
```diff
@@ -55,7 +55,7 @@ func _unhandled_input(event):
 		var from = get_node("CameraBase/Camera").project_ray_origin(event.position)
 		var to = from + get_node("CameraBase/Camera").project_ray_normal(event.position) * 100
 		var p = get_closest_point_to_segment(from, to)
-		
+
```
aaronfranke, line 58: Please get rid of these indentation changes.

3d/navmesh/navmesh.tscn
```diff
@@ -3,38 +3,41 @@
 [ext_resource path="res://navmesh.gd" type="Script" id=1]
 [ext_resource path="res://default_env.tres" type="Environment" id=3]
 
-[sub_resource type="SpatialMaterial" id=1]
+[sub_resource type="NavigationMesh" id=5]
```
aaronfranke, line 6: The IDs start at 5, when they should start at one. To fix, simply open and re-save the scene.

3d/navmesh/navmesh.gd
```diff
@@ -66,7 +66,7 @@ func _unhandled_input(event):
 			get_node("CameraBase").set_rotation(Vector3(0, camrot, 0))
 			print("Camera Rotation: ", camrot)
 
-
+
```
aaronfranke, line 69: This should also be removed.

smix8:
I agree that 4x scale was a bit overkill. The navmesh still had holes with 2x before I tweaked the agent settings and I left it.
I reduced it to 2x scale now to not hurt anyones eyes to much.

I would like to learn how to squash the commits myself but I am a bit lost rightnow how to do it. I read the command line instructions but as a github desktop peasants I am still searching for the right lead how to get the command line working on windows.

aaronfranke:
Don't use GitHub Desktop, it is a terrible Git client. Either use [command like Git](https://git-scm.com/downloads) or [GitKraken](https://www.gitkraken.com/) ([tutorial](https://www.youtube.com/watch?v=0XKy8YpP5Ww)).

PR #502, opened 2020-07-26
3D Minimap / Radar Demo

smix8:
Demo for Minimap / Radar overlay using renderlayers and viewport camera.

Can be used for both common corner minimap or peeping radar like in alien vs predator games.
Works for both horizontal and vertical placement of environment objects and actors.

Used the robot mesh from the navmesh demo for the actors.
Didn't spend much time to prettify it (but could) to keep filesize small and for GLES2 compatibility.


aaronfranke:
I don't like the approach that the main scene uses. There's a ton of fog, the player does not have any control over movement, and is immediately rotating and moving up/down. This is disorientating and makes it difficult for users to study the demo.

Perhaps instead a minimap demo should be part of a larger real time strategy demo, where users can select units, move units, freely move around and zoom the main camera, and have a minimap for the whole scene. If this were implemented, we could replace the navmesh demo, since the real time strategy demo would use a navmesh. It might also make sense to implement "fog of war" so that only enemies near to your units are visible, and this would have to be shown visually on the terrain and minimap.

smix8:
I wouldn't mind adding more stuff like a full player controller for movement, moving enemies or a more detailed, explorable environment or even craft an entire new demo with more features integrated like suggested.

Here I was afraid it would be to much as I read one requirement for demos was to keep it (very?) simple and focused on a core feature so I mimicked the (art)style of other demos.

The fog part was added with the intention to showcase that the radar shows more than the user can see.
The automatic up/down/rotation was added to keep the demo focused and only showcase the change of the map with the height and rotation.

aaronfranke:
I'm going to close this for now, and mark it as salvageable. It would be nice to have a demo for this, but I think we should first discuss what kind of demo we want to have to best showcase this. See the discussion [here](https://github.com/godotengine/godot-docs/issues/1070).

PR #504, opened 2020-07-28
Android IAP Response Code can now return null.

ricardoalcantara:
After this PR https://github.com/godotengine/godot-google-play-billing/pull/6 "Response Code" can return null.
I made a tine update on the GDscript since if it comes with null vale the interpolation with %d fails. And I chanced the C# version to nullable enum. And I update the messages for test purpose.

### DIFF
--- a/mobile/android_iap/iap_demo.gd
+++ b/mobile/android_iap/iap_demo.gd
@@ -116,7 +116,7 @@ func _on_QuerySkuDetailsButton_pressed():
 func _on_PurchaseButton_pressed():
 	var response = payment.purchase(TEST_ITEM_SKU)
 	if response.status != OK:
-		show_alert("Purchase error %d: %s" % [response.response_code, response.debug_message])
+		show_alert("Purchase error %s: %s" % [response.response_code, response.debug_message])
 
 
 func _on_ConsumeButton_pressed():
--- a/mono/android_iap/GodotGooglePlayBilling/BillingResult.cs
+++ b/mono/android_iap/GodotGooglePlayBilling/BillingResult.cs
@@ -51,7 +51,7 @@ namespace Android_Iap.GodotGooglePlayBilling
             try
             {
                 Status = (int)billingResult["status"];
-                ResponseCode = (billingResult.Contains("response_code") ? (BillingResponseCode)billingResult["response_code"] : BillingResponseCode.Ok);
+                ResponseCode = (billingResult.Contains("response_code") ? (BillingResponseCode?)billingResult["response_code"] : null);
                 DebugMessage = (billingResult.Contains("debug_message") ? (string)billingResult["debug_message"] : null);
             }
             catch (System.Exception ex)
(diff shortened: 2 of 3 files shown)


aaronfranke:
CC @timoschwarzer

ricardoalcantara:
I think that the GDScript demo is simpler. When I built the C# version I was focusing on implement the flow I would use on my real game.

timoschwarzer:
The C# demo contains more boilerplate code to make the overall project more clean (contains all the enums etc.) but users don't need most of that at the end. I think a simpler demo project would be more appropriate for the sake of just understanding how to use the billing plugin.

timoschwarzer:
Like, the current C# demo is great if you have a basic understanding of how the billing plugin works, because it provides all the status code constants etc. but I think if you have no clue about how to use the billing plugin, the GDScript demo provides a more clear starting point.

Maybe we can make the C# demo more simple but can keep providing the boilerplate code somewhere? Maybe just a separate file?

ricardoalcantara:
>The C# demo contains more boilerplate code to make the overall project more clean (contains all the enums etc.) but users don't need most of that at the end.

I think they need, I program with C# and it's extremely painful to work direct with the Godot.Object and Cast everything think to get value so for me have it encapsulated makes more sense, that's why I create the demo project to share it.

>The C# demo contains more boilerplate code to make the overall project more clean (contains all the enums etc.) but users don't need most of that at the end.

For me the GDScript didn't help, I came from zero and for me it was only a helper file, I couldn't even consider as a demo project, almost the same from the documentation with the extra three GUI methods.

I could replicate the flow in the Main.cs file to be like the GDScript, but for me the interface I created make a whole more sense for C# users. 

For me demo project it a project that really demonstrate something that people could use in project, people come to demo project do copy and paste the snippets and work most of the time.

aaronfranke:
I'm more concerned about things like the health bar, potions, and button layout differences. Enums are fine and often very helpful.

ricardoalcantara:
@aaronfranke I can update the GDScript Version if you think the demo project I made is useful, and I can also change the C# Version to be like the GD Version as well, the layout part.

My intention was to demonstrate how a game store would be implemented (a small one since I don't have more test products do demonstrate it), not to only demonstrate the API calls. I am open to suggestion.

aaronfranke:
I'm gonna merge this since it's an improvement, but a follow-up PR regarding the above would be nice.

timoschwarzer:
@ricardoalcantara In my opinion it would be awesome if you could strip off some stuff from the C# demo that's not related to IAP (health bar etc.) and try to make the C# and GDScript versions match as closely as possible :)

ricardoalcantara:
The health bar is really an issue. Well I just wanted to demonstrate how purchase an item could affect the game and in what part of the flow you change it...

@aaronfranke and @timoschwarzer I will remove all those things and create another PR, it will be the same flow from GDScript, I will just keep the plugin interface.

PR #505, opened 2020-08-04
Mobile multitouch view fix

pkruszynski:
Multitouch View demo in the Mobile section has problems when multiple fingers are used.

To replicate this problem place your thumb on the device screen and touch the screen with any other finger of the same hand. Lift your thumb and touch the screen with it again - you should see some dot flickering when both fingers are touching the screen now (it helps to make the dots a bit bigger than size 40 as your fingers might be covering them).

This Pull Request:
- Fixes the problem described above
- Makes demo code clear and concise
- Makes pointer property naming for this demo consistent with documentation (https://docs.godotengine.org/en/stable/classes/class_inputeventscreentouch.html#class-inputeventscreentouch-property-index)
- Fixes `Readme.md` and `screenshots/multitouch.png` files only mentioning/showing red dots (instead of colored dots this demo actually displays)

This fix has been tested on a real Android device.


aaronfranke:
@pkruszynski Can you also fix this for the VisualScript version? `TouchHelper.gd` is still GDScript, so that part can be copy/pasted.

pkruszynski:
@aaronfranke - I updated the `TouchHelper.gd` file in `visual_script/multitouch_view` directory as requested.

pkruszynski:
@aaronfranke - Commits has been squashed.

PR #506, opened 2020-08-04
Add a GUI theming override demo

Calinou:
Overriding theme items is a common point of confusion. This demo should hopefully make it easier to understand.

In the future, we could extend this demo to include things like theme switching as it's a common need in non-game applications.

See also https://github.com/godotengine/godot/pull/41018.


gui/theming_override/test.gd
```diff
@@ -0,0 +1,69 @@
+extends Control
+
+# This script demonstrates how to alter StyleBoxes at runtime.
+# Custom theme item properties aren't considered Object properties per se.
+# This means that you should use `add_stylebox_override("normal", ...)`
+# instead of `set("custom_styles/normal", ...)`.
+
+onready var label = $VBoxContainer/Label
+onready var button = $VBoxContainer/Button
+onready var button2 = $VBoxContainer/Button2
```
aaronfranke, line 10: This is nitpicking, but if there are numbered items I usually prefer the first one to also be numbered. So it would be "Button1" and "Button2", not "Button" and "Button2".

gui/theming_override/test.gd
```diff
@@ -0,0 +1,69 @@
...
+	button.add_stylebox_override("hover", null)
+	button.add_stylebox_override("pressed", null)
+
+	button2.add_stylebox_override("normal", null)
+	button2.add_stylebox_override("hover", null)
+	button2.add_stylebox_override("pressed", null)
+
+	# If you don't have any references to the previous color value,
+	# you can instance a node at runtime to get this value.
+	var default_label_color = Label.new().get_color("font_color")
```
aaronfranke, line 68: Is this really the best way? Maybe there should be an API for this in Godot.

aaronfranke:
Thanks. I look forward to seeing this demo improved in the future, theme switching would be nice.

PR #507, opened 2020-08-12
Rewriting Mono Android IAP to match with GDScript

ricardoalcantara:
As discussed here https://github.com/godotengine/godot-demo-projects/pull/504
I tried to match as much as possible with the GDScript version.

I renamed the project to be the '{same name} With C#' and changed the namespace for the new solution and project file.

I tested on my phone and everything is still working, I tested on my PC to check if it's not broken in other platforms.


aaronfranke:
CC @timoschwarzer

aaronfranke:
@ricardoalcantara You have a case mismatch, as-is this does not open on my case-sensitive Linux system (I'm surprised it works on your phone - I didn't test on one). Please rename `Main.tscn` to `main.tscn` (also the root node should be `Main`).

ricardoalcantara:
I did rename it before, IMO it should be `Main.tscn`, the last version was and the C# class is `Main` too, is it some guideline I've missed? The `project.godot` indeed is wrong but it has worked for me on my phone, but this could be some issue with windows, we should never trust windows when renaming only changing the case, maybe it didn't truly changed it in the end.
Can it all be `Main `or the scene needs to be `main`?

aaronfranke:
There is no hard rule yet, see the discussion [here](https://github.com/godotengine/godot-docs/issues/3104). If you want to name it `Main.tscn` that's fine too since there is no hard rule yet (but you need to update the `project.godot` file). Plus this kind of thing is easy to change later *en masse* if we do make a rule. I only suggested snake_case for now to be consistent with Godot's source code.

ricardoalcantara:
@aaronfranke  Thanks for the links, as I imagined on my windows it's lower case already... I will have find a workaround.
Windows shows lowercase
image
Git shows uppercase
image

ricardoalcantara:
@aaronfranke Done, I changed to ain.tscn, commit, changed to main.tscn, commit, squashed it, Deal with it Windows!!!

I hope this solves the problem!

aaronfranke:
Thanks!

PR #508, opened 2020-08-16
Tweak settings in various 3D demos

Calinou:
- Enable 4× MSAA in all 3D demos.
- Enable filmic tonemapping with a whitepoint of 6 in GLES3 3D demos.
- Add subtle fog to the Truck Town demo.
- Use the `2d` stretch mode and `expand` stretch aspect to support multiple resolutions and aspect ratios.


3d/platformer/project.godot
```diff
@@ -143,4 +143,3 @@ multithread/thread_rid_pool_prealloc=60
 
 quality/intended_usage/framebuffer_allocation=3
 quality/filters/msaa=2
-environment/default_environment="res://default_env.tres"
```
aaronfranke, line 146: You broke the sky! :P

Calinou:
It's surely possible to make the tonemapping result brighter by increasing the exposure slightly. I'll try increasing the exposure to 1.2 and see how it looks.

I've seen people complain about blown out highlights in that demo, so fixing them using filmic tonemapping is most likely worth it.

3d/platformer/project.godot
Calinou: Good catch, fixed.

Calinou:
> Maybe MSAA should be enabled by default for Godot 4.0?

Unfortunately, it's too expensive to do that (even 2× has a significant cost on slower GPUs). We could enable FXAA by default, but it introduces a significant amount of blur which is problematic in resolutions below 1440p (and especially below 1080p).

PR #511, opened 2020-08-18
Add a normal map test in the 2D lights and shadows demo

Calinou:
This was requested by lawnjelly so we can test whether normal map orientation behaves as expected.

The normal map was generated using Laigter (with the Y axis inverted).

### With batching (broken)

*Notice how all sprites but the top-left one have incorrect normal mapping. The Godot logo should appear as a "bump", not a "hole".*

2D lights and shadows

### Without batching (correct)

2D lights and shadows


2d/lights_and_shadows/light_shadows.tscn
```diff
@@ -186,11 +258,13 @@ occluder = SubResource( 1 )
 position = Vector2( 159.289, 452.441 )
 texture = ExtResource( 3 )
 color = Color( 1, 0.446392, 0.0576646, 1 )
-energy = 0.95
```
aaronfranke, line 189: I dimmed the lights to 95% in #414 to try and make the demo look similar to when it was using GLES 3. Is this tweak no longer necessary?

Calinou: Probably not, but I find that the demo looked a bit dark otherwise. (It made the normal mapping more difficult to notice.)

PR #512, opened 2020-08-19
Add a 3D waypoints demo

Calinou:
This is an example of displaying GUI elements such as Labels in a 3D world, without relying on viewports. This results in better readability and performance for use cases such as showing player names.

Some waypoints showcased in the demo will also snap to the window borders when outside the player's view.

No Sprite3D nodes are used in this demo.

## Preview

Waypoints

The freelook script is quite shoddy – we should probably aim to have a standard script we can copy-paste everywhere needed.

The way off-screen waypoints are displayed is also far from perfect. Still, displaying GUI elements in a 3D world without relying on viewports is a common question on community platforms. It's better than nothing to have a demo about it, even if it's only partially working.


3d/waypoints/waypoint.svg
```diff
@@ -0,0 +1 @@
+<svg height="128" viewBox="0 0 33.866666 33.866666" width="128" xmlns="http://www.w3.org/2000/svg"><path d="m4.2333324 3.1749994h25.3999996l-12.699999 24.3416666z" fill="#fff" stroke="#000" stroke-width="2.11667"/></svg>
```
aaronfranke, line 1: Run the formatting script or add a newline here.

3d/waypoints/camera.gd
```diff
@@ -0,0 +1,61 @@
...
+		motion *= 2
+
+	# Rotate the motion based on the camera angle.
+	motion = motion \
+		.rotated(Vector3(0, 1, 0), rotation.y - initial_rotation) \
+		.rotated(Vector3(1, 0, 0), cos(rotation.y) * rotation.x) \
+		.rotated(Vector3(0, 0, 1), -sin(rotation.y) * rotation.x)
+
+	# Add motion, apply friction and velocity.
+	velocity += motion * move_speed * delta
```
aaronfranke, line 59: You are way overcomplicating this :)

```suggestion
	velocity += move_speed * delta * transform.basis.xform(motion)
```

aaronfranke:
Merged as #567.

PR #524, opened 2020-09-30
Allow hiDPI in all demos that support multiple resolutions

Calinou:
This is required to benefit from crisp display on hiDPI monitors. This also fixes issues related to fullscreen and input handling when using an hiDPI display on Windows.

Command used:

```bash
# Install sd <https://github.com/chmln/sd> then run:
sd 'window/stretch/mode="2d"' 'window/dpi/allow_hidpi=true\nwindow/stretch/mode="2d"' **/*.godot
# After running `sd`, remove the duplicate line in `2d/platformer/project.godot`.
```

### DIFF
--- a/2d/finite_state_machine/project.godot
+++ b/2d/finite_state_machine/project.godot
@@ -26,6 +26,7 @@ config/icon="res://icon.png"
 
 window/size/width=1280
 window/size/height=720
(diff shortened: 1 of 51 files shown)


aaronfranke (approved):
I can't test this (I don't own a hiDPI display), so I'm going to assume it works.

PR #531, opened 2020-10-03
Update dodge_the_creeps to Godot 3.2.3 format

van800:


Duehok:
Copy of what I wrote on discord:
 You missed one nameof in Main.cs line 88
```csharp
  GetNode<HUD>("HUD").Connect("StartGame", mobInstance, nameof(Mob.OnStartGame));
```
could be
```csharp
GetNode<HUD>("HUD").Connect(nameof(HUD.StartGame), mobInstance, nameof(Mob.OnStartGame));
```

mono/dodge_the_creeps/Mob.cs
```diff
@@ -9,10 +9,10 @@ public class Mob : RigidBody2D
     [Export]
     public int maxSpeed; // Maximum speed range.
 
-    private String[] _mobTypes = { "walk", "swim", "fly" };
+    private readonly string[] _mobTypes = { "walk", "swim", "fly" };
 
-    // C# doesn't have GDScript's random methods, so we use System.Random insetad.
-    private static Random _random = new Random();
+    // C# doesn't have GDScript's random methods, so we use System.Random instead.
```
31, line 14: Don't we have `GD.Rand*` methods? The comment from `Main.cs` seems better:

```suggestion
    // We use 'System.Random' as an alternative to GDScript's random methods.
```

mono/dodge_the_creeps/.gitignore
```diff
@@ -0,0 +1,11 @@
+# Godot auto generated files
+*.gen.*
+.import/
+.mono/
+
+# Jetbrains IDEs
+.idea/
```
31, line 7: Add VS, VS Code ignores too?

https://github.com/github/gitignore/blob/master/VisualStudio.gitignore has a bunch of stuff, but maybe at least:

```
.vs/

_ReSharper*/
*.[Rr]e[Ss]harper
*.DotSettings.user
```

Not sure about VS Code. (If the extensions used there even put metadata in the project.)

aaronfranke: No, all of these should be removed. We shouldn't give treatment to special editors, and we already have a `.gitignore` and `.gitattributes` in the root of the demo projects repository.

mono/dodge_the_creeps/Dodge the Creeps with C#.csproj
```diff
@@ -1,26 +1,9 @@
...
-    <!--The following properties were overriden during migration to prevent errors.
-    Enabling them may require other manual changes to the project and its files.-->
-    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-    <Deterministic>false</Deterministic>
+    <TargetFramework>netstandard2.1</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="System" />
+    <Content Include="**/*.tscn"/>
```
aaronfranke, line 7: Why are you trying to compile `.tscn` files?

mono/dodge_the_creeps/Main.cs
```diff
@@ -66,10 +65,10 @@ public void OnMobTimerTimeout()
 
         // Choose a random location on Path2D.
         var mobSpawnLocation = GetNode<PathFollow2D>("MobPath/MobSpawnLocation");
-        mobSpawnLocation.SetOffset(_random.Next());
+        mobSpawnLocation.Offset = _random.Next();
 
         // Create a Mob instance and add it to the scene.
-        var mobInstance = (RigidBody2D)mob.Instance();
+        var mobInstance = (RigidBody2D) mob.Instance();
```
aaronfranke, line 71: https://docs.godotengine.org/en/latest/getting_started/scripting/c_sharp/c_sharp_style_guide.html

The style guide says no space after type cast.

aaronfranke (changes requested):

mono/dodge_the_creeps/Dodge the Creeps with C#.csproj
van800: not to Compile, just show in csproj. I can remove it, if you like.

van800: I made it None, would it be better?

aaronfranke:
@van800 I pushed some additional commits, let me know if they look good to you.

Please try to keep in mind that all of the C# demos are mirrors of GDScript demos, so they should be as similar as possible. For Dodge the Creeps in particular, the GDScript demo mirrors [this documentation](https://docs.godotengine.org/en/stable/getting_started/step_by_step/your_first_game.html), so if you want to make major changes then they would need to be changed in both demos and in the documentation.

Also, I noticed that the `csproj` in the demo as of this PR vs in a newly generated empty project is different. The demo as of this PR has `<AssemblyName>` which a newly generated project lacks, and a newly generated project has `<RootNamespace>` which the demo as of this PR lacks. Also, here we have `<TargetFramework>` as `netstandard2.1`, while a newly generated project has `net472`. Not sure what should be changed, but it seems strange that these are inconsistent. I think `<TargetFramework>` as `netstandard2.1` is preferred, but then why does Godot generate `net472`? CC @neikeq

van800:
@aaronfranke 
1. Thank you so much! I wasn't able to find place to use min_speed and max_speed.

2. Regarding `net472`, dotnetcore msbuild doesn't like it
```
8:49	Project 'Dodge the Creeps with C#' load finished with warnings: [MSB3644] The reference assemblies for .NETFramework,Version=v4.7.2 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks at (1177:5)
```
And `netstandard2.1` looks good with both dotnetcore and mono msbuilds.

3. Removing AssemblyName doesn't change anything. Feels safe to be removed.

4. I can add RootNamespace, not sure, what would be better.

5. I want to add `#pragma warning disable 649` around `private PackedScene _mobScene;`. What do you think?

6. Removing `<None Include="**/*.tscn" />` seems wrong thing to me. For Rider and VS showing scenes in the project helps to find usages in them. However I don't insist on it.

aaronfranke:
2\. Let's keep `netstandard2.1`, but we should probably change this in Godot itself too so that it generates instead of `net472`.

4\. Since it works fine without it, I'm leaning towards keeping it excluded, but I'm interested in @neikeq's thoughts.

5\. Oh, I didn't know that you made this change to avoid a warning. Perhaps we should keep `[Export]` but give it back the default value of ` = ResourceLoader.Load<PackedScene>("res://Mob.tscn");`.

6\. Oh, that is required to show the files in Rider and VS? Hmm, it's not required in VS Code, and I'm surprised that Rider and VS don't have a way to view all files (I don't use those IDEs personally). Anyway, I'm not sure that it's actually a 100% benefit to show them in Rider in VS since after all they aren't C# files, so I think we should leave it out.

van800:
5. Having it both ways at once feels too confusing.

aaronfranke (approved):
Thank you for doing work on this, it's much appreciated. I did somewhat end up commandeering this PR as a general "fix Dodge the Creeps" PR, and I opened up https://github.com/godotengine/godot-docs/pull/4208 as a companion to this PR, such that if both are merged we end up with consistency (at least in the code examples and the code in the demos).

What's left is that this PR needs to be squashed down to one commit since some of the later commits revert work done in earlier commits. After this is merged, it would be nice to update the other demos (the other 3 in this folder likewise have accompanying GDScript demos, but don't have documentation articles like DTC, and there is also the OS test demo which only has a tiny amount of C# and it's not in the mono folder).

van800:
Force pushed

aaronfranke (approved):
Btw, if you make follow-up PRs, I suggest making a separate branch (anything that isn't "master").

PR #533, opened 2020-10-08
Improve Dodge the Creeps demo

van800:
1. restore net472:
```
Sdk is referencing Microsoft.NETFramework.ReferenceAssemblies: https://github.com/godotengine/godot/blob/cc3c671f3b636cbfac9fc74095d5bee753e71ed5/modules/mono/editor/Godot.NET.Sdk/Godot.NET.Sdk/Sdk/Sdk.targets#L18-L33
```
I was mistaken previously because of this one time warning in Rider https://youtrack.jetbrains.com/issue/RIDER-52053

2. restore RootNamespace, since default one contains spaces.
Using default one would cause new added script be invalid, at least as I tried in Rider


aaronfranke (approved):
Squashing this would be nice but this looks good.

van800:
Squashed

aaronfranke (approved):

PR #536, opened 2020-10-10
Change Dodge the Creeps to be more consistent with the docs

aaronfranke:
I will update this PR if I need to make any further changes during the review of https://github.com/godotengine/godot-docs/pull/4208

### DIFF
--- a/mono/dodge_the_creeps/Main.cs
+++ b/mono/dodge_the_creeps/Main.cs
@@ -5,10 +5,10 @@ public class Main : Node
 #pragma warning disable 649
     // We assign this in the editor, so we don't need the warning about not being assigned.
     [Export]
-    private PackedScene _mobScene;
+    public PackedScene mobScene;
 #pragma warning restore 649
 
-    private int _score;
+    public int score;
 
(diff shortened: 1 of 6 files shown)


mono/dodge_the_creeps/Main.cs
```diff
@@ -5,10 +5,10 @@ public class Main : Node
 #pragma warning disable 649
     // We assign this in the editor, so we don't need the warning about not being assigned.
     [Export]
-    private PackedScene _mobScene;
+    public PackedScene mobScene;
 #pragma warning restore 649
 
-    private int _score;
+    public int score;
```
neikeq, line 11: I think the correct naming convention for public fields is PascalCase. It may be better to go with properties too.

aaronfranke: Keep in mind this is a beginner tutorial, so properties would just be adding complication for no benefit. Also, we've been using camelCase for public fields everywhere including the demos and `GodotSharp/Core`, and it's what I prefer since it distinguishes fields and properties.

PR #537, opened 2020-10-10
Update C# project files in C# demos for Godot 3.2.3

aaronfranke:
Updates all C# demos to have the C# project files be similar to Dodge the Creeps after @van800's work on that demo. This is mostly just removals, since many things aren't necessary to have in the C# project file as of Godot 3.2.3.


mono/2.5d/2.5D Demo with C#.csproj
```diff
@@ -1,31 +1,5 @@
-﻿<Project Sdk="Godot.NET.Sdk/3.2.3">
+<Project Sdk="Godot.NET.Sdk/3.2.3">
   <PropertyGroup>
-    <ProjectGuid>{5CA791DB-5050-44D0-989B-41D559AB1D50}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <RootNamespace>Empty.DMono</RootNamespace>
```
van800, line 5: Would be better to include RootNamespace here.

aaronfranke: Yeah, I wasn't sure what to put, so I just left it out. The name of the demo is "2.5D", so `TwoPointFiveD`?

van800: I would make it `Demo`

PR #540, opened 2020-10-15
Re-write the navigation path-following code to be easier to read + add comments

jak-sdk:
Hi,

I found the previous code that implemented the navmesh path following to be quite complex and a little difficult to follow.
I've tried to simplify it as much as I can and document what does what, and any considerations that went into it.

As far as I can tell it behaves identically to the previous code, I appreciate any second opinions though!

Please let me know if this is ok / if I've added too many comments, or any other recommendations.
I'm hoping this will be more approachable for newcomers to Godot.

Thanks,
Jak

### DIFF
--- a/3d/navmesh/navmesh.gd
+++ b/3d/navmesh/navmesh.gd
@@ -3,13 +3,13 @@ extends Navigation
 const SPEED = 10.0
 
 var camrot = 0.0
-
-var begin = Vector3()
-var end = Vector3()
 var m = SpatialMaterial.new()
 
 var path = []
-var draw_path = true
+var show_path = true
+
+onready var robot  = get_node("RobotBase")
+onready var camera = get_node("CameraBase/Camera")
 
 func _ready():
 	set_process_input(true)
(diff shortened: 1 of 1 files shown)


3d/navmesh/navmesh.gd
```diff
@@ -3,85 +3,83 @@ extends Navigation
...
+            draw_path(path)
+
+    if event is InputEventMouseMotion:
+        if event.button_mask & (BUTTON_MASK_MIDDLE + BUTTON_MASK_RIGHT):
+            camrot += event.relative.x * 0.005
+            get_node("CameraBase").set_rotation(Vector3(0, camrot, 0))
+            print("Camera Rotation: ", camrot)
+
+
+func draw_path(path):
```
aaronfranke, line 74: Function declaration of 'draw_path()' conflicts with a variable of the same name.

The local variable 'path' is shadowing an already-defined variable at line 8.

3d/navmesh/navmesh.gd
```diff
@@ -3,85 +3,83 @@ extends Navigation
...
-	else:
-		set_process(false)
+onready var robot  = get_node("RobotBase")
+onready var camera = get_node("CameraBase/Camera")
 
+func _ready():
+    set_process_input(true)
+    m.flags_unshaded = true
+    m.flags_use_point_size = true
+    m.albedo_color = Color.white
```
aaronfranke, line 18: Tabs are used for indentation in GDScript.

aaronfranke (changes requested):

3d/navmesh/navmesh.gd
```diff
@@ -18,47 +18,51 @@ func _ready():
...
+		robot.translation += direction.normalized() * step_size
+
+		# lastly lets make sure we're looking in the direction we're travelling
+		# clamp y to 0 so the robot only looks left and right, not up/down
+		direction.y = 0
+		# direction is relative, so apply it to the robot's location to
+		#	get a point we can actually look at
+		var look_at_point = robot.translation + direction.normalized()
+		# and look at it
+		robot.transform = robot.transform.looking_at(look_at_point, Vector3(0,1,0))
```
aaronfranke, line 52:
```suggestion
		robot.transform = robot.transform.looking_at(look_at_point, Vector3.UP)
```

3d/navmesh/navmesh.gd
```diff
@@ -18,47 +18,51 @@ func _ready():
...
+	var step_size = delta * SPEED
+
+	if path.size() > 0:
+		# direction is the difference between where we are now
+		#	and where we want to go
+		var destination = path[0]
+		direction = destination - robot.translation
+
+		# if the next node is closer than we intend to 'step' then take a smaller step
+		# otherwise we would go past it, and potentially through a wall or over a cliff edge!
```
aaronfranke, line 35: Please use proper sentence capitalization and punctuation. We should also try to avoid going over 80 characters wide in a single line. While shortening comments, if you can make the lines similar in length that would be ideal (else the last line should be shortest). For example, here I would do this:

```suggestion
		# If the next node is closer than we intend to 'step', then
		# take a smaller step. Otherwise we would go past it and
		# potentially go through a wall or over a cliff edge!
```

3d/navmesh/navmesh.gd
```diff
@@ -18,47 +18,51 @@ func _ready():
...
+func _physics_process(delta):
+	var direction = Vector3()
+
+	# we need to scale the movement speed by how much delta has passed
+	# otherwise the motion won't be smooth
+	var step_size = delta * SPEED
+
+	if path.size() > 0:
+		# direction is the difference between where we are now
+		#	and where we want to go
```
aaronfranke, line 30: Here you have a tab inside the comment, which is bad. There is one other instance in this file.

aaronfranke (changes requested):

3d/navmesh/navmesh.gd
jak-sdk: Sorry, I hadn't noticed my script editor had silently converted the whole thing to spaces..

3d/navmesh/navmesh.gd
jak-sdk: Ok, I'll commit this and apply a similar approach to the other comments, thanks

3d/navmesh/navmesh.gd
jak-sdk: Whoops.. Clearly some over zealous find and replace..

3d/navmesh/navmesh.gd
```diff
@@ -18,47 +18,53 @@ func _ready():
...
+		# If the next node is closer than we intend to 'step', then
+		# take a smaller step. Otherwise we would go past it and
+		# potentially go through a wall or over a cliff edge!
+		if step_size > direction.length():
+			step_size = direction.length()
+			# We should also remove this node since we're about to reach it.
+			path.remove(0)
+
+		# Move the robot towards the path node, by how far we want to travel.
+		# Note: for a kinetic body we would instead use move_and_slide
```
aaronfranke, line 43:
```suggestion
		# Note: For a KinematicBody, we would instead use move_and_slide
```

3d/navmesh/navmesh.gd
```diff
@@ -18,47 +18,53 @@ func _ready():
...
+		# so collisions work properly.
+		robot.translation += direction.normalized() * step_size
+
+		# Lastly lets make sure we're looking in the direction we're travelling.
```
aaronfranke, line 47:
```suggestion
		# Lastly let's make sure we're looking in the direction we're traveling.
```

aaronfranke (changes requested):

3d/navmesh/navmesh.gd
```diff
@@ -18,47 +18,53 @@ func _ready():
...
+		robot.translation += direction.normalized() * step_size
+
+		# Lastly let's make sure we're looking in the direction we're traveling.
+		# Clamp y to 0 so the robot only looks left and right, not up/down.
+		direction.y = 0
+		# Direction is relative, so apply it to the robot's location to
+		# get a point we can actually look at.
+		var look_at_point = robot.translation + direction.normalized()
+		# Make the robot look at the point.
+		robot.transform = robot.transform.looking_at(look_at_point, Vector3.UP)
```
aaronfranke, line 54: This code currently causes an error when pathing finishes, since `look_at_point` is the same as translation. This is due to `direction.normalized()` being zero. The fix here is to just check `if direction` to see if it's a non-zero vector. Note that we can't move this check anywhere else since we still need to process a Y-only movement, just not rotate for it. Also, Spatial has a shortcut for `robot.transform = robot.transform.looking_at`, we can just use `robot.look_at`.

```suggestion
		if direction:
			# Direction is relative, so apply it to the robot's location to
			# get a point we can actually look at.
			var look_at_point = robot.translation + direction.normalized()
			# Make the robot look at the point.
			robot.look_at(look_at_point, Vector3.UP)
```

aaronfranke (changes requested):

3d/navmesh/navmesh.gd
jak-sdk: Thanks, that's a good point.

aaronfranke:
The code looks good, and is a huge improvement in both readability and simplicity, thank you. The only thing left is to clean up the Git history and squash the commits into one. See [this documentation article](https://docs.godotengine.org/en/latest/community/contributing/pr_workflow.html) for more information. If you can't figure it out, GitHub does have a squash and merge feature, but it destroys history so it's better if the PR author does it.

jak-sdk:
Thanks :smile: 
I've squashed the commits, let me know if it's ok. 
Happy to have helped, in all honesty the demo as it was definitely helped me but I wanted to add comments and thought a little refactoring was worth doing too.
Thank you very much for your guidance and your help improving my code.

aaronfranke (approved):

aaronfranke:
Thanks!

One other thing, for next time, it's best practice to create a separate branch when making a PR, instead of just using master.

PR #552, opened 2020-11-12
Add automatic CI build to test each project

qarmin:
It will allow to find engine regressions, memory leaks etc.

I created small tool to automatically generate this CI - https://github.com/qarmin/GodotDemoProjectCICreator

### DIFF
--- a/.github/workflows/linux_default.yml
+++ b/.github/workflows/linux_default.yml
@@ -1,0 +1,599 @@
+name: 🐧 Linux Build Default
+on: [push, pull_request]
+
+jobs:
+  test-projects:
+    runs-on: "ubuntu-20.04"
+    name: Test demo projects
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Change sources.list
+        run: |
+          sudo rm -f /etc/apt/sources.list.d/*
+          sudo cp -f sources.list /etc/apt/sources.list
+          sudo apt-get update
+
+      - name: Download Godot
+        run: |
+          sudo apt-get install -y build-essential pkg-config libx11-dev libxcursor-dev libxinerama-dev libgl1-mesa-dev libglu-dev libasound2-dev libpulse-dev libudev-dev libxi-dev libxrandr-dev yasm wget2 unzip -y
+          wget2 https://downloads.tuxfamily.org/godotengine/3.2.3/Godot_v3.2.3-stable_x11.64.zip
+          unzip Godot_v3.2.3-stable_x11.64.zip
+          mv Godot_v3.2.3-stable_x11.64 godot
+
+      - name: loading/autoload
+        run: |
+          echo "" > sanitizers_log.txt
+          DRI_PRIME=0 timeout 25s xvfb-run ./godot --audio-driver Dummy -e    --path loading/autoload 2>&1 | tee -a sanitizers_log.txt || true
+          DRI_PRIME=0             xvfb-run ./godot --audio-driver Dummy -e -q --path loading/autoload 2>&1 | tee -a sanitizers_log.txt || true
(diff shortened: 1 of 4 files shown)


.github/workflows/static_checks.yml
```diff
@@ -1,19 +0,0 @@
-name: Static Checks
-on: [push, pull_request]
```
aaronfranke, line 2: Static Checks shouldn't be deleted.

aaronfranke (changes requested):

.github/workflows/static_checks.yml
qarmin: I know, but for now I don't need.
I will restore it when PR will be ready.

qarmin:
List of excluded projects
https://github.com/qarmin/GodotDemoProjectCICreator/blob/c71fab1b77979dfbc49fd11b9c10fc72c9d10939/src/main.rs#L130-L136

Manual testing shows that only `audio/mic_record` leak is reproducible(https://github.com/godotengine/godot/issues/34495)
Other crashes in CI from executed projects happened inside editor or when there was running
I think that there is an issue with Godot which doesn't properly handle importing multiple files.

`networking/webrtc_minimal` shows strange errors(entire log is polluted with it)
```

ERROR: create_data_channel: Condition "interface == __null" is true. Returned: __null
   At: modules/webrtc/webrtc_peer_connection_gdnative.cpp:79.
[Object:null]
ERROR: create_offer: Condition "interface == __null" is true. Returned: ERR_UNCONFIGURED
   At: modules/webrtc/webrtc_peer_connection_gdnative.cpp:84.
ERROR: _create: Default GDNative WebRTC implementation not defined.
   At: modules/webrtc/webrtc_peer_connection_gdnative.cpp:54.
ERROR: create_data_channel: Condition "interface == __null" is true. Returned: __null
   At: modules/webrtc/webrtc_peer_connection_gdnative.cpp:79.
```

visual_script/circle_pop shows error about 
```
ERROR: _disconnect: Nonexistent signal 'item_rect_changed' in [GraphNode:19172].
   At: core/object.cpp:1569.
```

This CI isn't pretty, but there is really easy to include/exclude projects from https://github.com/qarmin/GodotDemoProjectCICreator

Since checking projects with normal Godot takes 1h and with sanitizers support 2,5h I'm not sure if this should be used in each PR or as scheduled task executed every week/day.

aaronfranke:
@qarmin Due to the extreme length of time required, I think this would only make sense if it checked only the demos that changed in that particular PR. For example, if a PR changed the 2D platformer demo, it would only run checks for that demo.

For the excluded demos, the errors should eventually be fixed. I will investigate this eventually, but anyone is welcome to do so.

> I think that there is an issue with Godot which doesn't properly handle importing multiple files.

I know off hand that for the 2.5D demo, it always breaks the first time, because it tries to run the plugin script before the assets are imported. I opened [this bug report](https://github.com/godotengine/godot/issues/36713) to track this issue. In the meantime, maybe we can work around it, by having Godot open the project, import the files, then close it, then run the tests?

Also, it seems that the Mono demos aren't currently tested, nor is the Plugins demo project (that one is a top level folder, because it's really multiple separate "demos" in one, but they are stored together in one Godot project).

aaronfranke:
@qarmin Any update on this? It would be nice to have CI for building each project, but this needs a lot of work.

Calinou:
> Due to the extreme length of time required, I think this would only make sense if it checked only the demos that changed in that particular PR.

We can run this workflow in a weekly schedule so that it doesn't take time away from commits and pull requests.

qarmin:
I've been looking for a method to detect changes in specific repositories and only run specific actions in CI in those repositories, but unfortunately I haven't found anything useful.

qarmin:
Closing, since I have no idea how it could be implemented

PR #555, opened 2020-12-06
Add a control gallery demo

Calinou:
This project showcases what Godot's various Control nodes can do (and what they look like, so that new users can figure them out).

I could consider adding Tree in a future PR – probably using a VBoxContainer to put it on the right of the Lists panel.

## Preview

Control Gallery


aaronfranke:
Why not just improve the Control gallery that comes built into the engine for the theme editor?

Screenshot from 2020-12-06 12-06-52

Calinou:
@aaronfranke The built-in control gallery seems good enough, but I think it serves a different purpose.

The one I'm adding here is specifically meant to be referenced in the documentation as a screenshot, so that new users (who may not even know about the theme editor) know which Control nodes they should use to implement a given UI feature.

aaronfranke:
I changed my mind, I think this is a good idea. So now it's time to look for issues :)

I think the stretch mode should be set to "disabled" (and we can set aspect back to its default of "ignore").

When clicking on MenuButton and hovering or clicking on Actionitem, nothing happens and these errors appear:

```
E 0:00:04.231   get_node: Node not found: 1.
  <C++ Error>   Condition "!node" is true. Returned: __null
  <C++ Source>  scene/main/node.cpp:1381 @ get_node()

E 0:00:04.231   _activate_submenu: Item subnode does not exist: 1.
  <C++ Error>   Condition "!n" is true.
  <C++ Source>  scene/gui/popup_menu.cpp:151 @ _activate_submenu()
```

When pressing Control and hovering over LinkButton, nothing happens and these errors appear:

```
E 0:00:04.700   get_node: Node not found: 1.
  <C++ Error>   Condition "!node" is true. Returned: __null
  <C++ Source>  scene/main/node.cpp:1381 @ get_node()

E 0:00:04.700   get_node: Node not found: 3.
  <C++ Error>   Condition "!node" is true. Returned: __null
  <C++ Source>  scene/main/node.cpp:1381 @ get_node()
```

Calinou:
> I think the stretch mode should be set to "disabled" (and we can set aspect back to its default of "ignore").

Maybe, but this breaks the demo rendering on mobile and hiDPI displays where text will be tiny.

aaronfranke:
@Calinou It's not good as-is because it's not rendering at a higher resolution, it's just stretching, and it looks ugly:

Screenshot from 2020-12-08 22-04-56

If you want to implement a way for this to look good at high resolutions, that would be ideal (otherwise I would prefer no scaling to bad scaling). I took a look at the other demos and they seem to be able to handle scaling well, this is from the Translation demo:

Screenshot from 2020-12-08 22-02-58

Calinou:
@aaronfranke I could add a DynamicFont to this demo, but this means we have to add yet another TTF file to this repository, bloating it even further :stuck_out_tongue:

aaronfranke:
```
$ ls -ls *.otf
68 ... Titillium-BoldItalic.otf
56 ... Titillium-Bold.otf
68 ... Titillium-RegularItalic.otf
60 ... Titillium-Regular.otf
```

A few hundred kilobytes isn't a huge sacrifice for the sake of overcoming Godot's limitations, if we indeed desire high-DPI compatibility in this demo (and I guess all demos). Otherwise, I would just set stretch mode to disabled, and then we can worry about high-DPI compatibility in Godot 4.0 after https://github.com/godotengine/godot/pull/39603 solves this.

Calinou:
@aaronfranke Alright, I added a DynamicFont and updated the screenshot.

aaronfranke:
The text looks good now when scaled. However, the console errors I mentioned [here](https://github.com/godotengine/godot-demo-projects/pull/555#issuecomment-741281095) still exist.

aaronfranke:
Merged as #569.

PR #557, opened 2020-12-07
Simplify 3D in 2D demo

### LINKED ISSUE #556: 3D in 2D demo needlessly complicated (or outdated?)
3D in 2D

**Description:**
I clearly remember the first time I looked at the 3D in 2D demo. I thought it's super cool that Godot allows me to do this and that there is a demo available, but there seemed to be a lot of code involved I did not understand the purpose of.

Now that I have used Godot for more than two years instead of two days, I know this demo could showcase 3D in 2D without any code at all.

**All the user needs is a ViewportContainer (make sure to enable "stretch" property) with a Viewport node as child. Then add the 3D scene as child to the Viewport.**

The idea of a "Viewport Sprite" seems like a very custom solution to me (or maybe an old outdated way to do things?)
This code here I thing is way too specific and from my experience serves no real purpose if instead ViewportContainer+Viewport nodes are used:
```
func _ready():
	# We want Godot to load everything but be hidden for a bit.
	viewport_sprite.modulate = Color(1, 1, 1, 0.01)
	#warning-ignore:return_value_discarded
	get_viewport().connect("size_changed", self, "_root_viewport_size_changed")
	viewport_initial_size = viewport.size

	# Assign the sprite's texture to the viewport texture.
	viewport.set_clear_mode(Viewport.CLEAR_MODE_ONLY_NEXT_FRAME)

	# Let two frames pass to make sure the screen was captured.
	yield(get_tree(), "idle_frame")
	yield(get_tree(), "idle_frame")
	viewport_sprite.texture = viewport.get_texture()
	# Hide a little bit longer just in case.
	for _unused in range(50):
		yield(get_tree(), "idle_frame")
	viewport_sprite.modulate = Color.white # Default modulate color.
```

As a beginner, all it did was to confuse me, instead of showing off how simple it is to get the effect I want.
Having a 2D sprite in the same scene on the other hand is great, as it shows that 3D and 2D can coexist in the same scene without any problem. But that 2D Sprite too does not need code either. An AnimatedSprite node or Sprite+AnimationPlayer with Autoplay and loop activated is more than enough for a convincing effect. Same applies to the 3D scene.

aaronfranke:
Fixes #556. Testing would be appreciated. With Godot 3.2.3 on my system I don't experience the graphical artifacts that caused the delay to be added in the first place. It seems to also work fine in 3.2.2, but not 3.2.1.


golddotasksquestions:
I'd love to test it. How can I download these three files?

aaronfranke:
The ideal way is to use Git. You can either use command line Git ([Windows download link](https://git-scm.com/)) or a GUI client ([my GitKraken video](https://www.youtube.com/watch?v=0XKy8YpP5Ww)).

Or, for the lazy way: you can open the branch, where it says `aaronfranke:3din2d`, and then click Code -> Download ZIP.

golddotasksquestions:
Thank you!

I had a closer look at it, a couple of things: 

- I still think this demo needs no scripts at all. The code for viewport stretching is interesting, but I think this is not the place. Viewport stretching is not a requirement to have 3D in 2D. Besides, it seems to work just fine without.
- Imho the naming of the nodes should have their default names, in order to confuse beginners as little as possible. The AnimatedSprite node therefor just be names as "AnimatedSprite", same goes for the Sprite node.
- The Camera2D node is completely unnecessary.
- The Sprite node should not be rotated 180 degrees, but instead scaled -1 in order to be flipped. Note this is unnecessary with ViewportContainers.
- I also feel like This demo needs a scene showcasing the use of ViewportContainer+Viewport solution. ViewportTextures are also viable, so it makes a lot of sense to include two scenes showing both solutions.

I feel like the result is a lot more clear and easy to understand, without the need of comments or explanation.
Please let me know what you think:
[3d_in_2d_gold.zip](https://github.com/godotengine/godot-demo-projects/files/5649937/3d_in_2d_gold.zip)

Calinou:
> The Camera2D node is completely unnecessary.

The Camera2D node is required to properly support multiple aspect ratios in the demo, as explained in https://github.com/godotengine/godot-demo-projects/pull/341.

aaronfranke:
@golddotasksquestions Good points about the name of AnimatedSprite and setting the scale instead of the rotation, but please see @Calinou's comment as to why we need the Camera2D and why the script exists. If you want to suggest improvements that don't have drawbacks, that would still be appreciated.

golddotasksquestions:
> The Camera2D node is required to properly support multiple aspect ratios in the demo, as explained in #341. We want the demo projects to showcase good practices and prove that Godot is able of supporting any resolution and aspect ratio slightly_smiling_face
> 
> Also, make sure the demo still handles multiple resolutions correctly with this PR. The sprite should not become blurry or pixelated when you resize the window – that's what the root Viewport size_changed method connection is currently used for.

@aaronfranke  @Calinou 
Totally agree on keeping demos consistent with best practices. However this is not a demo about multiple resolutions or how to unblur pixelart. In my opinion these should be separate demos.
As I beginner I open these demos and assume they only contain the necessary things for what they are supposed to showcase. As I beginner I would not be able to differentiate what is multiple resolution code, what is not. I would think this code is all necessary to bring 3D into 2D. It would seem more dawnting for me than it has to me. It would be more confusing for me than it has to be, as I might not be able to differentiate the multiple resolution code yet from the purpose of bringing 3D into 2D.

If you guys want this in there, you should rename it to: "3D in 2D with multiple resolution and HDPI support demo"
Still I don't think this would be as useful for beginners as having those in separate projects.

If your point is that this Camera and viewport stretch code needs to be in every demo, because we need to communicate this code needs to be in every project to support as many aspect ratios as possible, than I would say it belongs in the project settings, not in every demo.

aaronfranke:
> As I beginner I would not be able to differentiate what is multiple resolution code, what is not.

There is this comment above the method that gets connected to the signal. The only other method in the script now is the `_ready()` function, which plays the AnimatedSprite, connects the signal to the method, and sets a value for the method to use.

```gdscript
# Called when the root's viewport size changes (i.e. when the window is resized).
# This is done to handle multiple resolutions without losing quality.
```

Also, the `robot_3d.gd` script is very simple and contains a comment, so I think that one is also fine.

> If your point is that this Camera and viewport stretch code needs to be in every demo

The Viewport stretch code is specific to this demo because we have this child Viewport that we need to resize. It's not needed in other demos because most of the demos don't have a child Viewport to resize.

golddotasksquestions:
I feel like I'm failing to convey my point. This demo is not a game (in contrast to the platformer bomber of pong demos). It showcases one thing. That one thing is in the name of the demo. If there is other code doing other things I would assume there is connection to what this demo is supposed to showcase. But there is no connection. This Camera and code has nothing to do with having 3D in 2D.

aaronfranke:
> I don't have to understand GDscript or know any programming at all to get what is going on here.

Viewports are not really a beginner topic. I think it's perfectly valid to require people to know GDScript before trying this demo.

> What about making a third scene then that has the viewport resolution update?

Sure, but it doesn't need to be three scenes. It can just be two, or you can just have both in one scene.

PR #558, opened 2020-12-09
Update projects to Godot 3.3

aaronfranke:
This PR updates the demos to Godot 3.2.4 by opening and saving all scenes in Godot 3.3 and fixing warnings/errors. I also made a few tweaks in a few places.


corjohnson:
Would it make sense to remove an unused var in https://github.com/godotengine/godot-demo-projects/blob/master/3d/kinematic_character/cubio.gd ? I don't see MAX_SLOPE_ANGLE used anywhere. Not sure if that's the kind of clean up you're wanting to do on this PR though!

aaronfranke:
@corjohnson Thanks for pointing that out, I fixed it in #590.

PR #559, opened 2020-12-12
Fixes and adjustments in 3D physics tests

pouleyKetchoupp:
* Add Functional Test / Stack & Pyramid
For testing stack stability.

* Add Functional Test / Raycasts
Visually test raycast on different shapes.

* Fix leaks on exit
Some Nodes are copied and removed from the scene to be used as templates,
they need to be freed manually on exit.

* Fix Performance Test / Contacts
Positions adjusted, some shape types were not created at the center.

### DIFF
--- a/3d/physics_tests/tests/functional/test_pyramid.gd
+++ b/3d/physics_tests/tests/functional/test_pyramid.gd
@@ -1,0 +1,58 @@
+extends Test
+
+
+export(int, 1, 100) var height = 10
+export(int, 1, 100) var width_max = 100
+export(int, 1, 100) var depth_max = 1
+export(Vector3) var box_size = Vector3(1.0, 1.0, 1.0)
+export(Vector3) var box_spacing = Vector3(0.0, 0.0, 0.0)
+
+
+func _ready():
+	_create_pyramid()
+
+
+func _create_pyramid():
+	var root_node = $Pyramid
+
+	var template_shape = BoxShape.new()
+	template_shape.extents = 0.5 * box_size
+
+	var template_collision = CollisionShape.new()
+	template_collision.shape = template_shape
+
+	var template_body = RigidBody.new()
+	template_body.add_child(template_collision)
+
+	var pos_y = 0.5 * box_size.y + box_spacing.y
+
+	for level in height:
+		var level_index = height - level - 1
+		var num_boxes = 2 * level_index + 1
+		var num_boxes_width = min(num_boxes, width_max)
+		var num_boxes_depth = min(num_boxes, depth_max)
+
+		var row_node = Spatial.new()
+		row_node.transform.origin = Vector3(0.0, pos_y, 0.0)
+		row_node.name = "Row%02d" % (level + 1)
+		root_node.add_child(row_node)
+
+		var pos_x = -0.5 * (num_boxes_width - 1) * (box_size.x + box_spacing.x)
+
+		for box_index_x in num_boxes_width:
+			var pos_z = -0.5 * (num_boxes_depth - 1) * (box_size.z + box_spacing.z)
+
+			for box_index_z in num_boxes_depth:
+				var box_index = box_index_x * box_index_z
+				var box = template_body.duplicate()
+				box.transform.origin = Vector3(pos_x, 0.0, pos_z)
+				box.name = "Box%02d" % (box_index + 1)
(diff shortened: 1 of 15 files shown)


pouleyKetchoupp:
Added new test (Performance / Broadphase).

3d/physics_tests/tests/functional/test_raycasting.gd
```diff
@@ -0,0 +1,78 @@
...
+
+
+func _add_raycast(space_state, pos_start, pos_end):
+	var result = space_state.intersect_ray(pos_start, pos_end)
+	if result:
+		_raycast_visuals.set_color(Color.green)
+	else:
+		_raycast_visuals.set_color(Color.red.darkened(0.5))
+
+	# draw raycast line.
```
aaronfranke, line 68:
```suggestion
	# Draw raycast line.
```

3d/physics_tests/tests/functional/test_raycasting.gd
```diff
@@ -0,0 +1,78 @@
...
+	_raycast_visuals.add_vertex(pos_start)
+	_raycast_visuals.add_vertex(pos_end)
+
+	# draw raycast arrow.
```
aaronfranke, line 72:
```suggestion
	# Draw raycast arrow.
```

3d/physics_tests/tests/functional/test_pyramid.gd
```diff
@@ -0,0 +1,57 @@
...
+				var box = template_body.duplicate()
+				box.transform.origin = Vector3(pos_x, 0.0, pos_z)
+				box.name = "Box%02d" % (box_index + 1)
+				row_node.add_child(box)
+
+				pos_z += box_size.z
+
+			pos_x += box_size.x
+
+		pos_y += box_size.y
```
aaronfranke, line 55: The initial conditions of this test (and the box stack test) are "too perfect" to thoroughly test physics. It would be much better if the starting position wasn't already perfectly stacked. For example, changing this line to `pos_y += box_size.y * 2` makes the boxes start in the air and fall down, and *immediately* showcases a problem when run with Godot Physics.

3d/physics_tests/tests/functional/test_pyramid.gd
```diff
@@ -0,0 +1,57 @@
...
+export(int, 1, 100) var depth_max = 1
+export(Vector3) var box_size = Vector3(1.0, 1.0, 1.0)
+
+
+func _ready():
+	_create_pyramid()
+
+
+func _create_pyramid():
+	var root_node = $Pyramid
```
aaronfranke, line 15: A pyramid has a square base. What you have created here is a triangle :P

Not sure if we want to have a triangle test, a pyramid test, or both, but as-is this isn't good because the name is incorrect.

aaronfranke (changes requested):

3d/physics_tests/tests/functional/test_pyramid.gd
pouleyKetchoupp: The main goal here is to test physics stability for stacking, without necessarily adding extra forces to the equation. You can set a higher pyramid and you'll showcase a problem already.

I'll add a separation between boxes as an option though, it's a good idea to allow the possibility to test more cases.

3d/physics_tests/tests/functional/test_pyramid.gd
pouleyKetchoupp: You can see it as a 2D pyramid too :)

Actually, this scene does both. If you set a max depth of 100 you'll get an actual 3D pyramid. I haven't made that the default because it was running extremely slowly with too many rigid bodies.

pouleyKetchoupp:
Thanks for your tests and feedback!

In general, I'm planning to add box fields when you run the tests so you can change parameters easily. Maybe that will make all the different options and use cases clearer.

pouleyKetchoupp:
@aaronfranke Just pushed changes from your review.

aaronfranke:
Is there a reason that "Broadphase" is not in the menu?

pouleyKetchoupp:
> Is there a reason that "Broadphase" is not in the menu?

No, I forgot to add it. Thanks for catching this! It's fixed now.

aaronfranke (approved):

PR #560, opened 2020-12-12
Add 2D Physics Tests

pouleyKetchoupp:
Similar to its 3D counterpart, with tests using 2D physics.

screenshot


pouleyKetchoupp:
Added new test (Performance / Broadphase).

2d/physics_tests/project.godot
```diff
@@ -0,0 +1,75 @@
...
+restart_test={
+"deadzone": 0.5,
+"events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":82,"unicode":0,"echo":false,"script":null)
+ ]
+}
+
+[rendering]
+
+quality/driver/driver_name="GLES2"
+environment/default_clear_color=Color( 0, 0, 0, 1 )
```
aaronfranke, line 74: I think the clear color should be consistent between the 2D and 3D physics tests.

2d/physics_tests/tests/functional/test_raycasting.gd
```diff
@@ -0,0 +1,69 @@
...
+
+func _add_raycast(space_state, pos_start, pos_end):
+	var result = space_state.intersect_ray(pos_start, pos_end)
+	var color
+	if result:
+		color = Color.gray
+	else:
+		color = Color.gray.darkened(0.6)
+
+	# draw raycast line.
```
aaronfranke, line 62:
```suggestion
	# Draw raycast line.
```

2d/physics_tests/tests/functional/test_raycasting.gd
```diff
@@ -0,0 +1,69 @@
...
+	add_line(pos_start, pos_end, color)
+
+	# draw raycast arrow.
```
aaronfranke, line 65:
```suggestion
	# Draw raycast arrow.
```

2d/physics_tests/tests/functional/test_raycasting.gd
```diff
@@ -0,0 +1,69 @@
...
+			Log.print_log("Raycast inside face: %s" % ("HIT" if res else "NO HIT"))
+
+
+func _add_raycast(space_state, pos_start, pos_end):
+	var result = space_state.intersect_ray(pos_start, pos_end)
+	var color
+	if result:
+		color = Color.gray
+	else:
+		color = Color.gray.darkened(0.6)
```
aaronfranke, line 60: I like how the arrows are colored red and green in the 3D demo, I think it would look good here too. Either way, the colors should probably be consistent between 2D and 3D physics tests.

aaronfranke (changes requested):
You accidentally left an empty `3d/physics_tests/tests/functional/test_raycasting.gd` file in this PR, which doesn't belong here (it's for 3D).

2d/physics_tests/tests/functional/test_raycasting.gd
pouleyKetchoupp: I've tried to pick colors that contrast well with the physics render debug colors and they are not the same for 3D and 2D. I'll see how much I can harmonize it while keeping it readable for both.

2d/physics_tests/tests/test_options.tscn
```diff
@@ -0,0 +1,16 @@
+[gd_scene load_steps=2 format=2]
```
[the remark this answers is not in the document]
pouleyKetchoupp: It's a little helper to avoid setting the correct margins, position and other parameters manually when I need this control.

What is the reason behind not allowing a scene with just one node? I personally use that all the time because it's much more practical and flexible than copying the same node between scenes, and I can't see a reason why we should tell users not to do it if it can save production time.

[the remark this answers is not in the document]
aaronfranke: Well, there's also the fact that this scene is only instanced in one place... so "when I need this control" is just once, in `test_perf_contacts.tscn`. Should we make every node its own scene? :upside_down_face: 

> What is the reason behind not allowing a scene with just one node?

To avoid making the demos more complicated than they need to be.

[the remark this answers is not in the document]
pouleyKetchoupp: We're only talking about a specific case, there's no need to be snarky here.

I'm going to use this node again for more scenes, so in this case I need to have it saved as a separate scene.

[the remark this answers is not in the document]
aaronfranke: Yeah. If there is a specific reason for this, then it's fine.

2d/physics_tests/tests/functional/test_raycasting.gd
pouleyKetchoupp: In the end I've used the same colors as in 3D. It's not the best contrast in some cases (convex polygon debug draw is red) but it's still readable.

pouleyKetchoupp:
@aaronfranke Just pushed changes from your review.

pouleyKetchoupp:
I've added an option for box spacing in pyramid and stack tests from https://github.com/godotengine/godot-demo-projects/pull/559#discussion_r541590559 to make 3D and 2D consistent.

aaronfranke:
We should compress the screenshot using `oxipng`. I tried to fix this for you, but I don't have permission to push to `nekomatata/godot-demo-projects`. Here is the optimized file:

screenshot

pouleyKetchoupp:
Thanks! Image replaced.

aaronfranke (approved):

PR #561, opened 2020-12-18
Add collision pairs test to 2D/3D physics tests

pouleyKetchoupp:
Functional test used for checking/debugging different collision cases for all possible pairs of shape types.

Minor changes in a few other tests to keep things consistent.

Note: In 3D, Bullet and GodotPhysics currently don't agree on whether contact points returned by `PhysicsDirectSpaceState::collide_shape` should be in local space or world space. I'm assuming world space to be consistent with 2D, but with Bullet they are not drawn in the correct location because of this discrepancy.

### DIFF
--- a/2d/physics_tests/test.gd
+++ b/2d/physics_tests/test.gd
@@ -9,13 +9,16 @@ var _timer_started = false
 
 var _wait_physics_ticks_counter = 0
 
-
-class Line:
-	var pos_start
-	var pos_end
+class Circle2D:
+	extends Node2D
+	var center
+	var radius
 	var color
 
-var _lines = []
+	func _draw():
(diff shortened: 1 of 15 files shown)


aaronfranke (approved):
This looks great! It works great and I don't see anything wrong.

It also occurred to me, would you like to add yourself to the `.github/CODEOWNERS` file for each of the physics demos? That way you can be pinged if anyone submits a PR to these demos.

pouleyKetchoupp:
> It also occurred to me, would you like to add yourself to the `.github/CODEOWNERS` file for each of the physics demos? That way you can be pinged if anyone submits a PR to these demos.

Good idea!

PR #548, opened 2020-11-04
Demos UI Updates

Mr-K-GameDev:
# Updating UI Projects

The goal by doing this, Is trying to change the Look and qulity of UI Demos (And Any other Demo that have UI in it)
And make them at a Higher standard.

## 2D Platformer Demo -*Closed*-

image

### Progress Update
Fixed The bugs That This Commit had, Thanks to @aaronfranke for noticing them, and Giving me really Good Feedback on it.

## Drag and Drop Demo

image

## Tween Demo UI Update

image

PR #570, opened 2021-01-03
Update the Drag and Drop demo and the Tween demo

aaronfranke:
This PR was inspired by #548, which raises some very good points, even if I don't think it's good as a whole. ~~This doesn't close 548 just yet, I haven't looked at the Tween script changes, a lot of which is probably useful.~~ EDIT: See 3rd commit.

Drag and Drop:

* Make all `ColorPickerButton` nodes be children of a `GridContainer`.

Tween:

* Rework the UI layout to use Container nodes.
* Add a font for compatibility with high-DPI displays.
* The window now has a design resolution of 800 pixels in height to accomodate the layout rework and slightly bigger font.
* Replace the icon and screenshot with updated ones.

Icon before/after:

icon -> icon

Screenshot before (note: the demo did not look exactly like this before, but still fairly similar, the background color is different):

tween

Screenshot after:

tween


Mr-K-GameDev:
The Tween Demo changes you've made look pritty nice, but i don't see the point of this PR
Mainly because all of the changes already exist in #548 PR and quite a few are Missing in this PR
- Cleaning-up Scene's Nodes 
- Orginazing Project File System
- Adding Theming Support For both Tween Demo and Drag and Drop Demo

aaronfranke:
* What do you mean by cleaning up the nodes?
* This PR is a lot simpler, so there are fewer files to organize, and there is no need for folders.
* This was intentional, I don't think that the theme is an improvement. Godot's default theme works fine.

Mr-K-GameDev:
Yes, I see what you're trying to do, but this doesn't fit with ower main issue ( Issue #390 ) which is making Demos more polished
and with a clear Design, and to do so we Need to use themes in UI focused Demos (Or any Demo that have UI in it)
Also the Scene's Nodes look very un-organized at a first look, So it's better to make them into scenes, And making Demos Simplier is a good thing we should do, But for big-ish Projects/Demos, It's better to try and organize the File System
and do some best practices, So Demos be more polished And good Learning Marerial for people to use.

NathanLovato:
> to do so we Need to use themes in UI focused Demos

That's not what I meant in #390. A consistent visual language is not so much about using a theme resource for UI as it is about writing a style guide and loosely following it. You can strictly share resources and apply the same theme resource everywhere, it may look inconsistent and off in some demos. It depends on the context.

To me, it's fine if all tool-like UI in the demos uses default UI styles, like in this example. It looks fine, it's clear where buttons are, what you can click, and it's visually clear the demo's a tool to test engine features.

<hr>

When it comes to code style and structure, I think that:

1. Demos like this one aren't meant for you to study their source code. They're more on the showcase side: it's a tool to test features interactively; not an example of how to write code in your game.
2. Extending the previous point, we have limited time and resources, and there are other demos that could benefit more from showcasing better programming practices maybe. Demos that focus on gameplay or various game mechanics, for example.

Mr-K-GameDev:
Now i want to Improve this Repo, Because i really like the idea of Showcasing Godot Features and ideas in simple Project
(Most of the time), And i don't think Theming every project would result in a consistent look, As @NathanLovato Mentioned before
So i think it's better to Improve this PR instead, And thanks for your honest reply/feedback Mr.Nathan.

---
So far it's really good Well done @aaronfranke , I recommend putting a Random int genrator for Drage and Drop Demo
If you want to here's my code, So the person wanting to see this feature, Can Drage and Drop right away!
```python
func _ready():
   var c = []
   randomize()
   for i in range(3):
      c.append(randi(0, 1))
   set_pick_color(c[0], c[1], c[2])
```

PR #573, opened 2021-01-04
Add One Way Collision for 2D Physics tests

pouleyKetchoupp:
Update 2D Physics tests with a new demo to test one way collision.

This test scene is inspired by the "Around the Clock" test made by @madmiraal in https://github.com/madmiraal/godot-gym in order to test this PR: https://github.com/godotengine/godot/pull/42574

Like the original test, it allows testing collision between a rigid/kinematic body object and a one-way collision platform at different angles.

This new version is a bit different to be able to use it in a more practical way within the physics test framework:
- There are options and sliders to set different test cases (kinematic or rigid body, platform size, etc.)
- You can still test all angles around the clock, but you can also test only one particular case for debug purpose
- When all angles have been tested, the log gives you information about which sets of parameters have failed
- Only the platform rotates for "around the clock" test and the rigid body goes horizontally to avoid redundant cases

Overall, the idea is to be able to test a set of parameters in less than a minute, check the log for failed cases and test these failed cases easily.

@madmiraal I think it covers the same cases as your original test, but if you have some time to have a look, please let me know in case I overlooked something.

### DIFF
--- a/2d/physics_tests/tests/functional/test_one_way_collision.gd
+++ b/2d/physics_tests/tests/functional/test_one_way_collision.gd
@@ -1,0 +1,277 @@
+extends Test
+tool
+
+const OPTION_OBJECT_TYPE_RIGIDBODY = "Object type/Rigid body (1)"
+const OPTION_OBJECT_TYPE_KINEMATIC = "Object type/Kinematic body (2)"
+
+const OPTION_TEST_CASE_ALL_ANGLES = "Test case/Around the clock (0)"
+
+const TEST_ALL_ANGLES_STEP = 15.0
+const TEST_ALL_ANGLES_MAX = 344.0
+
+export(float, 32, 128, 0.1) var _platform_size = 64.0 setget _set_platform_size
+export(float, 0, 360, 0.1) var _platform_angle = 0.0 setget _set_platform_angle
+export(float, 0, 360, 0.1) var _body_angle = 0.0 setget _set_rigidbody_angle
+export(Vector2) var _body_velocity = Vector2(400.0, 0.0)
+export(bool) var _use_kinematic_body = false
+
+var _rigid_body_template = null
+var _kinematic_body_template = null
+var _moving_body = null
+
+var _contact_detected = false
+var _target_entered = false
+var _test_passed = false
+var _test_step = 0
+
+var _test_all_angles = false
+var _lock_controls = false
+
+
+func _ready():
+	if not Engine.editor_hint:
+		$Options.add_menu_item(OPTION_OBJECT_TYPE_RIGIDBODY, true, not _use_kinematic_body, true)
+		$Options.add_menu_item(OPTION_OBJECT_TYPE_KINEMATIC, true, _use_kinematic_body, true)
+
+		$Options.add_menu_item(OPTION_TEST_CASE_ALL_ANGLES)
+
+		$Options.connect("option_selected", self, "_on_option_selected")
+
+		$Controls/PlatformSize/HSlider.value = _platform_size
+		$Controls/PlatformAngle/HSlider.value = _platform_angle
+		$Controls/BodyAngle/HSlider.value = _body_angle
+
+		$TargetArea2D.connect("body_entered", self, "_on_target_entered")
+		$Timer.connect("timeout", self, "_on_timeout")
+
+		_rigid_body_template = $RigidBody2D
+		remove_child(_rigid_body_template)
+
+		_kinematic_body_template = $KinematicBody2D
+		remove_child(_kinematic_body_template)
+
+		_start_test()
+
+
+func _process(_delta):
+	if not Engine.editor_hint:
+		if Input.is_action_just_pressed("ui_accept"):
+			_reset_test(false)
+
+
+func _physics_process(_delta):
+	if not Engine.editor_hint:
+		if _moving_body and _use_kinematic_body:
+			_moving_body.move_and_slide(_body_velocity)
+			if _moving_body.get_slide_count() > 0:
+				var colliding_body = _moving_body.get_slide_collision(0).collider
+				_on_contact_detected(colliding_body)
+
+
+func _input(event):
+	var key_event = event as InputEventKey
+	if key_event and not key_event.pressed:
+		if key_event.scancode == KEY_0:
+			_on_option_selected(OPTION_TEST_CASE_ALL_ANGLES)
+		if key_event.scancode == KEY_1:
+			_on_option_selected(OPTION_OBJECT_TYPE_RIGIDBODY)
+		elif key_event.scancode == KEY_2:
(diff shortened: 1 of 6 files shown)


2d/physics_tests/tests/functional/test_one_way_collision.gd
```diff
@@ -0,0 +1,238 @@
...
+const OPTION_OBJECT_TYPE_KINEMATIC = "Object type/Kinematic body"
+
+const OPTION_TEST_CASE_ALL_ANGLES = "Test case/Around the clock"
+
+const TEST_ALL_ANGLES_STEP = 15.0
+const TEST_ALL_ANGLES_MAX = 344.0
+
+export(float, 32, 128, 0.1) var _platform_size = 64 setget _set_platform_size
+export(float, 0, 360, 0.1) var _platform_angle = 0 setget _set_platform_angle
+export(float, 0, 360, 0.1) var _rigidbody_angle = 0 setget _set_rigidbody_angle
```
aaronfranke, line 14: These are exported as floats, but the inspector infers their type as int unless you add a trailing `.0`.

```suggestion
export(float, 32, 128, 0.1) var _platform_size = 64.0 setget _set_platform_size
export(float, 0, 360, 0.1) var _platform_angle = 0.0 setget _set_platform_angle
export(float, 0, 360, 0.1) var _rigidbody_angle = 0.0 setget _set_rigidbody_angle
```

2d/physics_tests/tests/functional/test_one_way_collision.gd
```diff
@@ -0,0 +1,238 @@
...
+	_on_timeout()
+
+
+func _should_collide():
+	var platform_rotation = round(rad2deg($OneWayRigidBody2D.rotation))
+
+	var angle = platform_rotation + 360
+	while angle >= 360:
+		angle -= 360
+
```
aaronfranke, line 208: We have a method for this :)

```suggestion
	var angle = fposmod(platform_rotation, 360)
```

2d/physics_tests/utils/slider.gd
```diff
@@ -0,0 +1,12 @@
...
+
+export(float) var snap_step = 1.0
+
+
+func _process(_delta):
+	if Input.is_key_pressed(KEY_SHIFT):
+		if has_focus():
+			step = snap_step
+	else:
+		step = 0.1
```
aaronfranke, line 12: Maybe it should be the other way around, to make the special cases like 180 degrees easily accessible? Also. `has_focus()` doesn't seem to be necessary to get the correct behavior.

```suggestion
	if Input.is_key_pressed(KEY_SHIFT):
		step = 0.1
	else:
		step = snap_step
```

2d/physics_tests/tests/functional/test_one_way_collision.tscn
```diff
@@ -0,0 +1,225 @@
...
+text = "Platform size"
+
+[node name="HSlider" type="HSlider" parent="Controls/PlatformSize"]
+margin_left = 120.0
+margin_right = 320.0
+margin_bottom = 16.0
+rect_min_size = Vector2( 200, 0 )
+min_value = 32.0
+max_value = 128.0
+step = 0.1
```
aaronfranke, line 58:
```suggestion
```

2d/physics_tests/tests/functional/test_one_way_collision.tscn
```diff
@@ -0,0 +1,225 @@
...
+margin_bottom = 15.0
+text = "Platform angle"
+
+[node name="HSlider" type="HSlider" parent="Controls/PlatformAngle"]
+margin_left = 120.0
+margin_right = 320.0
+margin_bottom = 16.0
+rect_min_size = Vector2( 200, 0 )
+max_value = 360.0
+step = 0.1
```
aaronfranke, line 94:
```suggestion
```

2d/physics_tests/tests/functional/test_one_way_collision.tscn
```diff
@@ -0,0 +1,225 @@
...
+margin_bottom = 15.0
+text = "Rigidbody angle"
+
+[node name="HSlider" type="HSlider" parent="Controls/RigidbodyAngle"]
+margin_left = 120.0
+margin_right = 320.0
+margin_bottom = 16.0
+rect_min_size = Vector2( 200, 0 )
+max_value = 360.0
+step = 0.1
```
aaronfranke, line 129:
```suggestion
```

2d/physics_tests/tests/functional/test_one_way_collision.gd
```diff
@@ -0,0 +1,238 @@
...
+			_start_test()
+	elif force_start:
+		_start_test()
+
+
+func _on_contact_detected(_body):
+	if _contact_detected or _target_entered:
+		return
+
+	if _rigidbody && _use_kinematic_body:
```
aaronfranke, line 183: Old habits die hard :)

```suggestion
	if _rigidbody and _use_kinematic_body:
```

2d/physics_tests/tests/functional/test_one_way_collision.gd
```diff
@@ -0,0 +1,238 @@
...
+	_next_test(true)
+
+
+func _next_test(force_start = false):
+	if _rigidbody:
+		remove_child(_rigidbody)
+		_rigidbody.queue_free()
+		_rigidbody = null
+
+	if _test_all_angles:
```
aaronfranke, line 166: It would be nice to update the slider during the test, but I tried doing this myself and it just caused the test to terminate prematurely.

aaronfranke (changes requested):

pouleyKetchoupp:
Thanks @aaronfranke! I've updated the test from all your suggestions.

pouleyKetchoupp:
I've just added proper support for `KinematicBody2D`, and key shortcuts.

2d/physics_tests/tests/functional/test_one_way_collision.tscn
```diff
@@ -0,0 +1,244 @@
...
+
+[sub_resource type="RectangleShape2D" id=2]
+extents = Vector2( 64, 32 )
+
+[sub_resource type="RectangleShape2D" id=3]
+extents = Vector2( 32, 32 )
+
+[node name="Test" type="Node2D"]
+script = ExtResource( 1 )
+_use_kinematic_body = true
```
aaronfranke, line 20: Is this intentional? It's fine either way, but it seems strange that the second option is the default one.

pouleyKetchoupp: Ah no it's not, I forgot to revert the change before pushing :)

aaronfranke (approved):

PR #575, opened 2021-01-18
[Joypads] Add SDL config re-mapping tool.

Faless:
Most of the code is in the remap folder, but it depends on the gamepad
diagram scene.
It allows remapping of pads to values that godot can understand.
It also comes with some default mapping for the HTML5 platform.

See godotengine/godot#45078

remap

### DIFF
--- a/misc/joypads/remap/joy_mapping.gd
+++ b/misc/joypads/remap/joy_mapping.gd
@@ -1,0 +1,137 @@
+extends Reference
+class_name JoyMapping
+
+
+enum TYPE {NONE, BTN, AXIS}
+enum AXIS {FULL, HALF_PLUS, HALF_MINUS}
+
+const PLATFORMS = {
+	# From gamecontrollerdb
+	"Windows": "Windows",
+	"OSX": "Mac OS X",
+	"X11": "Linux",
+	"Android": "Android",
+	"iOS": "iOS",
+	# Godot customs
+	"HTML5": "Javascript",
+	"UWP": "UWP",
+	# 4.x compat
+	"Linux": "Linux",
+	"FreeBSD": "Linux",
+	"NetBSD": "Linux",
+	"BSD": "Linux",
+	"macOS": "Mac OS X",
+}
+
+const BASE = {
+	# Buttons
+	"a": JOY_XBOX_A,
+	"b": JOY_XBOX_B,
+	"y": JOY_XBOX_Y,
+	"x": JOY_XBOX_X,
+	"start": JOY_START,
+	"back": JOY_SELECT,
+	"leftstick": JOY_BUTTON_8,
+	"rightstick": JOY_BUTTON_9,
+	"leftshoulder": JOY_L,
+	"rightshoulder": JOY_R,
+	"dpup": JOY_DPAD_UP,
+	"dpleft": JOY_DPAD_LEFT,
+	"dpdown": JOY_DPAD_DOWN,
+	"dpright": JOY_DPAD_RIGHT,
+
+	# Axis
+	"leftx": JOY_AXIS_0,
+	"lefty": JOY_AXIS_1,
+	"rightx": JOY_AXIS_2,
+	"righty": JOY_AXIS_3,
+	"lefttrigger": JOY_ANALOG_L2,
+	"righttrigger": JOY_ANALOG_R2,
+}
+
+const XBOX = {
+	"a": "b0",
+	"b": "b1",
+	"y": "b3",
+	"x": "b2",
+	"start": "b7",
+	"guide": "b8",
+	"back": "b6",
+	"leftstick": "b9",
+	"rightstick": "b10",
+	"leftshoulder": "b4",
+	"rightshoulder": "b5",
+	"dpup": "-a7",
+	"dpleft":"-a6",
+	"dpdown": "+a7",
+	"dpright": "+a6",
+	"leftx": "a0",
+	"lefty": "a1",
+	"rightx": "a3",
+	"righty": "a4",
+	"lefttrigger": "a2",
+	"righttrigger": "a5",
+}
+
+const XBOX_OSX = {
+	"a": "b11",
+	"b": "b12",
+	"y": "b14",
+	"x": "b13",
+	"start": "b4",
+	"back": "b5",
+	"leftstick": "b6",
+	"rightstick": "b7",
+	"leftshoulder": "b8",
+	"rightshoulder": "b9",
+	"dpup": "b0",
+	"dpleft": "b2",
+	"dpdown": "b1",
+	"dpright": "b3",
+	"leftx": "a0",
+	"lefty": "a1",
+	"rightx": "a2",
+	"righty": "a3",
+	"lefttrigger": "a4",
+	"righttrigger":"a5",
+}
+
+var type = TYPE.NONE
+var idx = -1
+var axis = AXIS.FULL
+var inverted = false
+
+
+func _init(p_type = TYPE.NONE, p_idx = -1, p_axis = AXIS.FULL):
+	type = p_type
+	idx = p_idx
+	axis = p_axis
+
+
+func _to_string():
+	if type == TYPE.NONE:
+		return ""
+	var ts = "b" if type == TYPE.BTN else "a"
+	var prefix = ""
+	var suffix = "~" if inverted else ""
+	match axis:
+		AXIS.HALF_PLUS:
+			prefix = "+"
+		AXIS.HALF_MINUS:
+			prefix = "-"
+	return "%s%s%d%s" % [prefix, ts, idx, suffix]
+
+
+func to_human_string():
+	if type == TYPE.BTN:
+		return "Button %d" % idx
+	if type == TYPE.AXIS:
+		var prefix = ""
+		match axis:
+			AXIS.HALF_PLUS:
+				prefix = "(+) "
+			AXIS.HALF_MINUS:
+				prefix = "(-) "
+		var suffix = " (inverted)" if inverted else ""
+		return "Axis %s%d%s" % [prefix, idx, suffix]
+	return ""
--- a/misc/joypads/remap/remap_wizard.gd
+++ b/misc/joypads/remap/remap_wizard.gd
@@ -1,0 +1,189 @@
+extends Node
+
+
+const DEADZONE = 0.3
+
+var joy_guid = ""
+var joy_name = ""
+
+var steps = JoyMapping.BASE.keys()
+var cur_step = -1
+var cur_mapping = {}
+var last_mapping = ""
+
+onready var joy_buttons = $Mapping/Margin/VBox/ViewportContainer/Viewport/JoypadDiagram/Buttons
+onready var joy_axes = $Mapping/Margin/VBox/ViewportContainer/Viewport/JoypadDiagram/Axes
+onready var joy_mapping_text = $Mapping/Margin/VBox/Info/Text/Value
+onready var joy_mapping_full_axis = $Mapping/Margin/VBox/Info/Extra/FullAxis
+onready var joy_mapping_axis_invert = $Mapping/Margin/VBox/Info/Extra/InvertAxis
+
+
+func _input(event):
+	if cur_step == -1:
+		return
+	if event is InputEventJoypadMotion:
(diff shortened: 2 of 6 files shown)


misc/joypads/remap/remap_wizard.gd
```diff
@@ -0,0 +1,166 @@
...
+	step_next()
+
+func _input(event):
+	if cur_step == -1:
+		return
+	if event is InputEventJoypadMotion:
+		var motion = event as InputEventJoypadMotion
+		if abs(motion.axis_value) > DEADZONE:
+			var idx = motion.axis
+			var key = steps[cur_step]
```
aaronfranke, line 107:
```
W 0:00:01.183   The local variable 'key' is declared but never used in the block. If this is intended, prefix it with an underscore: '_key'
  <C++ Error>   UNUSED_VARIABLE
  <Source>      remap_wizard.gd:107
```

misc/joypads/remap/joy_mapping.gd
```diff
@@ -0,0 +1,115 @@
...
+	"lefttrigger": "a4",
+	"righttrigger":"a5",
+}
+
+var type = TYPE.NONE
+var idx = -1
+var axis = AXIS.FULL
+var inverted = false
+
+func _init(type = TYPE.NONE, idx = -1, axis = AXIS.FULL):
```
aaronfranke, line 85:
```
W 0:00:01.306   The local variable 'type' is shadowing an already-defined variable at line 80.
  <C++ Error>   SHADOWED_VARIABLE
  <Source>      joy_mapping.gd:85
W 0:00:01.306   The local variable 'idx' is shadowing an already-defined variable at line 81.
  <C++ Error>   SHADOWED_VARIABLE
  <Source>      joy_mapping.gd:85
W 0:00:01.306   The local variable 'axis' is shadowing an already-defined variable at line 82.
  <C++ Error>   SHADOWED_VARIABLE
  <Source>      joy_mapping.gd:85
```

In this case, we can just do this:

```suggestion
# warning-ignore-all:shadowed_variable
func _init(type = TYPE.NONE, idx = -1, axis = AXIS.FULL):
```

misc/joypads/remap/joy_mapping.gd
```diff
@@ -0,0 +1,115 @@
...
+	self.type = type
+	self.idx = idx
+	self.axis = axis
+
+func _to_string():
```
aaronfranke, line 90: Per the [GDScript style guide](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html), there needs to be two lines between methods.

```suggestion


func _to_string():
```

misc/joypads/remap/remap_wizard.gd
```diff
@@ -0,0 +1,166 @@
...
+	return string + "platform:Javascript"
+
+func start(idx):
+	joy_guid = Input.get_joy_guid(idx)
+	joy_name = Input.get_joy_name(idx)
+	if joy_guid.empty():
+		printerr("Unable to find controller")
+		return
+	if OS.get_name() == "HTML5":
+		# Propose trying known mapping on HTML5
```
aaronfranke, line 35:
```suggestion
		# Propose trying known mapping on HTML5.
```

aaronfranke (changes requested):
Please ensure that PRs don't have warnings and follow all style guides.

aaronfranke (changes requested):
It doesn't work correctly in my testing, tested with both 3.2.3 and 3.2.4 beta 6. I tried remapping a Stadia controller on Windows. Firstly, I can't map the A button unless I click on a UI element to change the focused button, because the A button immediately closes the wizard (this needs to be fixed, you can probably just focus some other UI element to prevent this). Anyway, I clicked elsewhere and mapped it and this is what I got on 3.2.3:

    03000000d11800000094000000000000,Stadia Controller,a:b0,b:b1,y:b3,x:b2,start:b9,back:b8,leftstick:b6,rightstick:b7,leftshoulder:b4,rightshoulder:b5,dpup:b12,dpleft:b14,dpdown:b13,dpright:b15,leftx:a0,lefty:a2,rightx:a3,righty:a5,lefttrigger:b12,righttrigger:b11,platform:Javascript

There are several problems with this. For one, why does it say "platform:Javascript" when this is Windows? For another, the triggers don't seem to map correctly. What was *really* confusing is that **"Clear Mapping" doesn't actually set it to the default mapping,** and clicking "Clear Mapping messes up the whole mapping (while the default mapping when starting the program is correct).

For the left trigger, the default mapping has it as Button 6 (interpreted as left trigger), and Steam detects it as Button 12 (interpreted as left trigger), and the remapper detects it as Button 12 (interpreted as D-pad up), then clicking "Clear Mapping" sets it to Button 12 (interpreted as D-pad up).

For the right trigger, the default mapping has it as Button 7 (interpreted as right trigger), and Steam detects it as Button 11 (interpreted as right trigger), and the remapper detects it as Button 7 (interpreted as left trigger), then clicking "Clear Mapping" sets it to Button 11 (interpreted as the Start button).

For the D-pad, Godot detects it as buttons 12-15, but Steam detects it as "Hat 0.1" for up, "Hat 0.8" for left, "Hat 0.4" for down, and "Hat 0.2" for right. The remapper works for the D-pad, but this behavior seems strange.

When clicking "Clear Mapping", the triggers are interpreted as D-pad up and Start, and Select/Start are interpreted as the joystick clicks, and the joystick clicks are interpreted as the triggers. Nothing maps to select, two things map to D-pad up.

Then I tried on 3.2.4 beta 6 and I got the exact same behavior, same map string, same bad behavior, same broken "Clear Mapping" button.

There is also the popup with "Try a common mapping" that I can see in the Godot editor, it doesn't show up when clicking any of the buttons on the bottom (Remap starts the wizard immediately, Show brings up text, and Clear Mapping doesn't open anything), so it seems that this menu is impossible to get to (and should either become accessible, or deleted).

Faless:
> I can't map the A button unless I click on a UI element to change the focused button

That is indeed a bug, we need to ignore the gamepad UI actions, I did that in my test project but forgot about it.
I will try to find a workaround for it.

> What was _really_ confusing is that **"Clear Mapping" doesn't actually set it to the default mapping,** and clicking "Clear Mapping" messes up the whole mapping (while the default mapping when starting the program is correct).

As far as I understand resetting to the godot builtin mapping is not possible via scripting API, "clear mapping" removes the current mapping, and shows "raw" inputs.

> For the D-pad, Godot detects it as buttons 12-15, but Steam detects it as "Hat 0.1" for up, "Hat 0.8" for left, "Hat 0.4" for down, and "Hat 0.2" for right. The remapper works for the D-pad, but this behavior seems strange.

Godot exposes Hats as buttons. So, there isn't much we can do about it.

> There is also the popup with "Try a common mapping" that I can see in the Godot editor, it doesn't show up when clicking any of the buttons on the bottom (Remap starts the wizard immediately, Show brings up text, and Clear Mapping doesn't open anything), so it seems that this menu is impossible to get to (and should either become accessible, or deleted).

That's only for platforms that have known mappings (for now, HTML5 only).

Faless:
@aaronfranke 
Thanks a lot for the review.
I've fixed both the issue with the platform name, and the one about input not being properly consumed.

misc/joypads/remap/remap_wizard.tscn
```diff
@@ -0,0 +1,238 @@
...
+margin_bottom = 20.0
+text = "Xbox (OSX)"
+
+[node name="Label" type="Label" parent="Start/Layout"]
+margin_top = 98.0
+margin_right = 256.0
+margin_bottom = 112.0
+text = "Or start the wizard"
+
+[node name="buttons" type="HBoxContainer" parent="Start/Layout"]
```
aaronfranke, line 67: Node names should be in PascalCase, so this one should be `Buttons`, not `buttons`.

I'm not sure if you have a specific reason for `html5` and `xbox` to be lowercase, if not please change them too.

misc/joypads/remap/remap_wizard.gd
```diff
@@ -0,0 +1,188 @@
...
+onready var joy_mapping_full_axis = $Mapping/Info/Extra/FullAxis
+onready var joy_mapping_axis_invert = $Mapping/Info/Extra/InvertAxis
+
+var joy_guid = ""
+var joy_name = ""
+
+var steps = JoyMapping.BASE.keys()
+var cur_step = -1
+var cur_mapping = {}
+var last_mapping = ""
```
aaronfranke, line 18: Per [the style guide's code order section](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html#code-order), `onready var`s go below non-onready `var`s to be closer to the lifecycle methods (which aren't present in this script, but the rule still applies).

misc/joypads/remap/remap_wizard.gd
```diff
@@ -0,0 +1,188 @@
...
+
+func _on_Wizard_pressed():
+	Input.remove_joy_mapping(joy_guid)
+	$Start.hide()
+	$Mapping.popup_centered()
+	cur_step = 0
+	step_next()
+
+
+func _input(event):
```
aaronfranke, line 113: Per [the style guide's code order section](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html#code-order), built-in virtual methods go above others, so `_input` should go on top. The rest of the methods are in a good order.

aaronfranke (changes requested):
Now this is what I get when mapping a Stadia controller on Windows:

    03000000d11800000094000000000000,Stadia Controller,a:b0,b:b1,y:b3,x:b2,start:b9,back:b8,leftstick:b6,rightstick:b7,leftshoulder:b4,rightshoulder:b5,dpup:b12,dpleft:b14,dpdown:b13,dpright:b15,leftx:a0,lefty:a2,rightx:a3,righty:a5,lefttrigger:b12,righttrigger:b11,platform:Windows

I am able to map all inputs without the menu closing when pressing A, and the platform is reported correctly. Using the remapper tool, the only button bound incorrectly is the left trigger which is detected as D-pad up. In my last comment I reported that the right trigger was bound incorrectly, but it seems to be working now.

> As far as I understand resetting to the godot builtin mapping is not possible via scripting API, "clear mapping" removes the current mapping, and shows "raw" inputs.

I see, but we still need to make it clear what's happening, since "Clear Mapping" does not convey this. What about "Raw Mapping", "Set to Raw Mapping", or "Set to Raw"?

misc/joypads/remap/remap_wizard.gd
```diff
@@ -0,0 +1,188 @@
...
+
+func step_next():
+	$Mapping.window_title = "Step: %d/%d" % [cur_step+1, steps.size()]
+	joy_mapping_text.text = ""
+	if cur_step >= steps.size():
+		remap_and_close(cur_mapping)
+	else:
+		_update_step()
+
+func show_map():
```
aaronfranke, line 73:
```suggestion
		_update_step()


func show_map():
```

misc/joypads/remap/remap_wizard.gd
```diff
@@ -0,0 +1,188 @@
...
+	$Start.hide()
+	$Mapping.hide()
+	joy_guid = ""
+	joy_name = ""
+	cur_mapping = {}
+	cur_step = -1
+
+
+func step_next():
+	$Mapping.window_title = "Step: %d/%d" % [cur_step+1, steps.size()]
```
aaronfranke, line 66:
```suggestion
	$Mapping.window_title = "Step: %d/%d" % [cur_step + 1, steps.size()]
```

misc/joypads/remap/remap_wizard.gd
```diff
@@ -0,0 +1,188 @@
...
+	joy_mapping_text.text = ""
+	if cur_step >= steps.size():
+		remap_and_close(cur_mapping)
+	else:
+		_update_step()
+
+func show_map():
+	if OS.get_name() == 'HTML5':
```
aaronfranke, line 74:
```suggestion
	if OS.get_name() == "HTML5":
```

aaronfranke (changes requested):

Faless:
Okay, I think I addressed all the remaining issues.

aaronfranke (approved):
Looks good to me.

PR #577, opened 2021-01-22
Add physics tests for 2D character controller

pouleyKetchoupp:
Two tests for character controller, with options to use RigidBody2D, KinematicBody2D or KinematicBody2D with RayShape2D.

Tilemap: Tests for moving and jumping within tilemap blocks, with a specific one-way collision test case scenario based on Block Climb Test from https://github.com/madmiraal/godot-gym (CC @madmiraal).

Slopes: Tests for moving and jumping in slopes, with different cases based on snap and stop-on-slope parameters for kinematic bodies.


2d/physics_tests/project.godot
```diff
@@ -72,6 +78,21 @@ toggle_pause={
 "events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":80,"unicode":0,"echo":false,"script":null)
  ]
 }
+character_left={
+"deadzone": 0.5,
+"events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":16777231,"unicode":0,"echo":false,"script":null)
```
aaronfranke, line 83: Normally I would suggest binding WASD in addition to the arrow keys. However, in this project the arrow keys are used for the UI, so perhaps the arrow keys should be unbound too.

2d/physics_tests/tests.gd
```diff
@@ -18,6 +18,14 @@ var _tests = [
 		"id": "Functional Tests/Collision Pairs",
 		"path": "res://tests/functional/test_collision_pairs.tscn",
 	},
+	{
+		"id": "Functional Tests/Controller - Slopes",
+		"path": "res://tests/functional/test_controller_slopes.tscn",
```
aaronfranke, line 23: Instead of "Controller - Thing" I think it should be "Character - Thing". The file name can be changed too.

2d/physics_tests/project.godot
pouleyKetchoupp: Ok, I'm actually going to bind keys differently in a separate PR in order to free WASD keys, and I'll also unbind arrows from UI.

pouleyKetchoupp:
Renamed `controller` to `character` in test files and bound character movement to WASD on top of arrows (requires #578 to avoid key conflicts).

aaronfranke (approved):

PR #580, opened 2021-02-13
Simplify the 3D scaling demo setup, enable filtering by default

Calinou:
It turns out using a TextureRect node isn't necessary :slightly_smiling_face: 
I found this out while working on https://github.com/RodZill4/material-maker/pull/284.

The typical ViewportContainer + Viewport setup can be used just fine to enable filtering on the ViewportTexture returned by the Viewport.

The performance of the new method is equivalent to the old one.

### DIFF
--- a/viewport/3d_scaling/hud.gd
+++ b/viewport/3d_scaling/hud.gd
@@ -1,20 +1,18 @@
 extends Control
 
-# The viewport is displayed using a TextureRect node instead of a ViewportContainer.
-# This allows filtering the texture that's displayed in the root viewport.
-
 # The 3D viewport's scale factor. For instance, 1.0 is full resolution,
 # 0.5 is half resolution and 2.0 is double resolution. Higher values look
 # sharper but are slower to render. Values above 1 can be used for supersampling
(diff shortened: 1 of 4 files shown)


viewport/3d_scaling/hud.gd
```diff
@@ -1,20 +1,19 @@
...
-# This allows filtering the texture that's displayed in the root viewport.
-
 # The 3D viewport's scale factor. For instance, 1.0 is full resolution,
 # 0.5 is half resolution and 2.0 is double resolution. Higher values look
 # sharper but are slower to render. Values above 1 can be used for supersampling
-# (SSAA), but filtering must be enabled for this to work.
+# (SSAA), but filtering must be enabled for supersampling to work.
 var scale_factor = 1.0
 
 onready var texture_rect = $TextureRect
```
aaronfranke, line 9: This line needs to be deleted because there isn't a TextureRect anymore.

Tip: Check the debugger tab for errors before submitting a PR :)

aaronfranke (changes requested):

aaronfranke (approved):

PR #581, opened 2021-02-21
Improve controls and camera handling in the Truck Town demo

Calinou:
- Increase movement and steering speed significantly.
- Increase gravity slightly for a less floaty feeling.
- Simplify camera code (height is now constant).
- Tweaked camera distance and height to fit each vehicle.
- Fix script error when pressing Escape in the main menu (the demo will now quit).

### DIFF
--- a/3d/truck_town/car_base.tscn
+++ b/3d/truck_town/car_base.tscn
@@ -207,7 +207,9 @@ current = true
 fov = 74.0
 near = 0.1
 script = ExtResource( 5 )
-max_distance = 7.0
+min_distance = 3.0
+max_distance = 4.0
+height = 1.25
 
 [node name="AnimationPlayer" type="AnimationPlayer" parent="."]
 _import_path = NodePath("AnimationPlayer")
--- a/3d/truck_town/car_select.gd
+++ b/3d/truck_town/car_select.gd
@@ -18,8 +18,13 @@ func _load_scene(car):
 
 
 func _on_Back_pressed():
-	town.queue_free()
-	show()
+	if is_instance_valid(town):
+		# Currently in the town, go back to main menu.
+		town.queue_free()
+		show()
+	else:
+		# In main menu, exit the game.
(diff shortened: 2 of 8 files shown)


aaronfranke:
(Note: I'm not an expert in vehicle mechanics, so take what I say with a grain of salt)

> * Increase movement and steering speed significantly.
> * Increase gravity slightly for a less floaty feeling.

Don't these cancel each other out? I mean, the whole reason that the vehicles end up in the air is because they run into a bump at high speeds. By increasing the vehicle speed, you are increasing the need to pull the vehicles down, which also will increase the vehicle acceleration when moving downhill and deceleration when moving uphill, which might not be desired. I'm also not sure that increasing the speed is a good idea to begin with, at some point the vehicles start feeling like toys.

As far as I can tell we want the vehicles to be responsive when slow, while not allowing them to go crazy fast such that they go flying when they hit a bump. I think the root of the problem is that we are treating the W key as flooring it, as we don't have a way for the user to adjust the engine force themselves - so if this single value is too low, a vehicle can't climb a steep incline, and if it's too high, a vehicle will accelerate really fast.

So, why not just adjust the force value for the player, making it higher if the vehicle is slow? I made an experimental commit, it seems to work fairly well, let me know what you think. The force starts at 40, increases if the speed is less than 5 m/s (18 km/h or 11.1 mph), maxing out at 100 force, which occurs when the speed is less than or equal to 2 m/s (7.2 km/h or 4.5 mph).

https://github.com/aaronfranke/godot-demo-projects/commit/6c00cf8f85dcdb33c9f5ad8a3d3a29150be78627#diff-4f93ccc27dd799129ca767f92327c1e99f71d05456174509c95f7a97a6b52c42R16-R31

Calinou:
@aaronfranke I tested your PR and it seems like a good change to me. Feel free to merge this with your PR applied on top :slightly_smiling_face:

PR #584, opened 2021-02-23
Updated 2D character controller physics tests

pouleyKetchoupp:
Added new test for 2D character controller:
Character - Pixels
Functional test for pixel art related issues around KinematicBody and RigidBody character controllers.

Adjusted existing tests and added more test cases to cover most use cases from recent fixed issues and regressions for KinematicBody.

Added a more automated way to run all tests with checks to see which ones failed in character controller tests.

Also fixed some minor issues with the log scrollbar.

### DIFF
--- a/2d/physics_tests/tests/functional/test_one_way_collision.gd
+++ b/2d/physics_tests/tests/functional/test_one_way_collision.gd
@@ -2,23 +2,39 @@ extends Test
 tool
 
 
+signal all_tests_done()
+signal test_done()
+
 const OPTION_OBJECT_TYPE_RIGIDBODY = "Object type/Rigid body (1)"
 const OPTION_OBJECT_TYPE_KINEMATIC = "Object type/Kinematic body (2)"
 
-const OPTION_TEST_CASE_ALL_ANGLES = "Test case/Around the clock (0)"
+const OPTION_TEST_CASE_ALL = "Test Cases/TEST ALL (0)"
+const OPTION_TEST_CASE_ALL_RIGID = "Test Cases/All Rigid Body tests"
+const OPTION_TEST_CASE_ALL_KINEMATIC = "Test Cases/All Kinematic Body tests"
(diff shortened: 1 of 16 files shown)


2d/physics_tests/tests/functional/test_one_way_collision.gd
```diff
@@ -34,7 +48,13 @@ func _ready():
...
 		$Options.add_menu_item(OPTION_OBJECT_TYPE_KINEMATIC, true, _use_kinematic_body, true)
 
-		$Options.add_menu_item(OPTION_TEST_CASE_ALL_ANGLES)
+		$Options.add_menu_item(OPTION_TEST_CASE_ALL)
+		$Options.add_menu_item(OPTION_TEST_CASE_ALL_RIGID)
+		$Options.add_menu_item(OPTION_TEST_CASE_ALL_KINEMATIC)
+		$Options.add_menu_item(OPTION_TEST_CASE_ALL_ANGLES_RIGID)
+		$Options.add_menu_item(OPTION_TEST_CASE_ALL_ANGLES_KINEMATIC)
+		$Options.add_menu_item(OPTION_TEST_CASE_MOVING_PLATFORM_RIGID)
+		$Options.add_menu_item(OPTION_TEST_CASE_MOVING_PLATFORM_KINEMATIC)
```
aaronfranke, line 57: Instead of doing this, we should have `onready var options = $Options` and then reference it everywhere. This should also be done for the other files. We want to avoid the inefficiency of getting a node more times than necessary.

aaronfranke (changes requested):

2d/physics_tests/tests/functional/test_one_way_collision.gd
pouleyKetchoupp: It's true, even if it's not critical here it's not good practice. I'll change it in the different files. Thanks!

aaronfranke (approved):
Forgot to give this an approval earlier, looks good!

PR #586, opened 2021-02-28
Improve several demos' configuration

Calinou:
- Enable anisotropic filtering in 3D demos, and set the quality to 16× on desktop and 4× on mobile.
- Enable 4× MSAA on some 3D demos that didn't use it beforehand.
- On GLES3 demos, disable MSAA on mobile as these demos are often more demanding.
- Use more conservative framebuffer allocation settings for better performance.
- Use PCF13 shadow filtering in GLES2 demos on desktop to benefit from soft shadows in Godot 3.2.4 and later. In Godot 3.2.3, this will make shadows smoother but still "blocky", so this can be merged right away.
- Use Lossless compression instead of VRAM compression for small textures such as the voxel demo texture atlas.


3d/voxel/menu/main/main_menu.tscn
```diff
@@ -228,6 +228,7 @@ __meta__ = {
 }
 
 [node name="Options" parent="." instance=ExtResource( 5 )]
+
```
aaronfranke, line 231: Diffs like these are caused by Godot 3.2.4, which is not out yet. Please open and save scenes in Godot 3.2.3 (every scene that you modified in this PR).

aaronfranke (changes requested):
To be honest, I find it very difficult to tell the difference. But it does look slightly better, so overall this seems good.

aaronfranke (approved):

PR #588, opened 2021-02-28
Port the Bullet Shower demo from Godot 2.1

Calinou:
This demo showcases how to use low-level Servers to achieve better CPU performance when drawing large amounts of objects.

The code has been updated for Godot 3.2, cleaned up and has received additional comments.

## Preview

No collision

Collision


2d/bullet_shower/shower.gd
```diff
@@ -0,0 +1,34 @@
...
+	# in displaying the mouse cursor.
+	Input.set_mouse_mode(Input.MOUSE_MODE_HIDDEN)
+
+
+func _input(event):
+	if event is InputEventMouseMotion:
+		player.position = event.position - Vector2(0, 16)
+
+
+func _on_player_body_shape_entered(body_id: int, body: Node, body_shape: int, local_shape: int) -> void:
```
aaronfranke, line 25: Per Akien's request, the demos don't use static typing (with a few exceptions). We also need underscores to mark these arguments as unused.

```suggestion
func _on_player_body_shape_entered(_body_id, _body, _body_shape, _local_shape):
```

2d/bullet_shower/shower.gd
```diff
@@ -0,0 +1,34 @@
...
+	touching += 1
+	if touching >= 1:
+		player_sprite.frame = 1
+
+
+func _on_player_body_shape_exited(body_id: int, body: Node, body_shape: int, local_shape: int) -> void:
```
aaronfranke, line 31:
```suggestion
func _on_player_body_shape_exited(_body_id, _body, _body_shape, _local_shape):
```

2d/bullet_shower/bullets.gd
```diff
@@ -0,0 +1,82 @@
...
+
+
+func _ready():
+	randomize()
+
+	shape = Physics2DServer.circle_shape_create()
+	# Set the collision shape's radius for each bullet in pixels.
+	Physics2DServer.shape_set_data(shape, 8)
+
+	for i in range(BULLET_COUNT):
```
aaronfranke, line 29:
```suggestion
	for _i in range(BULLET_COUNT):
```

aaronfranke (changes requested):
Overall the code looks good. I noticed that @reduz said [here](https://github.com/godotengine/godot-proposals/issues/2380#issuecomment-787525803) that this is "very far from being the most optimized form", so I wonder if there are ways to optimize this further without engine changes (if so, that should be done in this demo).

The images can be compressed with oxipng for slightly less file size: [bullet-shower-images.zip](https://github.com/godotengine/godot-demo-projects/files/6058042/bullet-shower-images.zip)

aaronfranke:
The code in `shower.gd` is for the player, so I think it would make more sense to have a `player.gd` like this:

```gdscript
extends Node2D
# This demo is an example of controling a high number of 2D objects with logic
# and collision without using nodes in the scene. This technique is a lot more
# efficient than using instancing and nodes, but requires more programming and
# is less visual. Bullets are managed together in the `bullets.gd` script.

# The number of bullets currently touched by the player.
var touching = 0

onready var sprite = $AnimatedSprite


func _ready():
	# The player follows the mouse cursor automatically, so there's no point
	# in displaying the mouse cursor.
	Input.set_mouse_mode(Input.MOUSE_MODE_HIDDEN)


func _input(event):
	if event is InputEventMouseMotion:
		position = event.position - Vector2(0, 16)


func _on_player_body_shape_entered(_body_id, _body, _body_shape, _local_shape):
	touching += 1
	if touching >= 1:
		sprite.frame = 1


func _on_player_body_shape_exited(_body_id, _body, _body_shape, _local_shape):
	touching -= 1
	if touching == 0:
		sprite.frame = 0
```

\+ delete `shower.gd` + the signals need to be updated + maybe the comment at the top should be moved/copied to `bullets.gd`.

aaronfranke (approved):

TackerTacker:
The bullets in this example are all set to the default collision layer so I think they are checking collisions with each other.

Adding `Physics2DServer.body_set_collision_layer(bullet.body, 0)` at line 41 in `bullets.gd` makes the example a little bit more perfomant. After this change it can handle 1600 bullets instead of 1300 without frame drops on my machine.
Though I have to say, this is still pretty low IMO, especially for an example that's suppose to be about the most efficient way of doing this.

My hardware isn't great, but I can build the same example with 7 lines of code in Construct 3 and can have 15000 bullets without dropping frames.
Is this as good as it gets, or is there something wrong with the example? I don't think this is a great look for Godot since it is specifically made as an example about performance.

Calinou:
> After this change it can handle 1600 bullets instead of 1300 without frame drops on my machine.

See https://github.com/cart/godot3-bunnymark.

I would try building this Bullet Shower demo with nodes and see how it performs in comparison to the demo that uses servers.

PR #589, opened 2021-03-02
Improve the 3D platformer demo

Calinou:
- Increase gravity, player speed and air acceleration for a less floaty feeling.
- Clean up the code to use constants when relevant and remove unused enums.
- Don't use VRAM compression for the roughness/metallic texture as it's low-resolution.


aaronfranke (approved):
I still hesitate to change the gravity setting as I want the 3D demos to try and reflect real world scales. However, in this demo, it's generally not expected to be realistic, so it makes sense for here so that we can get accelerated movement mechanics. Also, using constants/enums/methods is very appreciated, as are the other tweaks.

PR #590, opened 2021-03-03
Improve the Kinematic Character 3D demo

aaronfranke:
This PR improves the Kinematic Character 3D demo in the following ways:

* Save meshes in GLTF `.glb` files and a `.mesh` file, and save GIProbe data to a `.res` file, instead of having everything stored as text in scene files.
* Use only one wood texture, with an albedo for the purple texture, instead of having a separate purple texture.
* Delete unused `purple_cube.tscn` scene.
* All of the above reduce the file size. Excluding screenshots, the demo has gone from 4.1 MB to under 1 MB.
* Scale down the world such that each cube is 1x1x1 instead of 2x2x2. The meshes themselves were modified for this. This also has the effect of making movement faster without changing gravity.
* Change the wood material so that it's vastly less metallic and reflective. I don't really know why we had shiny wood in the first place, wood is usually not shiny in the real world unless it's coated with a shiny finish.
* Organize into folders: a models folder, and a player folder.
* Update icon and screenshots. The icon is now 128x128.


Calinou:
> I don't really know why we had shiny wood in the first place, wood is usually not shiny in the real world unless it's coated with a shiny finish.

Gotta showcase GIProbe reflections somehow. :stuck_out_tongue:

PR #595, opened 2021-03-12
Small improvement for Bullet Shower Demo

dalexeev:
Move physics processing code from `_process()` to `_physics_process()`.

### DIFF
--- a/2d/bullet_shower/bullets.gd
+++ b/2d/bullet_shower/bullets.gd
@@ -52,21 +52,24 @@ func _ready():
 		bullets.push_back(bullet)
 
 
-func _process(delta):
+func _process(_delta):
+	# Order the CanvasItem to update every frame.
+	update()
+
+
+func _physics_process(delta):
 	var transform2d = Transform2D()
+	var offset = get_viewport_rect().size.x + 16
 	for bullet in bullets:
 		bullet.position.x -= bullet.speed * delta
 
 		if bullet.position.x < -16:
 			# The bullet has left the screen; move it back to the right.
-			bullet.position.x = get_viewport_rect().size.x + 16
+			bullet.position.x = offset
 
 		transform2d.origin = bullet.position
 
 		Physics2DServer.body_set_state(bullet.body, Physics2DServer.BODY_STATE_TRANSFORM, transform2d)
-
-	# Order the CanvasItem to update since bullets are moving every frame.
-	update()
 
 
 # Instead of drawing each bullet individually in a script attached to each bullet,


2d/bullet_shower/bullets.gd
```diff
@@ -52,22 +52,25 @@ func _ready():
 		bullets.push_back(bullet)
 
 
-func _process(delta):
+func _process(_delta):
+	# Order the CanvasItem to update since bullets are moving every frame.
+	update()
```
aaronfranke, line 57: Since they're only moving every *physics* frame, I suppose the `update()` call should go in `_physics_process` too.

dalexeev:
```
void update()

Queue the CanvasItem for update. NOTIFICATION_DRAW will be called on idle time to request redraw.
```

I thought drawing is about graphics, so I left `update()` in `_process()`. But if you are sure, then I will move it to `_physics_process()`.

dalexeev: I changed this as you requested. But it seems to me that when `update()` was in `_process()`, the movement was a little smoother.

aaronfranke: Well, this was just an idea. If you think it's better to keep it in `_process` then we can do that.

dalexeev: I'm not entirely sure, but my reasoning is this: if `update()` is in `_process()`, then every frame will be guaranteed to be redrawn. And if `update()` in `_physics_process()`, then due to the fact that the physics frames do not always clearly correspond to the graphics frames, situations may arise when the frame will not be redrawn, due to the fact that physics is a little behind. And I don't quite understand if this is good or bad.

aaronfranke: That makes sense, but do we want every idle frame to be redrawn, even if nothing happens? My reasoning was that if two frames are drawn for the same physics frame, then we don't need to update anything, because the second frame would just have the same content (ex: set the physics FPS to 10 and it's jittery either way, but I think that if `update()` is in `_process` then it will draw the same thing multiple times).

dalexeev: I also had this thought, but then I thought: maybe half of the `_physics_process()` will be executed before the idle frame moment comes, the bullet positions will be updated, but the `update()` function will not be called yet? That is, the changes will actually take place, but will not be rendered? Or is it out of the question? Sorry, I don't know much about the low-level part of the engine, so I'm asking this question.

lawnjelly:
This is just copy paste from what I put in rocket chat:
> I think the update should stay in _process
because you are moving the bullets manually.
in _process
If you move the bullets in physicsprocess, to get smooth motion you'd need to use interpolation in _process then call update in process.
But it's not worth it in this case, easier to do everything in process.
The physics being set in process is a bit questionable, but it's only for some rough collision detection by the look of it.
Normally you'd be using physics to drive motion not the other way around.
So using physics here is a bit overkill but it does serve as an example how to create physics bodies etc.

dalexeev:
Yes, in the case of straight uniform motion, `_physics_process()` isn't required. But in more complex cases, it's more correct to implement physics in `_physics_process()`, for the determinism of simulation. That is why I decided to fix this demo.

If I understand correctly, in single-threaded mode it doesn't matter where to place `update()` in this case. Probably, aaronfranke is even right that there is no point in redrawing what has not changed. But, probably, more conceptually correct to place `update()` in `_process()`.

aaronfranke (approved):
I'm still a little uncertain about this, but I think we should merge it now, and if something else is desired in the future, we can just follow this up with another PR.

PR #597, opened 2021-03-12
Handle up to 24 joypad buttons in the Joypads demo

### LINKED ISSUE #596: get an error from Joypads Demo because JOY_BUTTON_MAX's value changed

Joypads Demo / Tool

Screen Shot 2021-03-12 at 16 38 50

**Issue description:**

JOY_BUTTON_MAX is equal to 16 in 3.2.3 and before, but in 3.2.4 it's 22, so right now this will make an error when using "range(JOY_BUTTON_0, JOY_BUTTON_MAX)"

should add "if button_grid.get_node_or_null(str(btn)) and joypad_buttons.get_node_or_null(str(btn))" after "range(JOY_BUTTON_0, JOY_BUTTON_MAX)"
Screen Shot 2021-03-12 at 17 25 52

**Screenshots of issue:**

Screen Shot 2021-03-12 at 16 40 14
Screen Shot 2021-03-12 at 16 40 29
Screen Shot 2021-03-12 at 16 40 39

aaronfranke:
This should fix #596. The GridContainer of buttons has been increased to the range of 0 to 23 (even though it only goes to 22 in 3.2.4, it looks nicer to have a multiple of 8), and there is another hard limit on the controller diagram so that it doesn't go above 16 (perhaps we can add some things to the diagram, but I don't know where these extra buttons are supposed to map to). I also made it so that it gets children by index instead of by name.

@snougo Can you test this in both 3.2.3 and 3.2.4? Ideally it should work on both.

### DIFF
--- a/misc/joypads/joypad_diagram.tscn
+++ b/misc/joypads/joypad_diagram.tscn
@@ -125,6 +125,13 @@ flip_h = true
 region_enabled = true
 region_rect = Rect2( 50, 0, 14, 54 )
 
+[node name="16" type="Sprite" parent="Buttons"]
+position = Vector2( 4, 64 )
+scale = Vector2( 0.9, 0.9 )
+texture = ExtResource( 2 )
+region_enabled = true
+region_rect = Rect2( 0, 0, 45, 45 )
+
 [node name="Axes" type="Node2D" parent="."]
 
 [node name="0-" type="Sprite" parent="Axes"]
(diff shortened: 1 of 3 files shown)


snougo:
yeah, when it get updating in asset store I will download it, if no issue, I will close the issue so you know that it has be fixed

aaronfranke:
> (perhaps we can add some things to the diagram, but I don't know where these extra buttons are supposed to map to)

Tested with a Stadia controller, the Guide button maps to Button 16, and 17-23 are unused. I updated the diagram accordingly.

PR #606, opened 2021-04-13
Updated 2d/3d physics contact performance tests

pouleyKetchoupp:
* Added proper logs for physics ticks
* Optimizations and cleaning in scripts
* In 3D, use `PlaneShape` instead of `ConcavePolygonShape` for walls because it contains the shapes better

### DIFF
--- a/2d/physics_tests/tests/performance/test_perf_contacts.gd
+++ b/2d/physics_tests/tests/performance/test_perf_contacts.gd
@@ -10,11 +10,14 @@ const OPTION_TYPE_CONCAVE_POLYGON = "Shape type/Concave Polygon"
 
 export(Array) var spawns = Array()
 export(int) var spawn_count = 100
-export(int, 1, 10) var spawn_multiplier = 5
 
 onready var options = $Options
 
 var _object_templates = []
+
+var _log_physics = false
+var _log_physics_time = 0
+var _log_physics_time_start = 0
 
 
 func _ready():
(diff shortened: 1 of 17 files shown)


2d/physics_tests/tests/performance/test_perf_contacts.gd
```diff
@@ -132,33 +157,37 @@ func _spawn_objects(type_index):
 
 		Log.print_log("* Spawning: " + template_node.name)
 
-		for _index in range(spawn_multiplier):
-			for _node_index in spawn_count / spawn_multiplier:
-				var node = template_node.duplicate() as Node2D
-				spawn_parent.add_child(node)
+		for _node_index in spawn_count:
```
aaronfranke, line 160: For the sake of clarify I usually prefer `range(spawn_count)`, but either way is fine.

pouleyKetchoupp: Ok, I've changed it to use range, and in some other scripts too.

aaronfranke (approved):

2d/physics_tests/tests/performance/test_perf_contacts.gd
Calinou: Not using `range()` is a bit faster, so I'd recommend that when possible.

aaronfranke: @Calinou Why doesn't the engine optimize this case? It should be trivial to detect when range will have the same behavior as an int.

pouleyKetchoupp: @Calinou Do you know if this extra computation is just once for the whole loop or on each iteration? I haven't tested to see if there's a difference in performance, but if the cost is multiplied by the number of iterations it might be worth changing it back indeed, otherwise it's probably fine.

Calinou: @pouleyKetchoupp I think it's done once for the whole loop.

pouleyKetchoupp: I've made some tests to see what difference I get with performance in 3.2.3:
```
	var time = OS.get_ticks_usec()
	for i in range(100000):
		pass
	var time_total = OS.get_ticks_usec() - time
	Log.print_log("Time Test %.3f ms" % (0.001 * time_total))
```
And I see about the same numbers between `x` and `range(x)` with different values, so it doesn't look like the optimization makes a really significant difference in this case.

PR #607, opened 2021-04-21
Update README for the new 3.2 branch

aaronfranke:
This PR updates the README for Godot 3.3 and a few other misc things.

After this PR, there will be a 3.2 branch, and the master branch of the demos will be targeting Godot 3.3.x.

PR #608, opened 2021-04-21
[2.1] Update README for the new 3.2 branch

aaronfranke:
This PR updates the README of the 2.1 branch to reflect the changes in #607.

In this PR's README file, 2.1 is bold and italicized for easy recognition while browsing, like this: ***2.1***

This PR also backports the gitignore and other metadata, which ensures files have good formatting, and more importantly, avoids showing additions of generated files in other branches when switching to an old branch.


akien-mga:
I'm not sure it's worth updating all old branches with details on what the `master` branch or newer branches are about. The README for old branches should just be clear about what Godot version they are compatible with, but shouldn't have to give information about future releases IMO. Otherwise when would we stop updating those old branches?

aaronfranke:
@akien-mga Yeah, I've thought about this, IMO it's nice to be able to easily navigate to any branch from any other branch (at least, between Godot versions), and I have no problem with doing the work to keep them in sync. I suppose it's not really necessary, but I don't really see a strong reason to *not* do this.

PR #615, opened 2021-05-05
Fix hack to draw shapes in 3D physics tests

pouleyKetchoupp:
Using `Shape::get_debug_mesh` instead of the previous hack to draw a mesh instance based on a collision shape.

Needs https://github.com/godotengine/godot/pull/48316 to work (cherry-picked for 3.3.1).
It will be the only option after https://github.com/godotengine/godot/pull/48175 is merged.

### DIFF
--- a/3d/physics_tests/test.gd
+++ b/3d/physics_tests/test.gd
@@ -46,28 +46,19 @@ func add_sphere(pos, radius, color):
 
 
 func add_shape(shape, transform, color):
-	var body = StaticBody.new()
-	body.collision_layer = 0
-	body.collision_mask = 0
+	var debug_mesh = shape.get_debug_mesh()
 
-	var collision = CollisionShape.new()
-	collision.transform = transform
-	collision.shape = shape
+	var mesh_instance = MeshInstance.new()
+	mesh_instance.transform = transform
+	mesh_instance.mesh = debug_mesh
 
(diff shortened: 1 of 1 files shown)


aaronfranke:
> It will be the only option after godotengine/godot#48175 is merged.

If godotengine/godot#48175 breaks compatibility, then why is it marked as cherry-pick for 3.3? Shouldn't it only be 3.x at most?

pouleyKetchoupp:
>> It will be the only option after godotengine/godot#48175 is merged.
>
> If godotengine/godot#48175 breaks compatibility, then why is it marked as cherry-pick for 3.3? Shouldn't it only be 3.x at most?

In 3.3, debug mesh instances have been moved from being children of the `CollisionShape` nodes to children of the `CollisionObject` nodes. But it's causing issues as `CollisionShape` nodes now have unexpected siblings. We got several bug reports around that already.
So godotengine/godot#48175 can fix that in 3.3.1 without losing the original benefit (support for custom collision shapes).

The hack I've been using to retrieve debug mesh instances from the tree nodes in order to do custom drawing is more of an edge case. Even if it can be considered compatibility breaking in some way, it's probably fine to just use the new API instead when it's needed. I doubt many users will complain about it.

aaronfranke (approved):

PR #622, opened 2021-05-28
Have waypoints check if the camera is current

### LINKED ISSUE #605: Wrong waypoint drawing when switching camera
3D waypoint

Ubuntu 16 

**Issue description:**
Waypoints are not drawn at the right positions if switching camera at runtime.
Expected the waypoint system to detect and use the right current camera.

**Screenshots of issue:**
Just create 2 cameras in a scene and switch between cameras: see that the waypoints are hard coded to only draw according to the camera found at init/ready time.
See :
onready var camera = get_viewport().get_camera()
in waypoint.gd

A possible fix is to check in waypoint _process() that the 'camera' is current and if not, updating the camera var with the viewport camera. But there is perhaps better.

Tks for this cool util btw.

aaronfranke:
If not, get the current camera. Implements and closes #605. This isn't really necessary, but it's a simple fix and it's a cheap check.

### DIFF
--- a/3d/waypoints/waypoint.gd
+++ b/3d/waypoints/waypoint.gd
@@ -23,6 +23,9 @@ func _ready() -> void:
 
 
 func _process(_delta):
+	if not camera.current:
+		# If the camera we have isn't the current one, get the current camera.
+		camera = get_viewport().get_camera()
 	var parent_translation = parent.global_transform.origin
 	var camera_transform = camera.global_transform
 	var camera_translation = camera_transform.origin
--- a/3d/physics_tests/utils/rigidbody_ground_check.gd
+++ b/3d/physics_tests/utils/rigidbody_ground_check.gd
@@ -44,4 +44,3 @@ func ground_check():
 		_is_on_floor = true
 	else:
 		_is_on_floor = false
-


skaiware:
Thanks @aaronfranke

3d/waypoints/waypoint.gd
```diff
@@ -23,6 +23,8 @@ func _ready() -> void:
 
 
 func _process(_delta):
+	if not camera.current:
+		camera = get_viewport().get_camera()
```
Calinou, line 27:
```suggestion
	if not camera.current:
		# Set the current camera every frame. This ensures that switching cameras
		# doesn't break waypoints (e.g. when copying this code to another project).
		camera = get_viewport().get_camera()
```

Calinou (commented):
Maybe add a comment to explain why this is done:

3d/waypoints/waypoint.gd
aaronfranke: This comment is not necessary, and also the first part is not correct. It doesn't set the camera every frame, it checks if the camera is current every frame, and if it's not then only then it sets the camera.

PR #627, opened 2021-06-21
Replace built-in scripts and delete strange "e" scripts

aaronfranke:
Keeping GDScript files in their own file makes things easier to work with, and should be preferred for the demos.

Also, this PR deletes the strange "func e()" methods. No idea where they came from, but they're useless and now they're gone.


akien-mga (approved):

akien-mga:
Thanks!

PR #633, opened 2021-07-06
Use the Jump sound already present in the Platformer 2D demo

starptr:
Small PR to utilize the jump sound effect node already present in the Player node tree


aaronfranke (approved):
Nice catch! Must have been a regression from when GDQuest recreated the demo from scratch.

To be honest, I'm surprised that I didn't notice the lack of the sound before. Glad it's back though.

PR #638, opened 2021-08-04
Convert physics test projects to 4.0

pouleyKetchoupp:
Everything seems to work except for a few specific issues.
I'm listing all the changes I've made in case it helps for future conversions and for keeping track of possible issues for users.

**General changes:**
- `Control` nodes: `margin_*` renamed to `offset_*`
- Annotations: `@export`, `@export_range`, `@tool`, `@onready`
- Some export variables need to be static typed since export annotations don't include a type anymore
- Signal connections using `Callable`
- Variable setters and getters using set/get functions instead of setget keyword
- Some variable setters need to be changed to avoid a stack overflow
- Renamed 3D classes referenced in gdscript
- Using `await` instead of `yield`, no more `completed` signal
- Using `super` calls in built-on functions to call the parent script when needed
- Uppercase color constants
- Renamed `empty()` to `is_empty()` in several classes
- Manual conversion from `StringName` to `String`
- Renamed `pause_mode` to `process_mode`
- Can't set `material_override` in `GeometryInstance3D`, need to use `set_surface_override_material()` in `MeshInstance3D` instead
- Replace `ImmediateGeometry` node with `MeshInstance3D` node and `ImmediateMesh` resource
- Renamed `BUTTON_*` to `MOUSE_BUTTON_*`
- Renamed `scancode` to `keycode`
- Renamed `OS.get_ticks_usec()` to `Time.get_ticks_usec()`
- Renamed `instance()` to `instantiate()`
- Use `DisplayServer.window_get_mode()` and `DisplayServer.window_set_mode()` instead of `OS.window_fullscreen`
- Project setting `window/stretch/mode`: `canvas_item` instead of `2d`
- Project input mapping has to be set again by hand, all keys are missing
- Can't make a custom script that extends `Slider` anymore, need to use `HSlider` or `VSlider`
- `TileMap` collisions were offseted and cell size was reset

**Physics specific changes:**
- `RectangleShape2D` node: `extents` renamed to `size` and values doubled
- `KinematicBody` -> `CharacterBody` for character controllers
- `KinematicBody` -> `StaticBody` with `kinematic_motion` for moving platforms
- Renamed physics classes with 2D and 3D suffix in gdscript
- Renamed `GodotPhysics` engine string to `GodotPhysics2D` and `GodotPhyics3D`
- Renamed `MODE_RIGID` to `MODE_DYNAMIC` in `RigidBody`
- Properties instead of arguments in `move_and_slide()`
- 3D capsule is up by default
- One directional collision layers require new setup in some cases
- Adjustments for `RigidBody2D` due to increased default gravity
- RayShape2D/RayShape3D have been removed

**Notes:**
- ~Regression in "Character - TileMap": one-way collision are not working at the moment (PR: https://github.com/godotengine/godot/pull/51259).~
**edit:** fix is merged
- ~Regression in "Character - Pixels": test case "floor detection with motion changes" fails.~
**edit:** just made a typo when updating a rectangle size, fixed now.
- ~Rendering issue in "Character - Pixels": it was previously using a `Viewport` node with lower resolution. It's all blurry now and I didn't find how to properly set pixelated rendering in 4.0.~
**edit:** fixed by setting `size_2d_override` and `size_2d_override_stretch` in `SubViewport`


pouleyKetchoupp:
Switched to draft because I noticed a couple things I can still fix before merging.

pouleyKetchoupp:
Everything should be ready now, I've fixed the last remaining issues.

pouleyKetchoupp:
I've just updated the 2D character tests with needed changes from https://github.com/godotengine/godot/pull/51027 and also replaced references to Kinematic Body to Character Body in the scripts and UI.

pouleyKetchoupp:
Updated with changes from https://github.com/godotengine/godot/pull/51490: no more ray shapes and added new CharacterBody parameters instead.

PR #639, opened 2021-08-05
Upgrade demos to Godot 3.4

aaronfranke:
Most of this PR is just opening and saving all the demos, which mostly resulted in adding `process/normal_map_invert_y=false`. There are also a few other changes, such as a physical scancode field for each input in `project.godot`, and I also converted the input systems of many of the demos to use `Input.get_vector`.


3d/voxel/world/chunk.gd
```diff
@@ -61,7 +61,7 @@ func _generate_chunk_collider():
 			_create_block_collider(block_position)
 
 
-func _generate_chunk_mesh(_this_argument_exists_due_to_bug_9924):
+func _generate_chunk_mesh():
```
aaronfranke, line 64: It seems that this bug was fixed? This workaround isn't necessary anymore (and actually the workaround has to be removed or else the demo won't run). https://github.com/godotengine/godot/issues/9924

PR #640, opened 2021-08-14
Mono IAP signal method definition mismatch.

shahediqbal13:
Updated OnPurchaseConsumptionError() method to match with GooglePlayBilling's signal


aaronfranke:
@ricardoalcantara

PR #644, opened 2021-08-22
GitHub Actions: Update to Godot 3.3.3 for HTML5 deployment

Calinou:


aaronfranke (approved):
Should we be doing this for every patch release? I noticed that 3.3.1 and 3.3.2 were skipped.

Calinou:
> Should we be doing this for every patch release? I noticed that 3.3.1 and 3.3.2 were skipped.

Indeed, I forgot to open a PR for 3.3.1 and 3.3.2.

PR #654, opened 2021-10-15
Remove useless "_ready" in the Circle Pop demo

aaronfranke:
Circle Pop is mostly a great example of VisualScript, except for this function which was bogus. It has now been removed.


akien-mga:
Thanks!

PR #655, opened 2021-10-25
Don't check bullet collision with other bullets in the Bullet shower demo

Calinou:
This improves performance by reducing the number of collision checks that need to be done.

I tested the demo and it still works as expected.

Thanks to @TackerTacker for the [suggestion](https://github.com/godotengine/godot-demo-projects/pull/588#issuecomment-803272091) :slightly_smiling_face:


wangbinyq:
Why this will disable player and bullets collision in godot 4?

wangbinyq:
> Why this will disable player and bullets collision in godot 4?
It should set bullet's mask to zero.

PR #657, opened 2021-10-30
Allow 3D demos to fallback to GLES2

Calinou:
Follow-up to https://github.com/godotengine/godot-demo-projects/pull/656.

A GLES2-specific tweak is now documented for the GUI in 3D demo.

This also tweaks shadow quality and performance to find a better balance and result in smoother shadows, including when using GLES2 (thanks to the use of PCF13 shadow filter mode).

In demos that allow falling back to GLES2, ETC1 texture import is enabled to allow exporting to Android or HTML5 without having to tweak anything.


3d/material_testers/project.godot
```diff
@@ -24,6 +24,10 @@ window/dpi/allow_hidpi=true
 window/stretch/mode="2d"
 window/stretch/aspect="expand"
 
+[editor]
+
+main_run_args="--video-driver GLES2"
+
```
aaronfranke, line 30: I think this was left in here by mistake.

aaronfranke (commented):
The demos do look fairly different in GLES2, but allowing a fallback makes sense to me since it will only (positively) affect people trying to run the demos on low-end devices without GLES3.

The CI checks fail due to trailing whitespace.

3d/material_testers/project.godot
Calinou: Indeed, I left that here to test the GLES2 fallback locally. I removed it.

aaronfranke (approved):

PR #662, opened 2021-11-11
Fix stall when playing recorded audio

### LINKED ISSUE #661: mic_record freezes when playing recorded audio
/audio/mic_record/

Tested on Godot 3.4 / Windows 10, likely affects all platforms

**Issue description:**
Demo freezes when attempting to play back recorded audio

**Screenshots of issue:**
N/A

The problem is in the script attached to the main scene root node:

    func _on_PlayButton_pressed():
        print("Play button pressed")
        print(recording)
        print(recording.format)
        print(recording.mix_rate)
        print(recording.stereo)
        var data = recording.get_data()
        print(data)
        print(data.size())
        $AudioStreamPlayer.stream = recording
        $AudioStreamPlayer.play()

print(data) prints a byte array of the entire recording to the console. It doesn't actually crash Godot, but it does make it unresponsive for an extremely long time, even for short recordings. Removing this line fixes the issue.

IMO we should simply remove this line as it doesn't really provide any benefit to the demo.

benjarmstrong:
Resolves #661

### DIFF
--- a/audio/mic_record/MicRecord.gd
+++ b/audio/mic_record/MicRecord.gd
@@ -30,7 +30,6 @@ func _on_PlayButton_pressed():
 	print(recording.mix_rate)
 	print(recording.stereo)
 	var data = recording.get_data()
-	print(data)
 	print(data.size())
 	$AudioStreamPlayer.stream = recording
 	$AudioStreamPlayer.play()


Calinou (approved):
Thanks!

aaronfranke (approved):

PR #667, opened 2021-12-05
[3.x] Calling RPC functions in the webrtc signaling demo

mashumafi:
I am working with an application similar to the webrtc_signaling demo. I was experiencing issues where RPC calls are just dropped without warning. I was able to reproduce it using the signaling demo.

In this PR I setup the `network_peer` in order to get RPC working. I added a new label and and RPC function that randomly updates the color of that label. The function is marked `remotesync` and should update all clients whenever the button is cicked. In this demo though, there are some cases where clicking the button does not update the other client. This occurs even in a small network mesh.

Below is an example of running the application. You chose host in one and connect 1 peer in each client. When they are connected clicking the "Random Color" button should update the color in the other peers. Sometimes this will fail. See the below of an example where the color gets desync.

My best guess is that somehow the signaling server is interfering with the RTC calls.

*Make sure to download the WebRTC libraries in order to test this*

signaling-rpc-bug

### DIFF
--- a/networking/webrtc_signaling/demo/client_ui.gd
+++ b/networking/webrtc_signaling/demo/client_ui.gd
@@ -28,6 +28,7 @@ func _disconnected():
 
 
 func _lobby_joined(lobby):
+	get_tree().network_peer = client.rtc_mp
 	_log("Joined lobby %s" % lobby)
 
 
@@ -77,3 +78,9 @@ func _on_Seal_pressed():
 
 func stop():
 	client.stop()
+
+func _on_RPC_pressed():
+	rpc("update_color", Color(rand_range(0, 1), rand_range(0, 1), rand_range(0, 1), 1))
+
+remotesync func update_color(color: Color):
+	$VBoxContainer/HBoxContainer/Label.modulate = color
--- a/networking/webrtc_signaling/demo/client_ui.tscn
+++ b/networking/webrtc_signaling/demo/client_ui.tscn
@@ -39,7 +39,7 @@ margin_left = 77.0
 margin_right = 921.0
 margin_bottom = 24.0
 size_flags_horizontal = 3
-text = "ws://localhost:9080"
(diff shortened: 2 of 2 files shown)


mashumafi:
Further testing, works well in Chome. Could be an issue directly related to https://github.com/godotengine/webrtc-native on windows

networking/webrtc_signaling/demo/client_ui.gd
```diff
@@ -77,3 +78,9 @@ func _on_Seal_pressed():
 
 func stop():
 	client.stop()
+
+func _on_RPC_pressed():
+	rpc("update_color", Color(rand_range(0, 1), rand_range(0, 1), rand_range(0, 1), 1))
+
+remotesync func update_color(color: Color):
```
aaronfranke, line 85: Per the GDScript style guide, there should be two empty lines between functions.

```suggestion
	client.stop()


func _on_RPC_pressed():
	rpc("update_color", Color(rand_range(0, 1), rand_range(0, 1), rand_range(0, 1), 1))


remotesync func update_color(color: Color):
```

networking/webrtc_signaling/demo/client_ui.gd
```diff
@@ -28,6 +28,7 @@ func _disconnected():
 
 
 func _lobby_joined(lobby):
+	get_tree().network_peer = client.rtc_mp
```
Faless, line 31: The `SceneTree` can only have on network peer, so this will cause the demo to fail when using multiple clients.

Faless (commented):
There are at least 2 reasons why this doesn't work:
- There can only be one `network_peer` set to `get_tree()`, so this will only allow one peer per running instance.
- If you use the high level multiplayer API, you cannot keep the manual `rtc_mp.get_packet`.

aaronfranke:
Closing as abandoned.

PR #670, opened 2021-12-07
feat: update websocket-minimal to gdscript2.0

jonbonazza:
Update the networking/websocket-minimal project to use gdscript 2.0.

### DIFF
--- a/networking/websocket_minimal/Main.tscn
+++ b/networking/websocket_minimal/Main.tscn
@@ -1,12 +1,12 @@
-[gd_scene load_steps=3 format=2]
+[gd_scene load_steps=3 format=3 uid="uid://cxaa046x45suv"]
 
-[ext_resource path="res://server.gd" type="Script" id=1]
-[ext_resource path="res://client.gd" type="Script" id=2]
+[ext_resource type="Script" path="res://server.gd" id="1"]
+[ext_resource type="Script" path="res://client.gd" id="2"]
 
 [node name="Main" type="Node"]
 
 [node name="Server" type="Node" parent="."]
-script = ExtResource( 1 )
+script = ExtResource( "1" )
 
 [node name="Client" type="Node" parent="."]
-script = ExtResource( 2 )
+script = ExtResource( "2" )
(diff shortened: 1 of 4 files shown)


jonbonazza:
I wasn't sure if I was supposed to submit this PR to 4.0-dev or master. 🤷‍♂️

Calinou:
> I wasn't sure if I was supposed to submit this PR to 4.0-dev or master. man_shrugging

The `master` branch tracks Godot 3.x, while `4.0-dev` tracks the main repository's `master` branch.

I suppose could mimic the main repository structure in the demo projects repository, but there may be a reason as to why things are done this way currently. cc @aaronfranke

aaronfranke:
@Calinou It's because most people checking out the demos right now are using 3.x, so 4.0-dev is a separate branch.

jonbonazza:
I don't think it's an issue, I just wanted to make sure I made the PR to the right branch. :)

jonbonazza:
NOTE: the other networking demos still need updated, but I was using this one to test something, so I needed to update it anyway. I can probably go through the other ones in the next couple days and either add new commits to this PR for each one, or create separate PRs for each one. Whichever is more desired.

Also, we should probably get some `HTTPRequest` demos in there too at some point.

aaronfranke:
@jonbonazza I need to go through the demos and run the conversion script I made, and then the rest can be fixed up manually. I suggest not working on this further until this is done so that you don't spend time doing extra work.

jonbonazza:
Roger that.

akien-mga:
> @Calinou It's because most people checking out the demos right now are using 3.x, so 4.0-dev is a separate branch.

I mentioned it in the past but I still think that it would be better to have the same structure here and in the engine repository. It's confusing to have a `master` branch which doesn't correspond to Godot's `master` branch.

Instead, what you need for this case is simply to configure the would-be `3.x` branch as the default branch, so that's what gets checked out when people clone the repo. `master` is a branch like any other, it just happens to be the one configured to be default in most repos, but this can be changed.

aaronfranke:
Thanks!

PR #681, opened 2022-02-13
Don't export global illumination demo to HTML5

aaronfranke:
It freezes Firefox when I open it.

This PR also fixes the list items not ending in `</li>`


Calinou:
> This PR also fixes the list items not ending in </li>

The HTML5 specification technically allows this, but I agree it's better to use the traditional syntax.

Calinou (approved):

PR #683, opened 2022-02-17
Fix broken links for the your first 2D game tutorial

rubynho:
The link was sending to a not found page because the `Your first game` tutorial was moved to the `Your first 2D game` one. Also fix the tutorial name in the Dodge the Creeps project config file.

### DIFF
--- a/2d/dodge_the_creeps/README.md
+++ b/2d/dodge_the_creeps/README.md
@@ -4,7 +4,7 @@ This is a simple game where your character must move
 and avoid the enemies for as long as possible.
 
 This is a finished version of the game featured in the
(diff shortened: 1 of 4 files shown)


Calinou:
I think the URLs should be modified to point to the `/stable` version instead. `/latest` contains documentation that only pertains for Godot 4.0, but this repository's `master` branch tracks Godot 3.X.

aaronfranke (approved):
> but this repository's master branch tracks Godot 3.x.

Possibly not for long, once I get around to porting the demos to Godot 4... which hopefully won't be too far in the future.

PR #678, opened 2022-01-23
Added Alpha effect to joystick indicators

ScorpionInc:
Added alpha effect to joystick sprites via selfmodulator color on the canvasitem.

PR #684, opened 2022-02-28
Add axis label highlighting to the Joypads demo

ScorpionInc:
Highlights text labels for each respective axis in the same color/way as the buttons.
Allows easier viewing of the exact moment when the analog input goes over the activation threshold.
(Because small values would still be nearly transparent while being active.)

*Bugsquad edit: Follow-up to https://github.com/godotengine/godot-demo-projects/pull/678.*


Calinou:
This seems to work well, but I'm concerned about the text color being too dark when not pressed (which is bad for accessibility):

https://user-images.githubusercontent.com/180032/159286208-630f74e6-60cd-445e-8cfe-c200e393978c.mp4

Could you make the "unpressed" text slightly brighter? Alternatively, you can use the same color as the "pressed" text but with lower opacity (50% or 60%).

ScorpionInc:
Absolutely :) I used those colors only because it was the colors used by the button indicators on the bottom right. By changing the value of FONT_COLOR_DEFAULT I can effect both dynamically highlighted text groups at once. How much lighter would be desired? I could take a "shot in the dark" so to speak.

aaronfranke:
@ScorpionInc We could also make the buttons on the bottom-right brighter too.

ScorpionInc:
Light purple feels very pink to me. Had to double check that I had the blue channel high. It is easier to see but not partial to the color still. I'm open to suggestions, perhaps a bright green like the text at the top? Or maybe its fine and I've just been staring at a color wheel to long ;p

ScorpionInc:
temp
How about this coloring? Feels less pink but it's also not as brightly colored as the previous commit.

Calinou:
> How about this coloring? Feels less pink but it's also not as brightly colored as the previous commit.

Unfortunately, it's still pretty hard to read.

I'd suggest using `Color(1, 1, 1)` for pressed axes and `Color(1, 1, 1, 0.5)` for non-pressed axes. Simple but effective :slightly_smiling_face:

ScorpionInc:
Temp

misc/joypads/joypads.gd
```diff
@@ -8,6 +8,8 @@ extends Control
 # Licensed under the MIT license
 
 const DEADZONE = 0.2
+const FONT_COLOR_DEFAULT = Color(1.0, 1.0, 1.0, 0.5)
+const FONT_COLOR_ACTIVE  = Color.white
```
Calinou, line 12:
```suggestion
const FONT_COLOR_ACTIVE = Color.white
```

misc/joypads/joypads.gd
```diff
@@ -72,15 +74,19 @@ func _process(_delta):
 				joypad_buttons.get_child(JOY_ANALOG_R2).show()
 				# Transparent white modulate, non-alpha color channels are not changed here.
 				joypad_buttons.get_child(JOY_ANALOG_R2).self_modulate.a = scaled_alpha_value
+		# Highlight axis labels that are within the "active" value range. Simular to the button highlighting for loop below.
+		axes.get_node("Axis" + str(axis) + "/Label").add_color_override("font_color", FONT_COLOR_DEFAULT)
+		if abs(axis_value) >= DEADZONE:
+			axes.get_node("Axis" + str(axis) + "/Label").add_color_override("font_color", FONT_COLOR_ACTIVE)
```
Calinou, line 80:
```suggestion

		# Highlight axis labels that are within the "active" value range. Simular to the button highlighting for loop below.
		axes.get_node("Axis" + str(axis) + "/Label").add_color_override("font_color", FONT_COLOR_DEFAULT)
		if abs(axis_value) >= DEADZONE:
			axes.get_node("Axis" + str(axis) + "/Label").add_color_override("font_color", FONT_COLOR_ACTIVE)
```

aaronfranke (approved):
This could use a squash, but aside from that this looks good to me.

PR #687, opened 2022-03-13
Dodge the Creeps: Set the physics parameters of the mob before adding it to the scene tree

rburing:
This follows the recommendation from https://github.com/godotengine/godot/issues/45638.


Calinou (approved):
Thanks!

aaronfranke (approved):

PR #688, opened 2022-03-25
[3.x] Adds a deadzone control with indicator sprite

ScorpionInc:
Adds a deadzone spinbox to control the percentage strength on the axis deadzone as well as adding two sprites and a label for representing the deadzone value.

Thought it maybe a useful feature to allow people to change the deadzone value to see what it would look like at different values. I'd prefer to be able to do a "by device" dictionary lookup so that the demo or a game using the demo as reference could have a specific deadzone for each of their controllers remembered by a unique ID. This may already exist and I just don't know about it, but I figured a simple control shared for all joypad devices would suffice.

As always if this is an undesired addition or I messed up something please do let me know. Hope everyone is having a good day out there. o/

### DIFF
--- a/misc/joypads/deadzone.gd
+++ b/misc/joypads/deadzone.gd
@@ -1,0 +1,18 @@
+extends HBoxContainer
+
+const MAX_DEADZONE_SPRITE_SCALE:float = 1.5
+const PATH_TO_MASK:String = "Axis_Sprite/Mask"
+const PATH_TO_ROOT:String = "../../"
+
+var deadzone_sprite:Sprite = null
+var root_node:Node = null
+
+func _ready():
+	self.deadzone_sprite = self.get_node(self.PATH_TO_MASK)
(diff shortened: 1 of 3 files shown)


ScorpionInc:
Temp
Perhaps not the prettiest sprites, I don't like the white box around the joystick, but it requires no extra image files by reusing the already available images for textures.

ScorpionInc:
To know the input won't be activated until it is over the threshold yes, is trivial. This would simply make it easier to play with what the deadzone threshold is in-game when trying out axis input. At least that was the thought.

Calinou:
The demo has changed a fair bit in 4.x, so this would likely have to be redone from scratch on `master`.

I like the idea in principle, but I think this makes more sense for buttons than axes for the reasons @aaronfranke described.

Calinou:
Closing per the above comments. Thanks for the contribution nonetheless!

PR #702, opened 2022-03-31
Prepared Autoload(Singletons) demo for Godot 4

voylin:
Not really any fixing were required, some data changed between versions
when opening and testing everything. Did some minor changes to the project
settings so the new Godot 4 terms are used.

Everything works fine without problems.

### DIFF
--- a/loading/autoload/README.md
+++ b/loading/autoload/README.md
@@ -4,7 +4,7 @@ This demo shows how to use autoloads to change between scenes.
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan Mobile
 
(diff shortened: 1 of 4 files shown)


loading/autoload/README.md
```diff
@@ -4,7 +4,7 @@ This demo shows how to use autoloads to change between scenes.
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan
```
aaronfranke, line 7: If we're updating this text, we should be specifying if it's Vulkan Clustered or Vulkan Mobile.

voylin: Thanks! Fixed, squashed and pushed!

aaronfranke (approved):
Thanks!

PR #705, opened 2022-03-31
Fixes layout, nodes and theming for Control Gallery Demo - 4.0 dev

voylin:
A lot of nodes have changed which made this demo not work properly
anymore, nor display everything properly anymore. Everything is looking
like it did before and all buttons are working again.

There is however this slight bug in the engine right now which makes it
not possible to press menu buttons or option buttons if the window is
not the same as the content_scale_size of the main window. (By example
when you resize the window)
This is because the hitboxes for those buttons don't get notified of the
change in screen size which makes them stay in the same place where they
would be when resizing the window.

For the rest everything is working perfectly now, everything also has
the new Godot 4 default theme style.

Ps.: I followed yesterday's advice to "squash" commits together. I've tried, not sure if it worked, if it didn't please let me know. ^^"


gui/control_gallery/README.md
```diff
@@ -11,7 +11,7 @@ space between panels to resize them.
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan
```
aaronfranke, line 14: If we're updating this text, we should be specifying if it's Vulkan Clustered or Vulkan Mobile.

aaronfranke (commented):
Most of this looks good.

gui/control_gallery/README.md
voylin: Thanks for the info, I just changed the setting to Vulkan Mobile and made the change in the README.md file.

aaronfranke (approved):
Thanks!

Issue #697, opened 2022-03-28
[TRACKER] Porting the demos to Godot 4


aaronfranke:
This issue tracks the progress of porting the Godot demo projects to the current master branch of Godot Engine (what will eventually be Godot 4.0). All of this work is done on the [`4.0-dev`](https://github.com/godotengine/godot-demo-projects/tree/4.0-dev) branch as to not confuse 3.x users who want to download the demos. The `4.0-dev` branch will be merged into [`master`](https://github.com/godotengine/godot-demo-projects/tree/master) in the future, likely in the coming weeks. This list is in alphabetical order by folder path.

* ✅ and a checked box means that the demo is fully ported and working with absolutely 0 bugs. Even a false warning will prevent a demo from getting a ✅.
* 🚧 means that the demo has only had [the script](https://gist.github.com/aaronfranke/79b424226475d277d0035b7835b09c5f) run on it and still needs to be ported manually.
* 🐛 means that there are engine (not project) bugs that show up in this project. Projects that have bugs but the cause is unknown don't get 🐛.
* 🐞 means that there are minor bugs such as errors printing in the console.
* 🌋 means that there is a bug with rendering (possibly a problem with the project).
* 📜 means that there is a bug with scripting (can be GDScript/C#/VisualScript).
* 🖽 means that this project uses TileMap. 3.x tile data doesn't work in master.
* 💥 means that opening the editor with this project causes the engine to crash.

- [ ] 🚧⚛️ Bullet Shower
- [ ] 🚧🌋 Dodge the Creeps
- [ ] 🚧📜 Hierarchical Finite State Machine
- [ ] 🚧📜 GD Paint
- [ ] 🚧🌋 HDR for 2D
- [ ] 🚧🖽 Hexagonal Game
- [ ] 🚧 Scene Instancing Demo
- [ ] 🚧🖽 Isometric Game
- [ ] 🚧🖽 Kinematic Character 2D
- [ ] 🚧 2D Lights as Mask
- [ ] 🚧 2D Lights and Shadows
- [ ] 🚧 Navigation Polygon 2D
- [ ] 🚧🖽 Grid-based Pathfinding with Astar
- [ ] 🚧 2D Particles
- [ ] 🚧🖽 Physics-Based Platformer 2D
- [ ] 🚧📜 2D Physics Tests
- [ ] 🚧🖽 Platformer 2D
- [ ] 🚧 Pong with GDScript
- [ ] 🚧🖽 JRPG Demo
[...]

PR #706, opened 2022-03-31
Tested and fixed Scene Instancing Demo for 4.0-dev

voylin:
Everything works! Project settings are updated to Godot 4 new standards.
Another task which can be checked of #697.

### DIFF
--- a/2d/instancing/README.md
+++ b/2d/instancing/README.md
@@ -5,7 +5,7 @@ make many duplicates of the same object.
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan Mobile
 
 Check out this demo on the asset library: https://godotengine.org/asset-library/asset/148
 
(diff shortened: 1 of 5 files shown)


2d/instancing/README.md
```diff
@@ -5,7 +5,7 @@ make many duplicates of the same object.
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan
```
aaronfranke, line 8: If we're updating this text, we should be specifying if it's Vulkan Clustered or Vulkan Mobile.

2d/instancing/project.godot
```diff
@@ -15,25 +15,13 @@ config/description="A demo showing how to use scene instancing to
 make many duplicates of the same object."
 run/main_scene="res://scene_instancing.tscn"
 config/icon="res://icon.png"
+config/features=PackedStringArray("4.0")
 
 [display]
 
-window/size/width=800
```
aaronfranke, line 21: We should have `window/size/viewport_width=800`

voylin: Sorry about that. Fixed, squashed and pushed. ^^

aaronfranke (approved):

PR #707, opened 2022-03-31
Tested and Fixed Pong with GDScript Demo for 4.0-dev

voylin:
Made the usual changes to the project settings and tested to see if
everything works.

Everything works fine!
Another task which can be checked of #697.

### DIFF
--- a/2d/pong/README.md
+++ b/2d/pong/README.md
@@ -6,7 +6,7 @@ for game development in Godot, including
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan Mobile
 
 Note: There is a C# version available [here](https://github.com/godotengine/godot-demo-projects/tree/master/mono/pong).
 
--- a/2d/pong/project.godot
+++ b/2d/pong/project.godot
@@ -6,7 +6,7 @@
 ;   [section] ; section goes between []
 ;   param=value ; assign values to parameters
 
-config_version=4
+config_version=5
 
 [application]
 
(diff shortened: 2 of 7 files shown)


2d/pong/README.md
```diff
@@ -6,7 +6,7 @@ for game development in Godot, including
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan
```
aaronfranke, line 9: If we're updating this text, we should be specifying if it's Vulkan Clustered or Vulkan Mobile.

2d/pong/project.godot
```diff
@@ -15,19 +15,14 @@ config/description="A simple Pong game. This demo shows best practices
...
 run/main_scene="pong.tscn"
 config/icon="res://icon.png"
+config/features=PackedStringArray("4.0")
 
 [display]
 
+window/stretch/mode="canvas_items"
+window/stretch/aspect="expand"
 window/size/width=640
 window/size/height=400
```
aaronfranke, line 25:
```suggestion
window/size/viewport_width=640
window/size/viewport_height=400
```

aaronfranke (commented):
Most of this looks good.

2d/pong/project.godot
voylin: Thanks! This was something I missed in the settings, changed the settings, squashed and pushed.

aaronfranke (approved):
Thanks!

PR #708, opened 2022-03-31
Fixed Script Errors for GUI Input Mapping Demo - 4.0-dev

voylin:
Edited the project settings to match the new Godot 4 standard and fixed
the script errors which were making it impossible to run this demo.

Now everything works properly for the 4.0-dev version.
Another check that can be added for #697.

I really hope that I'm doing a good job helping to make all these demo's run properly on Godot 4,
if I'm doing anything wrong, please say so I can improve myself.

### DIFF
--- a/gui/input_mapping/ActionRemapButton.gd
+++ b/gui/input_mapping/ActionRemapButton.gd
@@ -8,9 +8,9 @@ func _ready():
 	display_current_key()
 
 
-func _toggled(button_pressed):
-	set_process_unhandled_key_input(button_pressed)
-	if button_pressed:
+func _toggled(is_button_pressed):
+	set_process_unhandled_key_input(is_button_pressed)
+	if is_button_pressed:
 		text = "... Key"
 		release_focus()
 	else:
@@ -21,7 +21,7 @@ func _unhandled_key_input(event):
 	# Note that you can use the _input callback instead, especially if
 	# you want to work with gamepads.
 	remap_action_to(event)
-	pressed = false
+	button_pressed = false
 
 
 func remap_action_to(event):
@@ -31,5 +31,5 @@ func remap_action_to(event):
 
 
 func display_current_key():
-	var current_key = InputMap.get_action_list(action)[0].as_text()
+	var current_key = InputMap.action_get_events(action)[0].as_text()
 	text = "%s Key" % current_key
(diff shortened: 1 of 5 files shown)


gui/input_mapping/ActionRemapButton.gd
```diff
@@ -21,7 +21,7 @@ func _unhandled_key_input(event):
 	# Note that you can use the _input callback instead, especially if
 	# you want to work with gamepads.
 	remap_action_to(event)
-	pressed = false
+	button_pressed = false
```
aaronfranke, line 24: This line is good, but we're not done here. We should rename the parameter variable on lines 11-13 to something else to avoid a conflict with the `pressed` -> `button_pressed` rename in Godot. Lines 11-13 shouldn't be any of `pressed` or `button_pressed` or `is_pressed` because all of those names conflict, we have to pick something else. Maybe `is_button_pressed`?

aaronfranke (commented):
Most of this looks good.

gui/input_mapping/ActionRemapButton.gd
voylin: Thank you for the advice, I made the changes to the files and the argument is called is_button_pressed. I could not come up with a better name for it. ^^"

gui/input_mapping/project.godot
```diff
@@ -18,13 +18,11 @@ config/description="A demo showing how to build an input key remapping screen.
   after the project is restarted."
 run/main_scene="res://InputRemapMenu.tscn"
 config/icon="res://icon.png"
+config/features=PackedStringArray("4.0")
 
 [display]
 
-window/size/width=640
-window/size/height=480
```
aaronfranke, line 25: We should keep these lines, but they should be `viewport_width` instead of `width` and `viewport_height` instead of `height`.

voylin: A question about this, should we always use 640x480 for viewport size? Because the defaults in 4.0 became bigger.

aaronfranke: The default in both 3 and 4 is 1024x600. Several of these demos have a smaller size because we just don't need a lot of space for such a simple demo. And no, we should definitely not always use 640x480. Just any demo that is already 640x480 in Godot 3 should still be that resolution in Godot 4.

aaronfranke (approved):
Thanks!

PR #709, opened 2022-04-02
Added format, mix_rate and stereo settings for Mic Record Demo - 4.0-dev

### LINKED ISSUE #470: Setting of format, mix_rate and stereo flag on mic_record
godot-demo-projects/audio/mic_record

**Description:**
The demo and [instructions](https://docs.godotengine.org/en/stable/tutorials/audio/recording_with_microphone.html?highlight=record) show how to print out the format, mix_rate and stereo flags, but not how to set any of them.  

The [AudioStreamSample](https://docs.godotengine.org/en/stable/classes/class_audiostreamsample.html#class-audiostreamsample) class has a format option FORMAT_IMA_ADPCM, which would compress the signal.  Even being able to set the stereo flag to false would save 50% of the bandwidth, as it's generally not relevant when you have a single microphone and therefore should be part of the example.

(If there is no way to set these values, maybe this issue should be filed with the core system.)

voylin:
Fixes #470


voylin:
@aaronfranke I'm sorry, something went kind of wrong with this pull request and whilst trying to fix it I probably made it worse :s

Calinou:
Superseded by https://github.com/godotengine/godot-demo-projects/pull/711.

PR #713, opened 2022-04-03
Add a 3D settings demo (Graphical settings) for 4.0-dev

voylin:
Could possible fix #600.
This demo has some of the basic graphical settings that game devs would want. I tried to keep it minimal because it's just a demo. I hope this is sufficient and that there are enough comments explaining everything. Everything works without any bugs or errors.

PR #719, opened 2022-04-08
Improve the 3D graphics settings demo

Calinou:
Follow-up to https://github.com/godotengine/godot-demo-projects/pull/713.

- New graphics settings:
  - Field of view.
  - Volumetric fog.
- Use Godot's built-in 3D viewport facilities instead of using a SubViewport.
  - This also allows for choosing AMD FidelityFX Super Resolution for resolution scaling, but it's not working yet due to a Godot bug.
  - Nearest-neighbor scaling is no longer available. It can be reimplemented in Godot core and then exposed in the future.
- Separate MSAA from FXAA settings, as both can be used at the same time.
- Add more quality settings for various effect settings.
- Rename sections and setting names to be more "player-oriented" (even if this results in the actual Godot terminology not being presented).
- Add performance hints to setting values.
- Add FPS counter and viewport resolution display (taking the resolution scale into account).
- Add more objects to the test scene and tweak environment settings.
  - ACES tonemapping is now used.
  - Debanding is now enabled, as this scene requires it to avoid visible banding.
  - Make the emissive box less bright to prevent MSAA and FXAA from being ineffective on it (due to [Godot not supporting correct HDR antialiasing](https://github.com/godotengine/godot/issues/59965)).
  - Remove unnecessary DirectionalLight3D, VoxelGI and ReflectionProbe.

There's more stuff that could be done in the future (such as adding shadow quality settings), but this is a good start already.


3d/graphics_settings/settings.gd
```diff
@@ -1,25 +1,37 @@
...
+@onready var camera := $Node3D/Camera3D
+@onready var fps_label := $FPSLabel
+@onready var resolution_label := $ResolutionLabel
 
-var current_quality := 2 # This is needed for when screen size changes
-var viewport_start_size := Vector2.ZERO
+# When the screen changes size, we need to update the 3D
+# viewport quality setting. If we don't do this, the viewport will take
+# the size from the main viewport.
+var viewport_start_size := Vector2(ProjectSettings.get_setting(&"display/window/size/viewport_width"), ProjectSettings.get_setting(&"display/window/size/viewport_height"))
```
aaronfranke, line 14: Doesn't this need to be `@onready var` if we're putting the ProjectSettings logic here?

However, I think it looks better the way it was, with 2 lines in `_ready()`. This line is super long.

3d/graphics_settings/settings.gd
```diff
@@ -1,25 +1,37 @@
...
+func _process(delta: float) -> void:
+	fps_label.text = "%d FPS (%.2f mspf)" % [Engine.get_frames_per_second(), 1000.0 / Engine.get_frames_per_second()]
 
 
+func update_resolution_label() -> void:
+	var viewport_render_size = get_viewport().size * get_viewport().scaling_3d_scale
+	resolution_label.text = "3D viewport resolution: %d × %d (%d%%)" \
+			% [viewport_render_size.x, viewport_render_size.y, round(get_viewport().scaling_3d_scale * 100)]
+
 func _on_HideShowButton_toggled(show_settings: bool) -> void:
```
aaronfranke, line 35:
```suggestion
			% [viewport_render_size.x, viewport_render_size.y, round(get_viewport().scaling_3d_scale * 100)]


func _on_HideShowButton_toggled(show_settings: bool) -> void:
```

GDScript style guide, two lines between functions.

3d/graphics_settings/settings.gd
```diff
@@ -132,49 +135,85 @@ func _on_ss_reflections_option_button_item_selected(index: int) -> void:
 		world_environment.environment.set_ssr_max_steps(32)
 	elif index == 3: # High
 		world_environment.environment.set_ssr_enabled(true)
-		world_environment.environment.set_ssr_max_steps(64)
-	# We set the fade in higher (default is 0.15) so it looks
-	# a better and cleaner.
-	world_environment.environment.set_ssr_fade_in(0.7)
+		world_environment.environment.set_ssr_max_steps(56)
 
 func _on_ssao_option_button_item_selected(index: int) -> void:
```
aaronfranke, line 140:
```suggestion
		world_environment.environment.set_ssr_max_steps(56)


func _on_ssao_option_button_item_selected(index: int) -> void:
```

Looks like this was missing a line here before too, but we can make it better in this PR.

3d/graphics_settings/settings.gd
Calinou: It doesn't need to use `@onready`, as ProjectSettings is a singleton which is initialized before any other nodes.

I improved formatting for readability.

PR #720, opened 2022-04-12
Improve visual appearance in the 2D Sprite Shaders demo

Calinou:
- Scale the sprites up so they take a larger portion of the window.
- Enable mipmaps on the sprite to avoid artifacts at low resolutions.


aaronfranke:
I'm not so sure about the background color. I think it's nicer to have more contrast with the background.

aaronfranke (approved):
I updated the PR with the old background color and cleaned up the transforms.

PR #732, opened 2022-05-02
Making Demo Dodge the Creeps ready for 4.0-dev

voylin:
Going through the demo's mentioned at: #697
to fix and prepare them for 4.0.
It was mentioned that there were minor bugs (errors printing in console), found non of that, only some logs about awaits not being needed (they were needed). Everything works again now and stretch_mode set to canvas_items.

### DIFF
--- a/2d/dodge_the_creeps/README.md
+++ b/2d/dodge_the_creeps/README.md
@@ -10,7 +10,7 @@ consider following the tutorial in the documentation.
 
 Language: GDScript
 
-Renderer: GLES 3 (particles are not available in GLES 2)
+Renderer: Vulkan Mobile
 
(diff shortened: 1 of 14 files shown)


2d/dodge_the_creeps/README.md
```diff
@@ -10,7 +10,7 @@ consider following the tutorial in the documentation.
 
 Language: GDScript
 
-Renderer: GLES 3 (particles are not available in GLES 2)
+Renderer: Vulkan Clustered
```
aaronfranke, line 13: We should probably switch this demo to Vulkan Mobile, it seems to look the same in my testing.

voylin: Change has been made.

aaronfranke (approved):

PR #734, opened 2022-05-02
GD Paint Demo Update to 4.0

voylin:
Part of #697
Demo is working now in 4.0.
There is only one slight issue, but that has to do with the engine viewport system I think. When saving the picture, the framing isn't always correct, really depends how you scaled the window. This could also be a me issue as I use i3 on linux (Tiling window manager).

But otherwise, it's working perfectly.

### DIFF
--- a/2d/gd_paint/README.md
+++ b/2d/gd_paint/README.md
@@ -6,7 +6,7 @@ and eraser, as well as a rectangle and a circle brush.
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan Mobile
 
 Check out this demo on the asset library: https://godotengine.org/asset-library/asset/517
 
--- a/2d/gd_paint/icon.png.import
+++ b/2d/gd_paint/icon.png.import
@@ -1,8 +1,9 @@
 [remap]
 
 importer="texture"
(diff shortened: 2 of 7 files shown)


aaronfranke (approved):

aaronfranke:
> There is only one slight issue, but that has to do with the engine viewport system I think. When saving the picture, the framing isn't always correct, really depends how you scaled the window. This could also be a me issue as I use i3 on linux (Tiling window manager).

I can reproduce this issue, and that's not the only one. I'm also having an issue where the brush strokes appear on top of the UI.

ff

voylin:
> > There is only one slight issue, but that has to do with the engine viewport system I think. When saving the picture, the framing isn't always correct, really depends how you scaled the window. This could also be a me issue as I use i3 on linux (Tiling window manager).
> 
> I can reproduce this issue, and that's not the only one. I'm also having an issue where the brush strokes appear on top of the UI.
> 
> ff

I'll try fixing it. Not sure if an engine issue or not. I don't have the drawing over the UI issue strangely enough.

voylin:
@aaronfranke Fixed the issues in #741. Saving also works now after trying to figure out what was wrong with the window.

PR #737, opened 2022-05-04
Window Management Update 4.0

voylin:
Updated to 4.0 everything should work.

Could not test minimize or maximize as those features are not available for my tiling window manager (i3) .

PR #740, opened 2022-05-07
Fix window management demo for 4.0

voylin:
Fixing the problem which @aaronfranke mentioned in PR #737
Some functions are not present inside of Godot 4.0 so for the time being I commented them out as this demo will probably need more changes by the end of the alpha.

### DIFF
--- a/misc/window_management/control.gd
+++ b/misc/window_management/control.gd
@@ -11,7 +11,7 @@ func _ready():
 
 
 func _physics_process(_delta):
-	var modetext = "Mode:\n"
+	var modetext = "Mode: "
 	if DisplayServer.window_get_mode() == DisplayServer.WINDOW_MODE_FULLSCREEN:
 		modetext += "Fullscreen\n"
 	else:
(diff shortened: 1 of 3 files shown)


aaronfranke:
The text in Godot 4 seems to be a lot larger by default, especially the size of the newline. The text on the right isn't rendering quite right, for example the value of "Size" is overlapping the "Mode". We should consider replacing the `\n` and have this info be inline at least for the properties with short names ("Screen0 Resolution:" probably stlll deserves two lines).

I think this is a Godot bug, but I'm getting this error from line 47 of `control.gd`:

```
E 0:00:01:0122   _physics_process: Condition "!windows.has(p_window)" is true. Returning: Size2i()
  <C++ Source>   platform/linuxbsd/display_server_x11.cpp:1719 @ window_get_size()
  <Stack Trace>  control.gd:47 @ _physics_process()
```

voylin:
> The text in Godot 4 seems to be a lot larger by default, especially the size of the newline. The text on the right isn't rendering quite right, for example the value of "Size" is overlapping the "Mode". We should consider replacing the `\n` and have this info be inline at least for the properties with short names ("Screen0 Resolution:" probably stlll deserves two lines).
> 
> I think this is a Godot bug, but I'm getting this error from line 47 of `control.gd`:
> 
> ```
> E 0:00:01:0122   _physics_process: Condition "!windows.has(p_window)" is true. Returning: Size2i()
>   <C++ Source>   platform/linuxbsd/display_server_x11.cpp:1719 @ window_get_size()
>   <Stack Trace>  control.gd:47 @ _physics_process()
> ```

I'm not seeing that error. But I fixed the UI issues by removig some of the '\n' and also by using VBoxContainers.

aaronfranke:
The buttons seem too close to the edge, and we can improve the anchor points to make the UI layout cleaner:

Screenshot from 2022-05-07 21-19-57
Screenshot from 2022-05-07 21-18-33

voylin:
> The buttons seem too close to the edge, and we can improve the anchor points to make the UI layout cleaner:

Really looks better like that indeed, added the padding.

aaronfranke (commented):
Looks good, but please update the commit message to something more descriptive than "Fix".

voylin:
> Looks good, but please update the commit message to something more descriptive than "Fix".

'Fix' is fixed. ;)

aaronfranke (approved):

PR #741, opened 2022-05-07
Fixing GD Paint Issues in 4.0-dev

voylin:
The clipping happened because the panel was see-through.
The saving is also fixed by changing some of the project settings.
To fix the issues from #734

### DIFF
--- a/2d/gd_paint/paint_root.tscn
+++ b/2d/gd_paint/paint_root.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=5 format=3 uid="uid://bhjmyer4wlwy2"]
+[gd_scene load_steps=6 format=3 uid="uid://bhjmyer4wlwy2"]
 
 [ext_resource type="Script" path="res://paint_control.gd" id="1"]
 [ext_resource type="Script" path="res://tools_panel.gd" id="2"]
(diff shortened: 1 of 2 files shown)


2d/gd_paint/paint_root.tscn
```diff
@@ -7,17 +7,22 @@
 [sub_resource type="StyleBoxFlat" id="StyleBoxFlat_4ksjc"]
 bg_color = Color(1, 1, 1, 1)
 
+[sub_resource type="StyleBoxFlat" id="StyleBoxFlat_npkcn"]
+bg_color = Color(0.254902, 0.286275, 0.294118, 1)
```
aaronfranke, line 11:
```suggestion
bg_color = Color(0.2, 0.2, 0.2, 1)
```

aaronfranke (commented):
Thanks for these fixes! It works well on my system. But the panel's background color seems a bit bright.

voylin:
> Thanks for these fixes! It works well on my system. But the panel's background color seems a bit bright.

You're welcome. I'm glad that I found out what was causing that strange saving bug.

aaronfranke (approved):

2d/gd_paint/tools_panel.gd
```diff
@@ -95,7 +95,7 @@ func brush_color_changed(color):
 
 func background_color_changed(color):
 	# Change the background color to whatever colorthe background color picker is.
-	get_parent().get_node(^"DrawingAreaBG").modulate = color
+	get_parent().find_child(^"DrawingAreaBG").modulate = color
```
aaronfranke, line 98: Why was this changed? This should still be `get_node`.

aaronfranke (changes requested):

2d/gd_paint/tools_panel.gd
voylin: Aah that got changed because I was testing with viewport containers instead. It was during the testing of SubViewports that I discovered the solution. Will change it right now!

aaronfranke (approved):

aaronfranke (approved):

aaronfranke:
Thanks for all of your work!

voylin:
> Thanks for all of your work!

I'm happy that I can help. \^^/

PR #743, opened 2022-05-18
Add a 3D antialiasing demo

Calinou:
This demo showcases the various 3D antialiasing techniques supported by Godot. Many potentially problematic cases for TAA are also included, so this will be used as a testing project for the purpose of improving Godot's TAA implementation.

## TODO (in future PRs)

- Add animated skinned mesh (skeleton) test case.
- Add animated skinned mesh (blend shape) test case.
- Add SoftBody test case.
- Add complex model test case to check for specular aliasing.


clayjohn:
We should keep our eyes open for a cc0 licenced model with a more complex profile. Something thin like a powerline is important to show how well different AA techniques preserve shape.

Calinou:
Rebased with TAA support, this should now be ready to be merged.

clayjohn:
Before merging we really need to use a better model. Cubes are not sufficient to clearly show the differences between AA methods. Consider using something like this https://polyhaven.com/a/dutch_ship_medium which has a good amount of thin geometry.

Calinou:
> Before merging we really need to use a better model. Cubes are not sufficient to clearly show the differences between AA methods. Consider using something like this [polyhaven.com/a/dutch_ship_medium](https://polyhaven.com/a/dutch_ship_medium) which has a good amount of thin geometry.

Done, I added this model in the test scene (with 1K textures to keep file sizes low).

clayjohn (commented):
Looks great to me! It was lots of fun to step through and it does a nice job of covering a wide range of edge cases. 

I will leave it up to someone with more knowledge about the demos to approve with respect to organization/code quality.

aaronfranke (approved):
Nice! Thank you, this demo is super polished.

I pushed some updates, including some minor changes to the camera script and default camera angle, an updated screenshot, renaming the scene and script, moving textures to a subfolder, and optimizing PNG files using `oxipng`.

PR #744, opened 2022-05-23
Add a text-to-speech demo

bruvzg:
Adds a demo for https://github.com/godotengine/godot/pull/56192


lesleyrs:
Hey you are really awesome for adding this to Godot!

I just played around with the demo for a bit and it's great for some use cases even outside of games.

Windows Narrator takes over your entire PC and doesn't allow you to use any keys because they're all keymapped to a billion things. So you can't multitask.

I think this is very powerful being able to effectively turn almost anything into an audiobook, offline, with your own key shortcuts.

This is for you: 👑 

A few problems though:

[This](https://github.com/godotengine/godot/pull/56192#issuecomment-1128259689) and it only detects en-us voice on Windows 10 even if I set the other language to be the default narrator (dutch).

Calinou (approved):
Thanks!

(I've pushed a second commit with cosmetic changes, but the behavior is the same.)

PR #747, opened 2022-06-09
Update loading in a thread demo for 4.0

jtnicholl:
This updates and improves the simple background loading demo (the one that doesn't use `load_interactive`). On the list in #697.
In addition to updating for 4.0, I made the following additional changes:

- Make the comments more clear and informative. I removed a few that I thought were just stating the obvious (like one after `return tex` that just says it returns something, or the one that says `ResourceLoader.load` loads a resource). Most of them I kept and just edited for clarity.
- Don't reuse `Thread` objects. According to the documentation, you shouldn't do this, so the demo now waits for the previous thread to finish (if there is one) and makes a new one.
- Add an `_exit_tree` method that makes sure the thread is finished, then frees it.
- Change the order of the methods in the script to match the order they are called, to make it easier to follow. They are now ordered as `_on_load_pressed`, `_bg_load`, `_bg_load_done`, `_exit_tree`.

Some of these changes are subjective, so let me know what you think. The last one is technically in violation of the GDScript style guide, since it puts a built-in virtual method after the private methods, but I think it improves readability enough to be worth it. I'm also not sure why the documentation says not to reuse thread objects, since doing so seems to work just fine.

### DIFF
--- a/loading/threads/README.md
+++ b/loading/threads/README.md
@@ -4,7 +4,7 @@ An example using a thread to load an image.
 
 Language: GDScript
 
-Renderer: GLES 2
+Renderer: Vulkan Mobile
 
 Check out this demo on the asset library: https://godotengine.org/asset-library/asset/144
 
--- a/loading/threads/mona.png.import
+++ b/loading/threads/mona.png.import
@@ -1,8 +1,9 @@
 [remap]
 
 importer="texture"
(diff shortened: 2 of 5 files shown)


aaronfranke (approved):
Looks good, thanks for this! And thanks for improving the engine too :)

PR #748, opened 2022-06-09
Update 3D navigation demo for 4.0

jtnicholl:
I have updated the 3D navigation demo for Godot 4.0. On the list in #697.
Since the new NavigationServer provides a lot of new features for dynamic obstacle avoidance and such, and this demo doesn't use them, I also renamed it from "3D Navigation Mesh" to "Basic 3D Navigation". I figured that an advanced 3D navigation demo could be made later to showcase these features. I also renamed the directory it's in from "navmesh" to "basic_navigation".
One other thing I did is I extracted the robot mesh, level mesh, and navmesh from the scene and saved them to their own .res files. They contain binary data so they shouldn't have been packed into a .tscn file like that. I also removed the unused AnimationPlayer node, I'm not sure why it was there.

Lastly, I ran into an issue where the robot won't cross one small section of the level. This doesn't happen in the 3.x version. I tried everything I could think of, but I couldn't figure out why it's happening. Here's a screenshot to demonstrate:
Screenshot_20220609_142934
Hopefully someone can figure out what's causing this. I think it might be an engine bug. I don't think it's so major that this can't still be merged for now, but it will have to be fixed later.


smix8:
@jtnicholl 
The pathfinding issue is due to the navmesh polygons being corrupted near the ramp. Godot also prints a region merge error at the scene start due to this. Clearing the baked navmesh and (re)baking solved the issue.

jtnicholl:
Not sure how I didn't notice that error message.
The error is caused by the settings used to generate the navmesh, clearing the navmesh before rebaking also clears all these settings back to the defaults. But the default settings won't work here, because it makes the ramp leading to the highest area up from the left uncrossable due to being too narrow.
The error says "This happens when the current `cell_size` is different from the one used to generate the navigation mesh" but this isn't true, the cell_size hasn't been changed after baking. After spending some time experimenting with the settings further, I managed to find a combination that makes the entire map navigable while also not printing the error message. It isn't ideal, though, it clips through the ground in a few small spots and also lifts above the ground quite a bit near the bases of some of the ramps, most noticeably the ramp that goes through the center ring. The cell_size should be lowered to fix this, but I couldn't find any combination of settings with a cell_size different from the default that worked.

smix8:
@jtnicholl 
cell_size = 0.1
agent_radius = 0.1
max_slope = 65

jtnicholl:
That's better, though it still lifts off the ground a bit near the bases of the ramps. I increased the agent_radius to 0.2 and that improved it a bit further.

aaronfranke (commented):
The scenes need to be opened and re-saved since Godot 4 now doesn't save extra spaces inside parenthesis.

> I figured that an advanced 3D navigation demo could be made later to showcase these features. I also renamed the directory it's in from "navmesh" to "basic_navigation".

I would prefer the folder name not have "basic" in it, unless/until there are multiple navigation demos. Even then, there's an argument in favor of having both "navigation" and "advanced_navigation" names.

aaronfranke (approved):

PR #754, opened 2022-07-10
Docs: 2D Platformer - Update Pause Mode Comments

HaywardMorihara:
I believe this comment got out-of-date (based on the fact that the `Stage` scene seems like something that was in an older version of the demo: https://github.com/godotengine/godot-demo-projects/blob/81441c42b79e38e54c2bc38acee3205f22431738/2d/platformer/Stage.tscn), so updated the comment to reflect the current state of the demo, along with some other information that I thought a developer might find helpful.

### DIFF
--- a/2d/platformer/src/Main/Game.gd
+++ b/2d/platformer/src/Main/Game.gd
@@ -25,10 +25,18 @@ func _unhandled_input(event):
 	if event.is_action_pressed("toggle_fullscreen"):
 		OS.window_fullscreen = not OS.window_fullscreen
 		get_tree().set_input_as_handled()
-	# The GlobalControls node, in the Stage scene, is set to process even
-	# when the game is paused, so this code keeps running.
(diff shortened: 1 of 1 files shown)


HaywardMorihara:
I'm confused about the GH Action failure (I don't see a message about what's wrong with the format of the comment?)

 I'm assuming it's because of the `*` in the comment? https://github.com/godotengine/godot-demo-projects/runs/7271893814?check_suite_focus=true#step:4:16 

Tried replacing the stars with a numbered list

Calinou:
> I'm confused about the GH Action failure (I don't see a message about what's wrong with the format of the comment?)

There appears to be trailing whitespace at the end of the life.

HaywardMorihara:
🤦‍♂️ my mistake, removed the whitespace!

aaronfranke (approved):
Makes sense to me.

PR #756, opened 2022-07-15
Add a 3D decals demo

Calinou:
This adds a demo showcasing various use cases for the Decal node.

Thanks @dzil123 for the help on run-time decal placement :slightly_smiling_face: 

## Preview

image


3d/decals/README.md
```diff
@@ -0,0 +1,23 @@
...
+- Filter modes with Mipmaps prevent decals from looking grainy at a distance, at
+  a small performance cost. When mipmaps are used without anisotropic filtering,
+  decals will look blurry when viewed at oblique angles.
+- Filter modes with Anisotropic don't look grainy at a distance and also avoid
+  looking blurry when viewed at oblique angles. However, filter modes with
+  Anisotropic have a greater performance cost than enabling Mipmaps alone.
+
+Language: GDScript
+
+Renderer: Vulkan
```
aaronfranke, line 19: Clustered?

aaronfranke (commented):
I'd appreciate some subfolders for this project, there are a lot of assets.

Calinou:
> I'd appreciate some subfolders for this project, there are a lot of assets.

Done. I also fixed the demo so it runs correctly on latest `master`.

I've noticed that clustering artifacts will pop up if you have decals without valid textures in the scene (and those decals will render with other valid decals' textures).

Calinou:
I've started adding code for run-time decal placement, but I'm having trouble getting the position to be correct (relative to cursor position and camera position/angle).

dzil123:
Do you mind if I take a stab at this?

Calinou:
> Do you mind if I take a stab at this?

Sure, go ahead :slightly_smiling_face:

Calinou:
I integrated the changes from https://github.com/godotengine/godot-demo-projects/pull/779 in this pull request.

aaronfranke (approved):
I haven't looked in depth but it looks good from a quick review.

3d/decals/textures/icon_outlined_normal.png.import
```diff
@@ -0,0 +1,34 @@
...
+compress/bptc_ldr=0
+compress/normal_map=0
+compress/channel_pack=0
+mipmaps/generate=false
+mipmaps/limit=-1
+roughness/mode=0
+roughness/src_normal=""
+process/fix_alpha_border=true
+process/premult_alpha=false
+process/normal_map_invert_y=false
```
jtnicholl, line 30: Some of these normal maps are DirectX-style, invert Y should be on.

Calinou: Normal mapping looks incorrect if **Normal Map Invert Y** is enabled on the decals' normal maps:

### Disabled

image

### Enabled

image

jtnicholl: Yeah, that's what I meant. All of the normal maps _except_ the scifi ones are DirectX style.

clayjohn:
I think this looks mostly good! If you still have some capacity to work on this I would put a bit more time into the water puddles. Right now they look a bit funny because the increase of roughness at the edges makes the edges look like they are glowing while the inside looks flat and non-reflective.

I think instead the roughness should be consistent within the puddle, but it should fade at the edges with an alpha blend. 

Also, the sun is behind the camera by default, so you do not see any reflection at all from the light in the puddle unless you move the camera. I think the camera should be moved to the opposite side of the puddles, or the light should be rotated around. 

Other than that, I think this is excellent. It does a great job at capturing some of the corner cases for decals.

Calinou:
> Right now they look a bit funny because the increase of roughness at the edges makes the edges look like they are glowing while the inside looks flat and non-reflective.

Unfortunately, this is an artifact of linear filtering that can't be fixed.

Calinou:
Rebased and tested again on 4.0.beta11, it works as expected.

PR #758, opened 2022-07-15
Improve 3D global illumination demo (4.0)

Calinou:
Follow-up to https://github.com/godotengine/godot-demo-projects/pull/718.

- Readd LightmapGI example.
- Allow toggling between the Once and Always update modes for the ReflectionProbe.
- Add a Decal node to act as a blob shadow for dynamic objects.
- Halve SSAO intensity for a better apperance in the demo's flat, bright surfaces.
- Halve ambient light sky contribution when using the disabled or lightmap-based GI modes. This prevents the level's shaded areas from looking too bright and blue (with lightmaps, this change only affects dynamic objects.)
- Enable FXAA and debanding for a better appearance.

PR #759, opened 2022-07-15
Improve 3D global illumination demo (3.x)

Calinou:
3.x version of https://github.com/godotengine/godot-demo-projects/pull/758.

- Add reflection probe mode toggle to use Always update mode (slower, but updates in real-time).
- Add SSAO toggle (with halved SSAO intensity to look better in the test scene).
- Tweak DirectionalLight shadow bias parameters to reduce shadow peter panning.
- Enable FXAA and debanding for a better appearance. Debanding is disabled on mobile due to known issues.


adamscott (approved):
Seems good to me.

Issue #650, opened 2021-09-27
Dynamic split screen also for 2D


ProfDrKrebs:
**Which demo project is affected:**
https://github.com/BenjaminNavarro/godot_dynamic_split_screen

**Description:**
Hello, is it possible to create the demo project: "dynamic split screen" also for 2D?
I tried it myself, but as a leisure programmer I quickly failed.
I'm not sure if my post is right here, but I would like to express my interest in a dynamic split screen for 2d.
Thank you all and greetings from Germany

voylin:
I'm not sure if it is really needed as it is very similar to the 3D version. If you would still like to have a 2D version and if the people with power over the godot-demo-projects think that this could be a good addition, then I would do this.

But not sure if the maintainers of Godot-demo-projects see any need in a 2D version. 
@aaronfranke Do you think it would be a good idea to also have a 2D version?

aaronfranke:
It should be easy for users to adapt the techniques in the 3D version to 2D, so I don't think we need an official demo for this.

However, do note that anyone is very welcome to make an unofficial demo and distribute it on the Godot Asset Library.

PR #762, opened 2022-08-03
[3.x] Add a 2D version of the dynamic split screen demo

BenjaminNavarro:
Here is an updated version of the dynamic split screen demo project that now includes a 2D implementation along the existing 3D one.

It took me some time to figure out a way to do this in 2D as cameras and viewports behave differently compared to 3D. ~In the 3D version there are only 2 additional viewports, one per camera. In 2D, I had to create a third viewport to hold all the game content and share its `2d_world` with the two other viewports. I don't know if there is a better way to do this.~

The shader is the same for both 2D and 3D but the scrips are all slightly different.

I also reworked the 3D scene so that it's more similar to the 2D one ~(due to its 3rd viewport limitation)~

See https://github.com/godotengine/godot-demo-projects/pull/762#issuecomment-1216254557

*Bugsquad edit: This closes https://github.com/godotengine/godot-demo-projects/issues/650.*


BenjaminNavarro:
After the offline suggestion from @NathanLovato to use only 2 viewports in 2D as well (didn't thought about it) I refactored the project a little to make it simpler and more consistent across the 2D and 3D versions.

I pushed a new commit so you can inspect both versions if you want, I'll squash them later after the reviews.

The main changes are:
 1. The node hierarchy is the same for both versions (except for `Node2D`/`Spatial` and `Camera2D`/`Camera` of course)
 2. I merged `2d/camera_controller.gd` and `3d/camera_controller.gd` into `camera_controller.gd`. I detect if the script runs for a 2D or 3D scene, set the `mode` enum and use it in a couple of places to handle the difference between the two modes. Since most of the code was the same I think it's better now considering future maintenance and updates but I can roll back to two different scripts if you think that would be clearer to users

aaronfranke:
This PR has a few conflicts and needs to be rebased.

In general a 2D version of this demo seems nice.

BenjaminNavarro:
Rebasing done on 3.x, nothing had to be changed

viewport/dynamic_split_screen/2d/player.gd
```diff
@@ -0,0 +1,16 @@
...
+export(int, 1, 2) var player_id = 1
+export(float) var walk_speed = 200.0
+
+
+func _physics_process(_delta):
+	var velocity = Vector2.ZERO
+	velocity.y = -Input.get_action_strength("move_up_player" + str(player_id))
+	velocity.y += Input.get_action_strength("move_down_player" + str(player_id))
+	velocity.x = -Input.get_action_strength("move_left_player" + str(player_id))
+	velocity.x +=  Input.get_action_strength("move_right_player" + str(player_id))
```
aaronfranke:
```suggestion
	velocity.x += Input.get_action_strength("move_right_player" + str(player_id))
```

aaronfranke (changes requested):
The 2D scene does not run for me in Godot 3.6.stable: Screenshot 2025-09-30 at 8 58 30 AM

BenjaminNavarro:
@aaronfranke Is there a chance that this gets merged eventually ? I provided the initial 3D implementation, opened this PR 3 years ago to provide a 2D alternative and did the rebase as requested a year and a half ago. I'm not sure I want to keep working on this if it gets nowhere

aaronfranke:
@BenjaminNavarro Apologies, it's my bad for not reviewing this sooner. It all depends on what you want to do with this. I'm looking through the demo projects PRs today and I'll merge this in if it's brought up to a good state. We don't actively maintain the 3.x branch but that also means I won't pose quite as high of a standard for this branch, if it works and has no obvious style issues I can merge this in.

BenjaminNavarro:
@aaronfranke Have you switched to GLES3 for your test? With default settings and a fresh build of godot 3.x things were still working fine for me.

Switching to GLES3 caused a black screen for the 1st player in the 2D demo due to an incorrect update mode on its viewport. This has been fixed with the latest commit.

Let me know if you still experience the same problem

PR #767, opened 2022-08-25
[4.0] Clean up and fix Truck Town

jtnicholl:
I've cleaned up the Truck Town demo and gotten it fully working.
I moved all of the mesh data out of the .tscn files into binary resource files. Except for the town itself, which I brought into Blender and did a clean up on; I fixed the broken normals that were messing up the lighting and also de-duplicated the meshes. It's now a .glb file.
I also did a bit of code cleanup, made the file paths lowercase, updated the screenshot and icon, and turned on FXAA.


3d/truck_town/vehicle.gd
```diff
@@ -3,19 +3,19 @@ extends VehicleBody3D
 const STEER_SPEED = 1.5
 const STEER_LIMIT = 0.4
 
-var steer_target = 0
+@export var engine_force_value := 40.0
 
-@export var engine_force_value = 40
+var _steer_target := 0.0
```
aaronfranke, line 8: Cleaning up the literals for things that should be floats is very helpful, if you want you can do it in a few more places, like changing `if speed < 5 and speed != 0:` below to `if speed < 5.0 and speed != 0.0:`

aaronfranke (commented):
This looks really good from a quick review, thank you!

I noticed that the `choose_*` files got bigger. You can use a tool called `oxipng` to compress PNG images, the command is `oxipng -o6 --strip all --zopfli --fix`.

jtnicholl:
> I noticed that the `choose_*` files got bigger. You can use a tool called `oxipng` to compress PNG images, the command is `oxipng -o6 --strip all --zopfli --fix`.

That's what I did, in order to optimize the new icon and screenshot. I ran it over the rest of the files again and there was no difference.
Viewing the commit in GitHub it seems to be saying the files went up in size by their total size, I think that's a bug in GitHub. The only change I made to those files was move them.

aaronfranke (changes requested):
Did a quick test. There are some problems. The trucks do not spawn in the correct place, and the joints are missing. These problems do not exist on the current 4.0-dev branch which [has Truck Town ported](https://github.com/godotengine/godot-demo-projects/pull/696) already.

jtnicholl:
Are you sure you're on the right branch? The joints are working fine when I test it. But when I test the current 4.0-dev branch the joints aren't attached. Also, the trucks are spawning in the same place on both branches. Where are they supposed to spawn?

jtnicholl:
Ok, I just tested with alpha 14 instead of building from latest master, and now I'm seeing the issues you mentioned. It looks like the node params on Joint got renamed some time between alpha 14 and now. As for spawning in the wrong location, I don't know why that's happens, but it's fixed if you build from latest master.

aaronfranke:
You're right, it does work with master. However, with master I noticed that there are weird shadows everywhere? It happens on the 4.0-dev branch and with your PR with Godot master, but it doesn't happen with alpha 14. Anyway, this isn't a problem with your PR, so we can merge it after alpha 15 is released.

Screen Shot 2022-08-26 at 1 21 38 PM

Calinou:
>  It happens on the 4.0-dev branch and with your PR with Godot master, but it doesn't happen with alpha 14.

This is a engine bug from the recent merge of octahedral compression.

jtnicholl:
I just cleaned up the vehicles and put them in a glb file as well, to fix the octahedral compression errors, a hole on the tow truck, and some backwards normals on the trailer. Also improved the wheels and branded the sides of the trailer with a waiting for Godot joke.

aaronfranke (approved):

aaronfranke:
Thanks! I made a follow-up PR: #768.

PR #770, opened 2022-09-03
Update 3D Platformer for 4.0

jtnicholl:
I've ported the 3D Platformer demo to 4.0.
I brought the player, enemy, and tiles into Blender for a cleanup. All of the binary data has been extracted out of the .tres and .tscn files.
Cleaning up the MeshLibrary for the GridMap was a nightmare. I needed it saved as text so I could change the order of the tiles to what they were before, otherwise they were scrambled. So I had to extract out the screenshots and set up the meshes and collision shapes manually, then edit the .tres file by hand to put them back in the right order. I ended up using an import script on the tiles to save the collision shapes generated during import, and used those. I also ran into an issue where the `save_to_file` importer option stopped working, so I saved the meshes using the import script too. I'll look into that further and open an issue later. I'm also going to write up a proposal for improving the GridMap workflow because that shouldn't have been so difficult. I don't like that I had to use an import script, especially for a demo.
I cleaned up the code a bit while updating it. It could still use more but it's better.
Finally I did some aesthetic changes to make it look nicer, like adjusting the sun direction, environment settings, and materials. I also removed the reverb effects since they weren't working and IMO they didn't really add anything when they were working.


jtnicholl:
Finally got around to opening the issues I mentioned. There's the proposal linked above, also godotengine/godot#65552 and godotengine/godot#65553.
Also, when I originally created the MeshLibrary, it didn't include the collision shapes. I thought that just wasn't a feature of the MeshLibrary export, but later I exported again and it did include them. I think that was another bug, but I haven't been able to replicate it again since.

jtnicholl:
Since it seems those issues might not get addressed in time for the 4.0 release, I've removed the tile import script and un-gitignored the meshes and collision shapes for now. We can change it with another PR later.
I also extracted the GridMap from the stage into its own binary `.scn` file since the tile data isn't really in a readable format.

adamscott:
Can you add screenshots? Or a video demo?

jtnicholl:
Yeah, this PR updates the screenshots included with the project. They were taken before I made some adjustments to the lighting and they looked grainy, so I've just retaken them.
[Screenshot 1](https://github.com/godotengine/godot-demo-projects/blob/a38c7f6aa4dea24b62cf2ba6fbccb0f7cded5ea5/3d/platformer/screenshots/face.png)
[Screenshot 2](https://github.com/godotengine/godot-demo-projects/blob/a38c7f6aa4dea24b62cf2ba6fbccb0f7cded5ea5/3d/platformer/screenshots/shoot.png)

Calinou:
Note that https://github.com/godotengine/godot-demo-projects/pull/782 also updates this demo, with a different visual rework. However, I haven't updated the demo's scripts yet. If you don't mind, I can borrow the code changes to this PR while crediting you as a co-author :slightly_smiling_face:

Calinou (approved):
Tested locally, it works.

I think it looks good aesthetically – I confused this with an earlier PR that changed the time of day to sunset.

Geometror (approved):
Tested the demo locally. Works fine and the code changes look good (e.g. the use of more descriptive variable names). Of course, there is still some room for improvement regarding the overall quality of the demo, but that's not the goal of this PR anyway. Great work!

Calinou:
Thanks!

PR #771, opened 2022-09-05
Check for obstacle before teleporting player in Grid-based Pathfinding with Astar

dev-gilbride:
Checks for obstacles before moving character when shift-clicking. Better matches normal clicking behavior for clicking on obstacles.

### DIFF
--- a/2d/navigation_astar/pathfind_astar.gd
+++ b/2d/navigation_astar/pathfind_astar.gd
@@ -130,6 +130,7 @@ func check_start_position(world_start):
 	var start_point = world_to_map(world_start)
 	if start_point in obstacles:
 		return false
+
 	return true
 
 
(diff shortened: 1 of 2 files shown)


2d/navigation_astar/pathfind_astar.gd
```diff
@@ -126,6 +126,13 @@ func is_outside_map_bounds(point):
 	return point.x < 0 or point.y < 0 or point.x >= map_size.x or point.y >= map_size.y
 
 
+func check_start_position(world_start):
+	var start_point = world_to_map(world_start)
+	if start_point in obstacles:
+		return false
+	return true
```
Calinou, line 133:
```suggestion
func check_start_position(world_start):
	var start_point = world_to_map(world_start)
	if start_point in obstacles:
		return false

	return true
```

Calinou:
Cherry-picked to the `4.0-dev` branch.

PR #774, opened 2022-09-16
Add compute shader demo

MoltenCoffee:
I've build a demo project using a compute shader for heightmap generation. It tackles both the calling of a compute shader as well as some important information on using shaders. All thoroughly commented so people can read along step by step. Furthermore it enables people to compare with CPU code and compare the performance of both options.

It is not the most simple compute shader demo imaginable, but I do feel it has the advantage of showing when a compute shader might be beneficial over CPU work, as well as give a concrete example of a use case.

Kind regards, TJ


misc/compute_shader/compute_shader.glsl
```diff
@@ -0,0 +1,45 @@
+#[compute]
+#version 460
+
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
```
clayjohn, line 4:
```suggestion
// Instruct the GPU to use 8x8x1 = 64 local invocations per workgroup
layout(local_size_x = 8, local_size_y = 8, local_size_z = 1) in;
```

I would use a local size of 8x8x1 as it is almost always better than 1x1x1. In this demo I get a 10% speedup on my intel iGPU (I expect there would be an even larger benefit on a dedicated GPU). 

More importantly, since this will be an official demo I guarantee this code will end up in many projects, so the added complexity will be worthwhile

misc/compute_shader/compute_shader.glsl
```diff
@@ -0,0 +1,45 @@
...
+
+// Prepare memory for the image, which will be both read and written to
+// 'restrict' is used to tell the compiler that the memory will only be accessed by the 'heightmap' variable
+layout(r8, binding = 0) restrict uniform image2D heightmap;
+layout(rgba8, binding = 1) restrict readonly uniform image2D gradient;
```
clayjohn, line 9:
```suggestion
layout(r8, binding = 0) restrict uniform image2D heightmap;
// `readonly` is used to tell the compiler that we will not write to this memory. This allows the compiler to make some optimizations it couldn't otherwise
layout(rgba8, binding = 1) restrict readonly uniform image2D gradient;
```

misc/compute_shader/compute_shader.glsl
```diff
@@ -0,0 +1,45 @@
...
+  // Multiply the pixel's red channel by the gradient's red channel (or any RGB channel, they are all the same except for alpha)
+  pixel.r *= gradient_color.r;
+  // If the pixel is below a certain threshold, this sets it to 0.0
+  // The step function is like a clamp, but it returns 0.0 if the value is below the threshold, and 1.0 if it is above
+  // This is why we multiply it by the pixel's value again, to get the original value back if it is above the threshold
+  // This shorthand replaces an if statement, which would cause branching and thus potential double execution of the shader
+  pixel.r = step(0.2, pixel.r) * pixel.r;
+
+  // Store the pixel back into the image
+  imageStore(heightmap, coords, pixel);
```
clayjohn, line 44:
```suggestion
  // Store the pixel back into the image
  // WARNING: make sure you are writing to the same coordinate that you read from. If you don't, you may end up writing to a pixel, before that pixel is read by a different invocation and cause errors
  imageStore(heightmap, coords, pixel);
```

misc/compute_shader/main.gd
```diff
@@ -0,0 +1,199 @@
...
+
+	var uniform_set := rd.uniform_set_create([heightmap_uniform, gradient_uniform], shader_rid, 0)
+
+	var pipeline := rd.compute_pipeline_create(shader_rid)
+	var compute_list := rd.compute_list_begin()
+	rd.compute_list_bind_compute_pipeline(compute_list, pipeline)
+	rd.compute_list_bind_uniform_set(compute_list, uniform_set, 0)
+	# This is where the magic happens! As our shader has a work group size of 1x1x1, we dispatch
+	# one for each pixel here. This ratio is highly tunable, and performance may vary
+	rd.compute_list_dispatch(compute_list, po2_dimensions, po2_dimensions, 1)
```
clayjohn, line 128:
```suggestion
	# This is where the magic happens! As our shader has a work group size of 8x8x1, we dispatch
	# one for every 8x8 block of pixels here. This ratio is highly tunable, and performance may vary
	rd.compute_list_dispatch(compute_list, po2_dimensions / 8, po2_dimensions / 8, 1)
```

If you accept my suggestion to change the workgroup size to 8x8x1, this will have to be changed to match.

misc/compute_shader/main.gd
```diff
@@ -0,0 +1,199 @@
...
+	rd.compute_list_end()
+
+	rd.submit()
+	# Wait for the GPU to finish
+	rd.sync()
```
clayjohn, line 133:
```suggestion
	# Wait for the GPU to finish
	# Normally you would do this after a few frames have passed so the compute shader can run in the background
	rd.sync()
```

clayjohn (commented):
Amazing work! I made a few suggestions that I think would improve this a little bit more, please let me know what you think

MoltenCoffee:
Seem like excellent suggestions to me! I've changed the name of the demo as well, to make it slightly more descriptive and to not hinder the possibility of adding more compute shader demos, I've seen multiple ray-tracing examples doing the rounds in the community, for example.

clayjohn:
> Seem like excellent suggestions to me! I've changed the name of the demo as well, to make it slightly more descriptive and to not hinder the possibility of adding more compute shader demos, I've seen multiple ray-tracing examples doing the rounds in the community, for example.

Wonderful! Looks like you missed the suggestion changing the workgroup size within the shader itself

MoltenCoffee:
> Wonderful! Looks like you missed the suggestion changing the workgroup size within the shader itself

Should be fixed now!

clayjohn (approved):
Looks good from the rendering content side of things. I'll leave it to the demo projects experts to confirm that every looks good to go from the demo project side of things

Great work! This will be a very helpful addition to the demos.

MoltenCoffee:
Thanks for taking the time! Especially the cleanup should have been there to begin with. I've taken the liberty of pulling the suggestions in.

Calinou:
I've pushed a commit to my fork which improves various things in the demo: https://github.com/Calinou/godot-demo-projects/tree/add-compute-shader-demo

If you want, you can incorporate those changes in your PR, but I can also merge the PR manually if that makes things easier for you.

MoltenCoffee:
Wonderful! Looks slick. I'm not sure I can pull those changes easily, so if you're able to do so easily, be my guest! If not I'll see if I can transfer changes manually later this week.

Calinou:
I think the demo is in a good state to merge now, so I can go ahead and do that this week.

Calinou:
Merged as part of https://github.com/godotengine/godot-demo-projects/pull/810. 

PR #775, opened 2022-09-18
Fix bomber demo in godot 4.0 beta 1

lejar:
Fix networking/multiplayer_bomber to work with godot 4.0 beta 1. Two major things seem to have changed:
1. Fonts objects no longer exist, and setting the override font should use the data from preload directly. See the code snippet at the end of the description here https://docs.godotengine.org/en/latest/classes/class_fontfile.html?highlight=font#description
2. The field names for the MultiplayerSpawner and MultiplayerSynchronizer nodes seem to have changed. 

In addition to those changes, the 4.0 beta 1 editor has lots of whitespace changes, and additional default fields. Also it automatically converted Position2D to Marker2D.


Faless:
Superseded by #784 .

PR #776, opened 2022-09-18
Update loading demos for 4.0

jtnicholl:
I have updated the rest of the demos in the loading folder.

- I completely removed Background Loading and Multiple Threads Loading, replacing them with a Threaded Loading demo. Those demos used `ResourceLoader.load_interactive`, which was removed. The new demo uses its replacements, `load_threaded_request` and `load_threaded_get`, which work differently.
- I added sprites to the Serialization demo to replace the `icon.png` placeholders. It's just some basic pixel art I did. The enemies are angry crabs, and the player is a sea turtle. There is an icon featuring these sprites.
- Scene Changer now uses both `change_scene_to_file` and `change_scene_to_packed`. There is a brief comment above the call to the latter explaining why you might use it over the former. I also added an icon that I made from the PackedScene icon.
- I added an icon to the Threads demo that was already ported and updated its screenshot.
- Autoloads got some minor code changes. I considered rewriting it because I don't think changing scenes is that great of an example usage of singletons, but I couldn't think of anything better. Also couldn't think of an icon to add.


jtnicholl:
The static checks seem to be complaining about the lack of a newline at the end of an SVG file. I don't think it's supposed to do that.

Calinou (approved):

aaronfranke:
> The static checks seem to be complaining about the lack of a newline at the end of an SVG file. I don't think it's supposed to do that.

It is supposed to do that. It's an easy fix if you open the SVG in the text editor.

Basically, whether the newline is desired depends on the answer to the question "Do we want our SVG files to be POSIX-compliant text files?", and the answer is yes at least for now.

jtnicholl:
Got it.
It had no newline at the end because I ran `svgcleaner --multipass` over it, as the docs suggest doing, which is why I was confused by it.

loading/serialization/enemy.gd
```diff
@@ -1,29 +1,30 @@
-extends CharacterBody2D
+class_name Enemy extends Node2D
 
-const MOVE_SPEED = 75
-const DAMAGE_PER_SECOND = 15
 
-# The node we should be "attacking" every frame.
-# If `null`, nobody is in range to attack.
-var attacking = null
+## Movement speed in pixels per second.
```
Geometror, line 4: Just a bit of nitpicking: You should use a single # for consistency here and below.

loading/serialization/player.gd
```diff
@@ -1,33 +1,38 @@
...
-@onready var progress_bar = $Sprite2D/ProgressBar
-
-
-func _process(delta):
-	# Player movement (controller-friendly).
-	var velocity = Vector2.ZERO
-	velocity.x = Input.get_axis(&"move_left", &"move_right")
-	velocity.y = Input.get_axis(&"move_up", &"move_down")
-	position += velocity * MOVE_SPEED * delta
+## Movement speed in pixels per second.
```
Geometror, line 4: Same here.

Geometror (approved):
Tested every demo locally and they all seem to work properly. Nice work!

loading/serialization/enemy.gd
Geometror: Oh, sorry, you're right. Reviewing while being quite tired is never good idea :) These doc comments should be kept as they are.

PR #778, opened 2022-09-18
Remove VisualScript demos from the 4.0 branch

jtnicholl:
If someone ends up maintaining VisualScript unofficially they can still get these demos from the history, if they want to update them and put them somewhere else.


aaronfranke (approved):
Also, even if VisualScript is brought back via GDExtension, the demos that are brought back would be significantly different, and potentially not even in this repo.

PR #779, opened 2022-09-20
Add 3d decals demo

dzil123:
Building off #756

Ran into godotengine/godot#64332 when zooming out all the way, so I replaced with lerpf.

To position the decals on keypress, I chose to replace the RayCast3D with a intersect_ray query because camera.project_position is in world space but RayCast3D's target is in local space.

To orient the decals, I opted to copy the camera's rotation. Another option would be to orient the decals toward the camera, or to use the raycast result's normal. I'm not sure what the best solution is here.


Calinou:
I integrated the changes into https://github.com/godotengine/godot-demo-projects/pull/756 (squashed) and added you as a co-author. Thanks :slightly_smiling_face:

PR #782, opened 2022-10-06
Update most demos for Godot 4.0.beta10

Calinou:
The HDR for 2D demo was renamed to Glow for 2D as autoexposure can no longer be used in 2D in 4.0.

The SDF font demo was renamed to MSDF font to denote that multi-channel distance field fonts are now used.

The OpenSimplexNoise demo was renamed to Noise Viewer as it now uses FastNoiseLite.

Various demos have received small improvements along the way (mainly cosmetic changes or usability improvements):

- All demos now use physical key inputs, so that non-QWERTY keyboard layouts work out of the box.
- GUI-only demos have low processor mode enabled and V-Sync disabled, for low input lag while keeping resource usage low. Tearing caused by the lack of V-Sync is usually not noticeable in GUI-only demos.
- In some demos, physics FPS was increased to 120 to improve smoothness on high refresh rate displays and decrease input lag. Demos have simple physics calculations compared to full-blown projects, so this change should be affordable on lower-end devices too.
- Glow for 2D demo now features a toggleable glow map (lens dirt effect).
- 2D particles demo now features turbulence, subemitters, collision and global coordinates animation examples (particles + PathFollow2D). Glow can also be toggled.
- Truck Town now features a race track on the outside of the town, built using CSGPolygon3D + Path3D.
- Truck Town now has a large floor plane that extends past the city for a better appearance. It can also be driven on if you need to see how vehicles behave on a large flat plane.
- Truck Town now features engine and impact sounds, blob shadows for cars, exhaust particles and dynamic FOV based on speed. These changes don't require much additional code, but improve the feel of the demo a lot when put together. The speedometer's color changes depending on speed thanks to a Gradient resource, and the unit can be changed by pressing <kbd>U</kbd> or <kbd>Select</kbd> as an alternative to clicking the button (implemented using BaseButton's shortcut property).
- Truck Town has revised textures (AI upscaled to 2×, grass color tweaked to be more pleasing to the eyes). The level geometry now features smoother terrain.
- Truck Town's UI is now keyboard and gamepad-friendly. A title label was also added.
- Platformer 3D now features a reset button (<kbd>R</kbd>) that is automatically used when falling off the map, like Kinematic Character 3D. Platformer 3D also has increased gravity and jump height for better jump feeling. Jump height can now be reduced by releasing the jump key mid-air.
- The Platformer 3D stage was extended to make more areas reachable.
[...]

PR #783, opened 2022-10-15
Change 3D Platformer Visuals

MangoDragonHub:
The following changes I made is updating the visuals for the 3D Platformer Demo for Godot 4.
I made a new Core Material and separate texture to update the original art direction that was in the original demo

Before (Godot 3)
Screenshot from 2022-10-02 21-21-09

After (Godot 4)
Screenshot from 2022-10-04 14-05-46
)


aaronfranke:
Why is the player's body transparent?

Also, please don't title things as "update" when they are significantly changing things.

adamscott:
@MangoDragonHub is a designer, not much a developer. I think this PR must be on the updated visual rather than the code itself.

MangoDragonHub:
@aaronfranke I think the reason why the player looks transparent is because of how metallic and reflective the Player's material is. 
Also, when I upgraded the project from Godot 3 to 4, a lot of things broke.

Calinou:
> I think the reason why the player looks transparent is because of how metallic and reflective the Player's material is.

No, the player's body is clearly invisible on the screenshot.

MangoDragonHub:
Ah, you're talking about the body it self, not the material. Yeah, due to the Godot 3 to Godot 4 conversion tool not being ready, A LOT of stuff broke. Not just the scripting, all the character animations broke too. That's why the character and even the enemies in the background looks like that.

jtnicholl:
Ignoring the fact that there are unresolved merge conflicts committed and it doesn't even run, I don't like these changes aesthetically.
The level geometry is still very blocky yet the textures are very detailed, it ends up looking like those [super-HD Minecraft texture packs](https://i.redd.it/2ioobk3vhgn21.jpg) or something. It's not very consistent. It also doesn't tile seamlessly.
I made a PR to update this demo and made changes to the visuals there as well (#770), and I went in the opposite direction. I adjusted the lighting to make the scene brighter, and I disabled texture filtering and set the texture compression to lossless to make it _more_ cleanly pixelated, and while it's just my opinion I think that looks a lot nicer. More HD detail is not always better.

Calinou:
Superseded by https://github.com/godotengine/godot-demo-projects/pull/770 and https://github.com/godotengine/godot-demo-projects/pull/782. Thanks for the contribution nonetheless!

Issue #786, opened 2022-10-29
Web demos are broken


rpaladin:
**Bug description:** Game does not start. Tested in both latest Edge and Firefox versions, no luck.
**Reference link:** <https://godotengine.github.io/godot-demo-projects/3d/platformer/>

## Console Errors - Edge

![image](https://user-images.githubusercontent.com/69180012/198813711-f300f8e6-509f-44a3-9555-69bbb6269162.png)

## Console Errors - Firefox

![image](https://user-images.githubusercontent.com/69180012/198813692-d3aa29f5-5ef1-4103-8777-46ef52eaaa89.png)

Calinou:
I can confirm this on Firefox 105 (Linux). The online demos are using a old Godot version (3.3.3) and it needs to be updated to 3.5.

antonio-hickey:
Does Web Signaling Demo work for anyone else actually exported to web? 

I'm on Godot 4.1.1, and exported the Web Signaling Demo to HTML5, and set up my signaling server but the web is not calling it. (it does from being ran in Godot).

Calinou:
> I'm on Godot 4.1.1, and exported the Web Signaling Demo to HTML5, and set up my signaling server but the web is not calling it. (it does from being ran in Godot).

The demo was last updated for 4.0.beta4, so I don't know if it's expected to work. cc @Faless

JoelCreatesGames:
confirming that https://godotengine.github.io/godot-demo-projects/3d/platformer/ and https://godotengine.github.io/godot-demo-projects/2d/isometric/ are still broken

I get `indirect call to null` in firefox and `null function or function signature mismatch` on chrome

FracPro:
Yeah same issue, none of it works out of the box, another home grown adventure appreciate the hard work though!

Calinou:
@FracPro Please open a new issue with more details (which demos don't work, which OS/browser you're using, and so on).

MariaCodesXoXo:
I've tried the demo https://godotengine.github.io/godot-demo-projects/3d/platformer/ on both Microsoft edge and duckduckgo and it worked flawlessly, the issue might've now become only firefox specific as [comment ](https://github.com/godotengine/godot-demo-projects/issues/786#issuecomment-1295969701) suggests

<img width="1917" height="1023" alt="Image" src="https://github.com/user-attachments/assets/120e87e4-0ae6-4660-9edd-193f59ccefcb" />

[...]

PR #788, opened 2022-10-29
Update online HTML5 demos to use Godot 3.5.1

Calinou:
This closes https://github.com/godotengine/godot-demo-projects/issues/786. Wait for CI passing before merging.


aaronfranke (approved):
Pre-emptively approving, we can merge when you are ready.

Calinou:
Platformer 3D demo still fails to run in Chromium 102 after being updated to Godot 3.5.1:

image

aaronfranke:
Can you take another stab at this? We updated the demos on master to 3.5 so that may help.

aaronfranke:
I think this isn't needed anymore since the demos online should use 4.x, so I'll close this.

PR #789, opened 2022-11-15
[WebRTC] Update demos to beta4.

Faless:
Updated the minimal and signaling demos to Godot 4.0 beta4 ([extension](https://github.com/godotengine/webrtc-native/releases/tag/1.0.0-beta4)).

The minimal demo is mostly unchanged, while the signaling demo has been improved:
- The signaling server protocol has been rewritten to use JSON format to be more readable.
- Lobbies now support both mesh and client/server modes (selected during creation).
- The client/server mode uses the SceneMultiplayer relay mode as implemented in beta4 (https://github.com/godotengine/godot/pull/67094).
- The demo now uses an RPC for pinging, and connects to the MultiplayerAPI instead of using the raw MultiplayerPeer.


Faless:
> Although, the README.md appears to no longer accurately describe the singnalling protocol: rather than using strings, it's using JSON now.

I totally forgot about the readme, updated now.

PR #795, opened 2022-11-24
[3.x] [WebSocket] Fix peer disconnection not closing the network layer.

### LINKED ISSUE #793: WebSocket Multiplayer - Server doesnt receive peer disconnect event
WebSocket Multiplayer

Arch Linux x86_64, Godot v3.5.1.stable.official

**Issue description:**
Server doesnt receive event when client peer disconnects using the disconnect button (although it does when client window is closed without using the disconnect button)
Expected: network_peer_disconnected to be fired
Result: it is not when using the disconnect button, leaving a "zombie" player in server list
It does disconnect when client connects again

Faless:
We were removing it from the MultiplayerAPI but not immediately destroying it (to avoid tricky in-signal de-referencing). We still need to properly close the connection in this case if we want the remote peer to be notified immediately.

Fixes #793

### DIFF
--- a/networking/websocket_multiplayer/script/main.gd
+++ b/networking/websocket_multiplayer/script/main.gd
@@ -79,6 +79,10 @@ func _on_Host_pressed():
 
 
 func _on_Disconnect_pressed():
+	if peer is WebSocketServer:
+		peer.stop()
+	elif peer is WebSocketClient:
+		peer.disconnect_from_host()
 	_close_network()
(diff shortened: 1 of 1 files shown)


OsakiTsukiko:
I think tree network peer is preferred over using the peer var

akien-mga (approved):

akien-mga:
Thanks!

PR #801, opened 2022-11-30
Add an occlusion culling and mesh LOD demo

Calinou:
~~**Question:** Should we convert screenshots to WebP? It would be 30% smaller for this image, while remaining lossless. The asset library backend should be modified to allow linking to WebP images if this isn't done already.~~

**Edit:** Screenshot included in the PR has been converted to WebP.

occlusion_culling_mesh_lod


jtnicholl:
What's the reasons the screenshots need to be lossless?

Calinou:
> What's the reasons the screenshots need to be lossless?

They don't *need* to be, but it's better if they are :slightly_smiling_face:

akien-mga:
Using WebP for screenshots sounds good yeah.
I also think using lossy would be fine if it helps reduce the size further. We don't really need lossless for this.

Calinou:
> Using WebP for screenshots sounds good yeah. I also think using lossy would be fine if it helps reduce the size further. We don't really need lossless for this.

I've replaced the screenshot with a lossy quality 100 WebP. GitHub is able to display it: https://github.com/Calinou/godot-demo-projects/tree/add-occlusion-culling-demo/3d/occlusion_culling_mesh_lod

akien-mga (approved):
Haven't tested the code but this should be fine to merge IMO.

PR #766, opened 2022-08-10
Rework basic 3D navigation demo for Godot 3.5

smix8:
Reworks basic 3D navigation demo to use new NavigationServer from Godot 3.5.
Also changes the segment query to be more self-explaining.

PR #765, opened 2022-08-10
Rework basic 2D navigation demo for Godot 3.5 and later

smix8:
Reworks basic 2D navigation demo to use new NavigationServer and NavigationAgent from Godot 3.5 / 4.0.

Keeps the demos scope to basic navigation paths and movement, so no avoidance or physics.

Fixes tutorial issue from https://github.com/godotengine/godot-docs/issues/6055

The demo is prepared for Godot 3.5 but I did not find a 3.5 branch so I picked master.
For Godot 4.0 port change 'Navigation2DServer' to 'NavigationServer2D' and export/onready to new syntax.

PR #806, opened 2022-12-18
Convert 3D navmesh demo to use NavigationAgents

idbrii:
This is an extension to @smix8's #766 that uses NavigationAgents like #765.

Use agents and update to match the style of PR #765 which does the same change for 2d navigation:
* Separate character from input scripts.
* Character uses navagent to determine next location and constant speed
  to determine how much to move.
* Use a line node.

This layout is closer to how users are likely to structure code and the separation makes it easier to understand the individual pieces. (Put character path following on the character since users are likely to have multiple characters and put line drawing on a separate node to make it easy to remove.)


aaronfranke:
Can you rebase this PR?

idbrii:
@aaronfranke Done. Used smix's rebased commit from their PR which is now just adding locals for clarity since 9e68af38 actually did the 3.5 compatibility rework, but I didn't modify their commit message.

Still works as expected.

3d/navmesh/line3d.gd
```diff
@@ -0,0 +1,23 @@
+extends ImmediateGeometry
+
+
+var m := SpatialMaterial.new()
```
aaronfranke, line 4: I would prefer `material`. At a minimum `mat`. Avoid single-letter variable names.

3d/navmesh/character.gd
```diff
@@ -0,0 +1,44 @@
+extends Spatial
+
+const Line3D = preload("res://line3d.gd")
+
+
+export(float) var character_speed := 10.0
+export var show_path := true
+
+onready var navigationagent: NavigationAgent = $NavigationAgent
+onready var navigationpathdebugline: Line3D = $Line3D
```
aaronfranke, line 10: These variables are missing underscores. Also, the node should probably be named "DebugLine".

3d/navmesh/character.gd
```diff
@@ -0,0 +1,44 @@
+extends Spatial
+
+const Line3D = preload("res://line3d.gd")
+
+
+export(float) var character_speed := 10.0
+export var show_path := true
```
aaronfranke, line 7: For Godot 3.x, the preferred style in the demos is to not use static typing. See https://github.com/godotengine/godot-demo-projects/pull/350#discussion_r319752695

3d/navmesh/character.gd
idbrii: I can rename Line3D -> DebugLine but I called it that because of [Line2D](https://docs.godotengine.org/en/stable/classes/class_line2d.html). Drawing unit paths isn't inherently debug. (Although I should navigationpathdebugline -> navigationpathline.)

aaronfranke: No, I am fine with the class being called Line3D. But the instance does not need to be named the same as its class. The node should be given a usage-context-sensitive name. If I had a KinematicBody node named Player, that does not mean I want the class to be renamed to Player.

Calinou (approved):

PR #818, opened 2023-01-08
Fix and update the Operating System Testing demo

zoeesilcock:
Some fixes for the `misc/os_test` demo. One of them is an API change in 4.0 and the other two are platform specific. I wouldn't be surprised if the file URL change could be relevant on Linux also, but I don't have a machine setup to test that.

The demo now works without errors. Tested on beta 12.


Calinou (approved):

PR #828, opened 2023-01-23
Fix joypads demo button/axis mappings for 4.0

akien-mga:
- Many of the values in the diagram were out of sync with their current enum values.
- Properly moved triggers from buttons to axes.
- Removed 'Axis 10' (11th axis), we only support 10.
- Removed buttons beyond SDL_MAX.
- Show approximate position for MISC button, though we lack it on the gamepad image.

Couldn't really test the latter as with my Xbox Series gamepad on Linux it errors with:
```
ERROR: Index (int)p_button = 260154452 is out of bounds ((int)JoyButton::MAX = 128).
   at: joy_button (core/input/input.cpp:953)
```
But that's a Godot bug.


Calinou (approved):
Thanks!

PR #835, opened 2023-01-30
Improve and update Control Gallery for Godot 4.0

Calinou:
- Add a Tree example.
- Add more tabs to make tab scroll buttons visible.
- Use default project font, which is now a DynamicFont in Godot 4.0.
- Make splitters thinner but have a larger draggable area, thanks to the new Minimum Grab Thickness theme item.
- Tweak background color to be more visually pleasant.
- Use Compatibility backend to maximize performance and compatibility.
- Disable V-Sync to reduce input lag (while keeping low-processor mode).

## Preview

Control Gallery


gui/control_gallery/tree.gd
```diff
@@ -0,0 +1,13 @@
+@tool
+extends Tree
+
+
+func _ready() -> void:
+	var root := create_item()
```
aaronfranke, line 6: Since `TreeItem` is not explicitly written on the line, I would prefer to write the type here. Since this method is called multiple times we could just put it on the first one, or we could put it on all of them.

aaronfranke (approved):

aaronfranke (approved):

PR #838, opened 2023-01-31
Add quotes to rpc flags

Williangalvani:
I'm looking at the multiplayer examples and they didn't work on Beta16.
See https://github.com/godotengine/godot/issues/72218


aaronfranke (approved):

akien-mga:
I pushed an update to keep the:
```
@rpc(...)
func name():
```
formatting consistent, there was a mix of both inline and standalone annotation usage in the existing code.

akien-mga:
Thanks!

PR #839, opened 2023-02-08
Update Global Illumination demo for 4.0.rc1

Calinou:
This uses lossless compression for EXR lightmaps to prevent overly long import times and decreased quality. The increased memory usage shouldn't be too noticeable since this is a small scene.


clayjohn (approved):
seems good to me. Indeed using lossless will have to be an acceptable workaround for now until we fix the .exr import bug

PR #840, opened 2023-02-08
Fixed isometric demo rendering for Godot 4.0-rc1

mrcdk:
`y_sort` wasn't enabled in any node so it wasn't rendering correctly. For example, the troll was always rendered on top of the tilemaps. I enabled the property where it was necessary.

I took the opportunity to tweak a few things: 

- I merged the different atlases the `TileSet` had into one single atlas (using the Atlas Merging Tool)
- I merged both tilemaps into one `TileMap` with 2 layers (`floor` and `walls`). Each layer has `y_sort` enabled and correct `y_sort_origin` values.
- I tweaked the door's collision shape and also tweaked their `y_sort_origin` to avoid parts of the troll to clip between door's tiles.
- I tweaked the troll's shadow position and skewed it a bit.


Calinou (approved):
Thanks!

This will be useful as part of my TileSet/TileMap documentation updates :slightly_smiling_face:

PR #848, opened 2023-02-18
misc/window_management: Fix screen size and label order

raggi:


Calinou (approved):
Tested locally, it works.

There are outstanding issues we should resolve though (see my other comment).

Calinou:
The demo runs extremely slow for me, making even X11 mouse cursor slow. Querying the screen parameters constantly is likely a source of performance problems, at least on Linux.

raggi:
Thanks for filing that bug @Calinou. That issue didn't repro on Windows, so I agree perhaps it should be cached in the library somewhere.

I will take a look at formatting, but I don't see any output from the CI check. Do you know what's up there?

raggi:
>This should be fixed at the engine level by caching parameters somehow. As a stop-gap solution, you could change the label's setter (only for the refresh rate) by moving it to _ready() instead of _physics_process().

done!

Calinou (approved):

PR #851, opened 2023-02-21
Update, clean up, and reorganize the 2D platformer

jtnicholl:
I've updated and cleaned up the (non-physics) 2D platformer on the 4.0-dev branch.

- Everything has been updated and fixed for latest Godot master branch.
- Files have been reorganized. Instead of `assets` and `src` folders, there is `level`, `player`, `enemy`, and `gui`. All files are now in snake case.
- The player movement has been changed. The player has some momentum and takes 1/6th of a second to accelerate and decelerate. They can also now double jump, which provides a short boost to horizontal momentum. (I found this made it more fun to jump around the level, though this is subjective. I can remove it if you don't agree.) Their overall speed is also slightly faster.
- Enemies now pass through players. Before, they could push players inside walls, this was the simplest way to fix that. They also move a bit slower.
- The level now has a StaticBody2D with two world boundary shapes, to prevent players from escaping with their new double jump.
- Switching to splitscreen and back is now done via a button on the pause menu, instead of an input.
- Refactoring and code cleanup, especially with regards to splitscreen.


jtnicholl:
Has the process for setting up 2D splitscreen changed significantly in 4.0? I can't get the second viewport to track player 2. Setting player 2's camera to current only stops player 1's camera from moving.

Calinou:
"Current" Camera2D handling was reworked in 4.0 by @KoBeWi.

jtnicholl:
I did see that, but it seems like setting player 2's camera to current after assigning it to viewport 2 should still work. Instead, it causes player 1's camera to stop moving, which doesn't make any sense to me.

KoBeWi:
There is a bug in Camera2D. Multiple viewports are broken right now .-.

hsandt:
Oh, that's many more changes that I was working on!

Fortunately I didn't go too far in my own PR, but to avoid overlapping work on porting the demos in the future, I suggest that we notify others on the tracker issue (https://github.com/godotengine/godot-demo-projects/issues/697) when we start/finish a certain work (if you started this PR last year though, maybe the tracker issue didn't even exist at the time!)

Now, you got me curious as I had a hard time converting certain things, so I compared my changes with yours where relevant:

- the tileset was pretty badly setup, I had to fix the slope meta-tile and all its variants. Plus, a visual bug on transposed tiles made editing even harder. Basically, I adjusted the Texture Origin and added a offset/rotation to the polygon points. On your side, I see you extended the meta-tile height to 96px, which also had the effect to fix the offset, and happened to work because there was nothing above said meta-tile. You also removed the unused ceiling meta-tiles, so you didn't have to bother with the terrible cases I went through with transposed tiles. I'd say it's okay, we can always re-add them later if we need them (hopefully after the visual bug is fixed). I was also told that the demo doesn't need to focus too much on tilemap features, so it should be enough to just make it work.

image

- I extracted gravity effect into some `Actor.apply_gravity` method (since super method `Actor._physics_process` is not automatically called in Godot 4 anymore) called by Player and Enemy. Whereas you removed Actor entirely, so Player and Enemy just use their own gravity variable. I don't mind, but it'd be nice to know if the original intention was to demonstrate some inheritance and possible add more common code to Actor later, or if it was just done this way because it was convenient with the automated super._physics_process call, which is now irrelevant.

- Good work on using consistent documentation comments with `##`, I did the same. But you also removed the lengthy comment on _physics_process. Is it something we want to keep (and adapt to new code) for beginners?

Other things I like:
- single Music player for splitscreen instead of 2 Music players played at -6dB... It removed the need for Music.gd entirely.

Otherwise, I have some difficulties comparing all the file changes because the move/rename and changes have been squashed together and my git client doesn't manage to track the files down, so I admit at least 2 separate commits would have helped in this case. Possibly more to distinguish "porting to Godot 4" and "adding features" (it would also make it easier to git bisect if we find regressions later).

jtnicholl:
Yeah, poor timing. I think I might have seen your comments there and then forgotten about them in the meantime, apologies. Opening draft PRs before much work has been done is probably the best way to go.

> But you also removed the lengthy comment on _physics_process. Is it something we want to keep (and adapt to new code) for beginners?

I'm not entirely sure about this, actually. I think the 2D platformer was written to be a beginner tutorial, but none of the other demos have such detailed explanations. Not even Dodge the Creeps, which was created for the absolute beginner tutorial in the docs. I figured it'd be better to make it consistent with the other demos.

akien-mga:
Thanks!

hsandt:
Oops, I forgot to talk about the most important physics change for this demo, namely how `move_and_slide` works. I did some study to compare previous and new behavior, namely speed on slopes:

Godot 3 demo:
- moving down a slope at low gravity multiplies flat speed by 1.41 (sqrt(2)). More gravity increases this factor (normal gravity gives 1.56).
- moving up a slope at low gravity multiplies flat speed by 1.41 (sqrt(2)). Yes, character was moving faster, but only because we were keeping a stable speed along X. The upward projection on a diagonal slope would cause the increase in velocity magnitude. More gravity decreases this factor (normal gravity gives 1.20).

Godot 4 demo with this PR:
- moving down a slope at low gravity multiplies flat speed by ~1.44 (a little more than (sqrt(2)) at low gravity. Gravity doesn't seem to affect that number.
- moving up a slope at low gravity multiplies flat speed by ~0.70 (a little less than 1/sqrt(2)). More gravity reduces this factor (normal gravity gives 0.65)

Godot 4 demo with my PR draft:
- moving down a slope multiplies flat speed by ~1.43 (a little more than (sqrt(2)) at low gravity. Gravity doesn't seem to affect that number.
- moving up a slope multiplies flat speed by ~0.71 (1/sqrt(2)). More gravity reduces this factor further (normal gravity gives 0.65)

So not surprisingly, our similar formulas gave the same result, and the new algorithm applying a factor = dot product between intended motion and ground tangent gives more organic variations of speed depending on the way we go. We can see later if we prefer a more stable speed (ideally we'd give pointers to users on how to customize this, but I need to think about how to do this without cluttering the demo).

Anyway, I'm glad there is a stable basis to work on from now on!

PR #858, opened 2023-02-25
Fix "is mount_event" check in gui_in_3d to work with both 4 beta and 4 rc 

KMouratidis:
The existing code for the `gui_in_3d` demo doesn't work on the latest RC, it throws this error:

    Error at (39, 18): Could not find type "mouse_event" in the current scope".

I changed it to use get_class and a simple match which works on beta17 and rc5

### DIFF
--- a/plugins/addons/main_screen/main_screen_plugin.gd
+++ b/plugins/addons/main_screen/main_screen_plugin.gd
@@ -29,7 +29,7 @@ func _make_visible(visible):
 
 # If your plugin doesn't handle any node types, you can remove this method.
 func _handles(object):
-	return object is preload("res://addons/main_screen/handled_by_main_screen.gd")
+	return is_instance_of(object, preload("res://addons/main_screen/handled_by_main_screen.gd"))
 
 
 func _get_plugin_name():
--- a/viewport/gui_in_3d/gui_3d.gd
+++ b/viewport/gui_in_3d/gui_3d.gd
@@ -36,7 +36,7 @@ func _unhandled_input(event):
 	# Check if the event is a non-mouse/non-touch event
 	var is_mouse_event = false
 	for mouse_event in [InputEventMouseButton, InputEventMouseMotion, InputEventScreenDrag, InputEventScreenTouch]:
-		if event is mouse_event:
+		if is_instance_of(event, mouse_event):
 			is_mouse_event = true
 			break
 


viewport/gui_in_3d/gui_3d.gd
```diff
@@ -35,10 +35,9 @@ func _mouse_entered_area():
 func _unhandled_input(event):
 	# Check if the event is a non-mouse/non-touch event
 	var is_mouse_event = false
-	for mouse_event in [InputEventMouseButton, InputEventMouseMotion, InputEventScreenDrag, InputEventScreenTouch]:
-		if event is mouse_event:
+	match event.get_class():
```
dalexeev, line 38: This is an acceptable solution, but many users misunderstand `get_class()` (it doesn't work with custom classes). An alternative is the new `is_instance_of()` function, which was added for dynamic type checking.

dalexeev:
By the way, could you fix this line as well:

https://github.com/godotengine/godot-demo-projects/blob/c8bb3b874aec8e8bd4a6c5c7aed855d988d9e89f/plugins/addons/main_screen/main_screen_plugin.gd#L34

```gdscript
return is_instance_of(obj, preload("res://addons/main_screen/handled_by_main_screen.gd"))
```

See godotengine/godot#73489 for details.

viewport/gui_in_3d/gui_3d.gd
KMouratidis: `is_instance_of` seems to only be available in `rc` (I'm using 5) and not `beta`. Is that fine?

dalexeev: I think yes. Since in beta the `is` operator worked fine with variable expressions.

KMouratidis: Done. Thanks!

plugins/addons/main_screen/main_screen_plugin.gd
```diff
@@ -29,7 +29,7 @@ func _make_visible(visible):
 
 # If your plugin doesn't handle any node types, you can remove this method.
 func _handles(object):
-	return object is preload("res://addons/main_screen/handled_by_main_screen.gd")
+	return is_instance_of(obj, preload("res://addons/main_screen/handled_by_main_screen.gd"))
```
dalexeev, line 32:
```suggestion
	return is_instance_of(object, preload("res://addons/main_screen/handled_by_main_screen.gd"))
```

Sorry, I made a mistake in my comment.

KMouratidis: My bad for copying blindly.

dalexeev (approved):
Looks good to me! I've checked the `.gd` files in this repository with a regular expression, and it looks like `is` doesn't need to be replaced with `is_instance_of` anywhere else.

You should squash the commits into one to keep the Git log clean. See [Pull request workflow](https://docs.godotengine.org/en/stable/community/contributing/pr_workflow.html).

KMouratidis:
Done (sorry, forgot to comment earlier)

akien-mga (approved):

akien-mga:
Thanks!

PR #859, opened 2023-02-26
Improve Audio Spectrum demo

Calinou:
- Draw fake reflections for spectrum bars.
- Increase the bars' height to better make use of the window space.
- Use Compatibility rendering method for greater performance and compatibility.
- Update screenshot and icon.
- Remove a stray `.import` file.

## Preview

spectrum


akien-mga:
Thanks!

PR #860, opened 2023-02-26
Add a 3D visibility ranges (HLOD) demo

Calinou:
This demo showcases how to set up a hierarchical LOD system using visibility ranges.

This can improve performance significantly in 3D scenes by reducing the number of draw calls and polygons that have to be drawn every frame.

Use WASD or arrow keys to move, and use the mouse to look around. Press <kbd>L</kbd> to toggle the use of visibility ranges. Press <kbd>F</kbd> to toggle the fade mode between *transparency* (the default in this demo) and *hysteresis* (which is slightly faster, but results in more jarring transitions).

> [!NOTE]
>
> Performance is expected to decrease significantly after disabling visibility ranges, as all trees will be drawn with full detail regardless of distance.

## Preview

visibility_ranges webp


jtnicholl:
I think it would be good to also demonstrate what the `visibility_parent` option does, and how to use it.

Calinou:
> I think it would be good to also demonstrate what the `visibility_parent` option does, and how to use it.

I'm not sure how it should be used exactly. For instance, I don't really understand why you need `visibility_parent` when you can just hide and show child nodes as I did.

JFonS:
`visibility_parent` can be used to automatically show/hide instances based on whether the lower LOD levels are visible. This way you can control lower levels with visibility ranges and the higher levels will only show when needed.

Here I found one of the test scenes I used while testing my PR, it shows how the setup works with three LOD levels: 
[LOD_test.tar.gz](https://github.com/godotengine/godot-demo-projects/files/11986813/LOD_test.tar.gz)

Calinou:
I've gotten Visibility Parent to work locally, but this prevents fading from working at all in the demo, unfortunately. I'd prefer keeping the fading as it helps reduce visible pop-in a lot.

PR #862, opened 2023-02-27
Remove unnecessary `randomize()`

dalexeev:
[@GlobalScope.randomize](https://docs.godotengine.org/en/latest/classes/class_%40globalscope.html#class-globalscope-method-randomize):

> Note: This function is called automatically when the project is run. If you need to fix the [seed](https://docs.godotengine.org/en/latest/classes/class_%40globalscope.html#class-globalscope-method-seed) to have consistent, reproducible results, use seed to initialize the random number generator.

### DIFF
--- a/2d/bullet_shower/bullets.gd
+++ b/2d/bullet_shower/bullets.gd
@@ -24,8 +24,6 @@ class Bullet:
 
 
 func _ready():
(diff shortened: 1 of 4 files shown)


Calinou (approved):
Thanks!

PR #866, opened 2023-03-01
Update "Dodge the Creeps" demo

dalexeev:
* Fix `AnimatedSprite2D` (`playing` → `play()`, `frames` → `sprite_frames`).
* `emit_signal("signal_name")` → `signal_name.emit()`.
* Use `Array.pick_random()` instead of `randi() % array.size()`.
* Use `Vector2.clamp()` instead of `clamp()` on components.
* Use `StringName` literals for consistency.
* Switch renderer to Vulkan Mobile to match description in README (this is a bug due to project setting rename).

### DIFF
--- a/2d/dodge_the_creeps/Main.gd
+++ b/2d/dodge_the_creeps/Main.gd
@@ -12,7 +12,7 @@ func game_over():
 
 
 func new_game():
-	get_tree().call_group("mobs", "queue_free")
+	get_tree().call_group(&"mobs", &"queue_free")
 	score = 0
 	$Player.start($StartPosition.position)
 	$StartTimer.start()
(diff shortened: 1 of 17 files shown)


akien-mga (approved):

2d/dodge_the_creeps/Main.gd
```diff
@@ -12,7 +12,7 @@ func game_over():
 
 
 func new_game():
-	get_tree().call_group("mobs", "queue_free")
+	get_tree().call_group(&"mobs", &"queue_free")
```
vgezer, line 15: are these ampersands necessary? unfortunately, could not find a topic related to that. if yes, the documentation must also be updated accordingly.

PR #867, opened 2023-03-01
Fix renamed `vulkan/rendering/back_end=1` setting (Vulkan Mobile)

dalexeev:
See #866. These demos work on Forward+ in master, but READMEs say Vulkan Mobile.


akien-mga:
Thanks!

PR #882, opened 2023-03-12
Change Icon header ordering for 2.5d GD Script Demo

Starkium:
Just following the rules! 
Moved up the icon statements above the other includes. This got rid of errors preventing plugin from loading.

There's still a race condition when the plugin starts up. It's looking for an editor interface child that hasn't been loaded yet. May want to use _ready to define some behavior before _enter_tree is used.


aaronfranke (approved):
Yup this was changed in Godot 4.0, thanks!

PR #883, opened 2023-03-13
Update 3D Graphics Settings demo for 4.0

Calinou:
- Fix scripts for Godot 4.0.stable.
- Tweak Low preset (enable FXAA, disable MSAA to improve performance).
- Colorize FPS counter depending on framerate (red = bad, yellow = OK, green = good, cyan = great).
- Use first-class Signal syntax to reduce reliance on strings.

### DIFF
--- a/3d/graphics_settings/settings.gd
+++ b/3d/graphics_settings/settings.gd
@@ -9,6 +9,8 @@ extends Control
 @onready var fps_label := $FPSLabel
 @onready var resolution_label := $ResolutionLabel
 
+var counter := 0.0
+
 # When the screen changes size, we need to update the 3D
 # viewport quality setting. If we don't do this, the viewport will take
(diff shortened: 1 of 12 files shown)


3d/graphics_settings/settings.gd
```diff
@@ -292,10 +303,8 @@ func _on_glow_option_button_item_selected(index: int) -> void:
 		world_environment.environment.glow_enabled = false
 	if index == 1: # Low
 		world_environment.environment.glow_enabled = true
-		RenderingServer.environment_glow_set_use_high_quality(false)
```
Calinou, line 295: For future reference: This function was removed before 4.0.stable was released, as the high-quality glow mode was removed in 4.0. It didn't work correctly and often had no quality benefit, despite being slower.

PR #885, opened 2023-03-13
Use 128×128 WebP icons for all demos

Calinou:
This results in better-looking icons on hiDPI displays.

This also fixes the icon not appearing in the Pseudolocalization demo scene.

https://github.com/nihui/waifu2x-ncnn-vulkan was used to upscale 64×64 icons to 128×128. Script run in the repository root to upscale icons, then convert them to lossless WebP:

```bash
for icon in $(identify $PWD/**/icon.png | rg -i "64x64" | cut -d " " -f 1); do
    waifu2x -i $icon -o $icon
done

parallel "cwebp -z 9 {} -o {.}.webp && rm {}" ::: **/icon.png
```

PS: I'd like to add a C# badge on top of all C# demos. Putting this here so I remember for a future PR :slightly_smiling_face:

[csharp-icon-overlay.zip](https://github.com/godotengine/godot-demo-projects/files/10962850/csharp-icon-overlay.zip)


aaronfranke (approved):
Unfortunately GitHub is not able to preview the `.webp` files in the web view. I am blindly assuming this is good. But in general 128x128 WebP thumbnails is very good (the only thing better would be SVG files).

PR #892, opened 2023-03-19
Add an optional extra RESET step in Tween demo

### LINKED ISSUE #890: Tween Demo: icon is not reset between animation loop iterations
2d/tween

irrelevant

**Issue description:**
The icon is no reset between each loop of animation, so after the first one, it start with its last scale & position (and any other parameters according to selected steps)

steps:
-set loop param to 3 or more
-deselect some transform and increase anim speed
-run the demo

result: the second iteration show icon start in its last position in big size

suggestion:
add a final step to set icon back to its former size & position (and any other parameters)

**Screenshots of issue:**
tween_demo_issue

KoBeWi:
Closes #890

Adds extra checkbox, enabled by default:
image

### DIFF
--- a/2d/tween/main.gd
+++ b/2d/tween/main.gd
@@ -121,16 +121,25 @@ func start_animation():
 		tween.tween_callback(icon.show)
 		tween.tween_callback(icon.set_self_modulate.bind(Color.WHITE))
 
+	# RESET step
+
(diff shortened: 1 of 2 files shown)


Calinou (approved):
Thanks!

PR #894, opened 2023-03-23
Add a large world coordinates demo

Calinou:
This allows testing double-precision builds and see if they work as expected, while providing an interactive demonstration.

Some core shaders fail to load when using a double-precision build (which leads to ambient shading being pure black); I don't know why. This doesn't occur on a brand new project.

### Using single-precision build (including official builds)

large_world_coordinates_single_precision_build webp

### Using double-precision build

large_world_coordinates_double_precision_build webp


joao-pedro-braz:
About some core shaders failing to load, for me that only happens when using the Mobile rendered.
Forward+ works flawlessly.

moth-boi (approved):
I tested both Godot versions and looked over the code and this looks great to me.

PR #898, opened 2023-03-29
Improved art assets of 2d platformer

ValeS-Ramatak:
- Added new sprites and adjusted size and colors of already existing sprites:

image
image

- Modified this in the enemey and player script so they could flip at different sizes:
image

- Changed bullet velocity in the script.

- Changed screen resolution of project to match new sprite sizes.

This PR was sponsored by Ramatak with 💚


Calinou:
This is pretty good! However, this has the side effect of making the lack of integer scaling more obvious due to the higher base resolution. Therefore, I'd prefer to wait until https://github.com/godotengine/godot/pull/75784 is merged, then we can move along and merge this PR too (while enabling integer scaling in pixel art projects).

Calinou:
> @Calinou this targets the 3.x branch though, [godotengine/godot#75784](https://github.com/godotengine/godot/pull/75784) won't help with this.

Ah, I didn't notice that. In this case, I guess we will need to integrate https://github.com/Yukitty/godot-addon-integer_resolution_handler in the `3.x` branch of this demo.

ValeS-Ramatak:
Regarding point 1, I will make them all lowercase again if that's the preferred naming convention.

I scaled the project to 1920x1080 because that's the standard starting size I worked with on all projects so far were for console/mobile. However I am new to Godot and I surely lack understanding of how it works. I changed the camera value to be able to see more of the background, stretch mode 2d was the original mode before I added the new assets.

I'll change to the resolution you think it's best for this case to get the crisp aestethic.

Calinou:
> Perhaps, at the cost of making the camera a bit less "comfy", we could use something like 480x270? I don't know, as I said we should experiment further.

A common base resolution for PC 2D platformers in the late 90s is 640×480 (like Jazz Jackrabbit 2), though some players used 800×600.

ValeS-Ramatak:
> IIRC snake_case is the preferred convention. They _are_ lowercase in the filesystem, but for some reason the scenes are still pointing to files with an uppercase initial. Perhaps you renamed them outside of Godot?

I changed the names with the rename option within Godot, but I think I found where the problem with anamy and coin sprites was:
coinUppercase
coinLowercase


enemyUppercase
lowercaseEnemy

I reloaded the sprites so they now point to lowercase.



> For reference, a common base resolution for PC 2D platformers in the late 90s is 640×480 (like Jazz Jackrabbit 2), though some players used 800×600. For a slower-paced platformer, 512×384 might work out OK. 320×240 is definitely too cramped though.

The original resolution was this one:
image

and this is the original values in the camera:
image

Calinou:
> @ValeS-Ramatak IIRC for some reason the zoom works kinda backwards, so a 0.5 value should be equal to a 2x scale. Setting it to 1 should work fine with the new hi-res sprites.

Camera2D's `zoom` behavior was inverted in 4.0 (higher values are more zoomed in). However, in 3.x, lower values are more zoomed in.

Calinou:
Merged manually with https://github.com/godotengine/godot-demo-projects/commit/c17d803a1becb2b0fac47fa87d3572486fd77131, which includes some additional changes:

- All file names changed to `snake_case`.
- Increased player run speed, gravity and jump height for a better gameplay feel.
- Increased run animation speed to match the increased speed (and to make it look more "dynamic").

Issue #897, opened 2023-03-28
Adding different material roughness parameters for easy testing to 3d/material_tests


armmah:
The material tests scene demo has a range of materials currently on display for the demo, but it lacks a grid of abstract materials with specific increments for testing purposes of PBR.

While iterating on SSR I set up this grid that has both dielectrics and metals, and varying roughness value from 1.0 on the left to 0.0 on the right by a 0.1 step (simply duplicating the materials and setting the metal and roughness parameters on it).

![image](https://user-images.githubusercontent.com/31006916/228181726-55e939d3-0ced-4fcb-9b01-220b77994305.png)

No changes done to the demo itself, the new platform is outside of the camera frustum when going through the materials in the demo.
![image](https://user-images.githubusercontent.com/31006916/228182205-2ef7bdee-b16e-43cd-9785-433a37917540.png)

I already have this setup locally and if this can be useful I can open a PR.

Calinou:
https://github.com/godotengine/godot-demo-projects/pull/809 has a setup with all roughness and metallic levels, but I think it makes sense to add this to the Material Testers demo as well.

Feel free to open a pull request for this :slightly_smiling_face:

armmah:
`visible = false` added on a separate dedicated script, didn't want to introduce unncessary dependencies on the demo test script.

PR created https://github.com/godotengine/godot-demo-projects/pull/899

PR #899, opened 2023-03-29
Add editor-only roughness and metallic testers to Material Testers

armmah:
See https://github.com/godotengine/godot-demo-projects/issues/897#issue-1643489412

P.S. I edited the scene and created the script on engine version 4.1, but I don't think it should matter for the files I'm committing.

### DIFF
--- a/3d/material_testers/disable_on_startup.gd
+++ b/3d/material_testers/disable_on_startup.gd
@@ -1,0 +1,7 @@
+extends Node
+
+func _ready():
+	var optionalNode = self;
+	print(optionalNode.name);
+	optionalNode.set_process(false);
+	optionalNode.visible = false;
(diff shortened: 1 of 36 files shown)


3d/material_testers/disable_on_startup.gd
```diff
@@ -0,0 +1,7 @@
+extends Node
+
+func _ready():
+	var optionalNode = self;
+	print(optionalNode.name);
+	optionalNode.set_process(false);
+	optionalNode.visible = false;
```
Calinou, line 7: This can be simplified a lot. Also, follow the GDScript style guide: https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html

```suggestion
extends Node

func _ready():
	visible = false
```

Calinou (approved):
Thanks!

armmah:
Ah I even had some debug stuff in there, sorry and thanks for clearing out the mess

PR #926, opened 2023-06-26
Add tags to all demo projects

Calinou:
This makes sorting them in the project manager easier, as you can click tags in the project manager to filter to a specific tag.

Project tags are a new feature in 4.1, but this new setting won't break the demos when opened in 4.0.x.

## Preview

*We could improve the display of some tags by using the editor property capitalization when displaying tags in the project manager. This is a change that needs to be performed in Godot itself, not in this PR.*

Screenshot_20230626_175340


moth-boi:
It would be nice if the tags were sorted in some way instead of random.

aaronfranke (approved):

PR #928, opened 2023-07-05
Fix typo in Rich Text Label with BBCode

Calinou:
- Allow toggling pause by pressing the <kbd>Pause</kbd> key, in addition to the existing button.


akien-mga (approved):
Looks good.

akien-mga:
Thanks!

PR #516, opened 2020-08-29
Update translation demo project to include changes from GSoC 2020 i18n

SkyLucilfer:
## NOTE: This is for Godot 4.0

As part of my GSoC work on i18n, I have updated the translation demo project to incorporate the newly implemented features (CSV plurals, PO plurals, POT generation etc.)

I'm making a pull request first for mentors to review. They should have other plans.

PR #930, opened 2023-07-07
Update GUI Translation demo for 4.0

Calinou:
- Add PO support in addition to CSV.

Note that this PR does not include support for CSV plurals, as [that feature isn't merged in Godot yet](https://github.com/godotengine/godot/pull/41519).

___

- This salvages the work done in https://github.com/godotengine/godot-demo-projects/pull/516. Thanks @SkyLucilfer for the original PR :slightly_smiling_face:


Alex2782:
https://github.com/godotengine/godot-demo-projects/pull/974#pullrequestreview-1686715380

> Also, do we need to ship all the font with the demos? For the Mirroring and Translation demos system font fallback should be fine, neither require specific font dimensions to work.

I wanted to check if the 'fonts' can be removed here too, but it already says in the description that they are necessary on some systems.

@Calinou this PR works for you? And `master` branch not? https://github.com/godotengine/godot-demo-projects/pull/974#pullrequestreview-1683694780

- Label `HELLO_TEXT` has here normal 'droid_sans.tres' font
- Label `HELLO_TEXT` has in `master` no font (empty) and this errors (but I have no problems with the Japanese font.)


```
  Class 'Font' or its base class cannot be instantiated.
  editor/plugins/editor_preview_plugins.cpp:803 - Condition "sampled_font.is_null()" is true. Returning: Ref<Texture2D>()
```


image

aaronfranke (approved):
Looks good to me.

PR #935, opened 2023-07-18
Demo project for testing Saveload API

nlupugla:
Use this demo to test the Saveload API https://github.com/godotengine/godot/pull/79644 still currently in progress.


TheYellowArchitect:
idk what im doing wrong. ```git clone https://github.com/nlupugla/godot-demo-projects``` does not have any save directory inside loading directory, even though I see it on github.

Also it would be great if it was zipped, so any1 can download and use (like an MRP) instead of downloading all of godot's demo projects.

nlupugla:
Thanks for the interest!

I'll see if I can attach the demo as a zip to the feature PR. It probably isn't working because you need to clone my saveload_api branch.

TheYellowArchitect:
you are right, git clone defaults to master.

`git clone -b saveload_api https://github.com/nlupugla/godot-demo-projects`
The above works :)

PR #937, opened 2023-07-23
fsm: Map attack to gamepad X; revise text

idbrii:
Move attack map from gamepad Start to X so it's more discoverable.

Change label on state stack. "Pushown" was a typo for "pushdown", but understanding that requires reading a wikipedia page.

Add some periods to Explanations for consistency.


Calinou (approved):
Thanks!

PR #939, opened 2023-07-26
Add checks to the saving and loading demo to prevent deserializing Objects

jtnicholl:
`ConfigFile`'s `parse`/`load` methods and `str_to_var` can deserialize objects, and their `_init` methods execute immediately. This is something I've seen discussed online, but it's not well documented.
Checking that the text does not contain "Object(" before parsing is an easy way to prevent this and avoid malicious code.

### DIFF
--- a/loading/serialization/project.godot
+++ b/loading/serialization/project.godot
@@ -8,22 +8,6 @@
 
 config_version=5
 
-_global_script_classes=[{
-"base": "Node2D",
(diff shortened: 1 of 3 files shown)


jtnicholl:
I decided that I don't believe a check like this should be necessary and opened godotengine/godot#80562 instead.
If that issue is closed as not planned I'll reopen.

PR #943, opened 2023-08-23
Fix joypad demo for Godot 4.x

noidexe:
This fixes a number of issues preventing to use the remap feature:
- The gamepad images with the button layout were not visible
- Gamepad input was not being detected while the remap window was focused
- Some porting issues where `.pressed` appeared where `.button_pressed` was meant
- The wrong buttons being highlighted when remapping triggers
- The remap window was accepting events from all gamepads rather than only the one being remapped, which could be an issue if another connected controller has analog drift of is accidentally pressed for whatever reason.

~Input detection only works with native subwindows. I couldn't get it to work when using embedded subwindows. The subwindow seems to stop event propagation even with handle_input_locally disabled. You can workaround it by clicking outside the remap window before pressing a button but it should get a proper fix~
Checking the Godot source code it seems to be by design. Subwindows will always handle input when focused. Managed to fix it by connecting to Window.window_input

### DIFF
--- a/misc/joypads/joypads.tscn
+++ b/misc/joypads/joypads.tscn
@@ -46,6 +46,7 @@ corner_detail = 3
 
 [node name="JoypadsDemo" type="Control"]
 layout_mode = 3
+anchors_preset = 15
 anchor_right = 1.0
 anchor_bottom = 1.0
 offset_left = 30.0
--- a/misc/joypads/joypads.tscn
+++ b/misc/joypads/joypads.tscn
@@ -44,17 +44,14 @@ corner_radius_bottom_right = 3
 corner_radius_bottom_left = 3
 corner_detail = 3
 
-[node name="joypads" type="Control"]
+[node name="JoypadsDemo" type="Control"]
 layout_mode = 3
-anchors_preset = 8
-anchor_left = 0.5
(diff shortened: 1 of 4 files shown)


Calinou:
When you first start the demo with a controller connected, it's possible to see pressed buttons/axes but the UI appears grayed out (and the controller name doesn't appear):

image

You need to change the SpinBox value back and forth for the UI to look correct:

image

noidexe:
> Tested locally. On 4.1.1, when you first start the demo with a controller connected, it's possible to see pressed buttons/axes but the UI appears grayed out (and the controller name doesn't appear):

I cannot reproduce that, at least on windows and with a couple of xinput and dinput gamepads. I don't have a PS5 one. 

Can you add `_on_joy_connection_changed(joypad, true)` to the for loop in joypads.gd::_ready ? Maybe that'll fix it.

Calinou:
I've just tested this again. It looks like it works as-is on 4.2.dev godotengine/godot@6758a7f8c but not on 4.1.1.stable.

aaronfranke (approved):
Thank you for this! Tested, works great. I pushed a few small tweaks to the UI layout and code style.

PR #945, opened 2023-09-07
[Squash The Creeps] Use `Basis.looking_at` to control facing

AThousandShips:
Not technically wrong as `Player` is in global space but easily confusing to new users if they keep using `position` for this method.

Related:
* https://github.com/godotengine/godot-docs/pull/7890

### DIFF
--- a/3d/squash_the_creeps/Player.gd
+++ b/3d/squash_the_creeps/Player.gd
@@ -26,7 +26,8 @@ func _physics_process(delta):
 	if direction != Vector3.ZERO:
 		# In the lines below, we turn the character when moving and make the animation play faster.
 		direction = direction.normalized()
-		$Pivot.look_at(position + direction, Vector3.UP)
+		# Setting the basis property will affect the rotation of the node.
+		$Pivot.basis = Basis.looking_at(direction)
 		$AnimationPlayer.speed_scale = 4
 	else:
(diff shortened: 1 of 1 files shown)


3d/squash_the_creeps/Player.gd
```diff
@@ -26,7 +26,7 @@ func _physics_process(delta):
 	if direction != Vector3.ZERO:
 		# In the lines below, we turn the character when moving and make the animation play faster.
 		direction = direction.normalized()
-		$Pivot.look_at(position + direction, Vector3.UP)
+		$Pivot.look_at(global_position + direction, Vector3.UP)
```
aaronfranke, line 29: Could be made even simpler, we don't need the position at all.

```suggestion
		$Pivot.basis = Basis.looking_at(direction, Vector3.UP)
```

On one hand it could be argued that Basis is too complicated for beginners, but on the other hand I think if users are going to be using 3D in Godot they should learn what Basis is at some point early on. We could maybe add a comment of `# Setting the basis property will affect the rotation of the node.`

AThousandShips: True, will check it and do the relevant changes in the docs too, as this was inspired by that

aaronfranke (approved):

Calinou (approved):

PR #949, opened 2023-09-17
fix: mic_record script a bit for mac

swarnimarun:
Realised this demo was a bit broken while testing something irrelevant to the changes.

Also using all 3 names as iirc, OSX was what we used in 3.x. Though TBH not sure been a while since I used Godot 3. And I am not sure about the capitalisation part.

### DIFF
--- a/audio/mic_record/MicRecord.gd
+++ b/audio/mic_record/MicRecord.gd
@@ -1,7 +1,7 @@
 extends Control
 
 var effect  # See AudioEffect in docs
-var recording  # See AudioStreamSample in docs
+var recording  # See AudioStreamWAV in docs
 
 var stereo := true
 var mix_rate := 44100  # This is the default mix rate on recordings
@@ -93,4 +93,8 @@ func _on_StereoCheckButton_toggled(button_pressed: bool) -> void:
 
 
 func _on_open_user_folder_button_pressed():
-	OS.shell_open(ProjectSettings.globalize_path("user://"))
+	match OS.get_name():
+		"macOS", "MacOS", "OSX":
+			OS.shell_open("file://" + ProjectSettings.globalize_path("user://"))
+		_:
+			OS.shell_open(ProjectSettings.globalize_path("user://"))
--- a/audio/mic_record/MicRecord.gd
+++ b/audio/mic_record/MicRecord.gd
@@ -94,7 +94,7 @@ func _on_StereoCheckButton_toggled(button_pressed: bool) -> void:
 
 func _on_open_user_folder_button_pressed():
 	match OS.get_name():
-		"macOS", "MacOS", "OSX":
+		"macOS":
 			OS.shell_open("file://" + ProjectSettings.globalize_path("user://"))
 		_:
 			OS.shell_open(ProjectSettings.globalize_path("user://"))


audio/mic_record/MicRecord.gd
```diff
@@ -93,4 +93,8 @@ func _on_StereoCheckButton_toggled(button_pressed: bool) -> void:
 
 
 func _on_open_user_folder_button_pressed():
-	OS.shell_open(ProjectSettings.globalize_path("user://"))
+	match OS.get_name():
+		"macOS", "MacOS", "OSX":
```
AThousandShips, line 97: Only `macOS` is a valid name from `OS` here in 4.x AFAIK

AThousandShips:
What is the error you're seeing? Please provide some issue report for the reason behind this change

swarnimarun:
Aah, sorry, somehow thought it was obvious. So if you don't use `file://` with shell open on mac you will get -50 perm error, attached an image example below. 
image

audio/mic_record/MicRecord.gd
swarnimarun, line 97:
```suggestion
		"macOS":
```

AThousandShips:
That sounds like a bug to me, please open a report [here](https://github.com/godotengine/godot/issues)

audio/mic_record/MicRecord.gd
```diff
@@ -93,4 +93,8 @@ func _on_StereoCheckButton_toggled(button_pressed: bool) -> void:
 
 
 func _on_open_user_folder_button_pressed():
-	OS.shell_open(ProjectSettings.globalize_path("user://"))
+	match OS.get_name():
+		"macOS":
+			OS.shell_open("file://" + ProjectSettings.globalize_path("user://"))
+		_:
+			OS.shell_open(ProjectSettings.globalize_path("user://"))
```
swarnimarun, line 100:
```suggestion
	if OS.get_name() == "macOS":
		OS.shell_open("file://" + ProjectSettings.globalize_path("user://"))
	else:
		OS.shell_open(ProjectSettings.globalize_path("user://"))
```

swarnimarun: @AThousandShips lmk if this would be better now that we only have one matcher

swarnimarun:
@AThousandShips yep this is only specific to MacOS where the resource URI is required for correct call.

And there is already an issue for it, https://github.com/godotengine/godot/issues/52828, there was one even older than this, I haven't been involved in Godot engine dev for a couple years so not sure why it hasn't been fixed but it should not be a big change, I could document a fix for engine there but I think the existing comments already make the issue quite clear.

AThousandShips:
Okay! I asked for an issue before and got no answer so that's why I was confused 🙃

Calinou:
@bruvzg Can we make it so we don't need to manually prefix `file://` on macOS by modifying the engine core?

bruvzg:
> Can we make it so we don't need to manually prefix file:// on macOS by modifying the engine's OS::shell_open() implementation?
>
> This would prevent the need for this check in the project so the method can work in a cross-platform manner.

What's the behavior on other platforms? It `file://` is not required on Linux/Windows, I guess we can assume it's a file and auto add it if there's no URI schema in the `shell_open` argument.

bruvzg:
> Can we make it so we don't need to manually prefix file:// on macOS by modifying the engine's OS::shell_open() implementation?

https://github.com/godotengine/godot/pull/88126

aaronfranke:
Judging from the discussion, it seems that this PR is no longer necessary as of https://github.com/godotengine/godot/pull/88126. Closing.

PR #954, opened 2023-09-20
Add dynamic TileMap layer demo

KoBeWi:
I've seen some people asking for a way to disable tile collisions without using alternate tiles. It's possible, but not obvious, so I decided to make a demo. It's based on kinematic character demo.

https://github.com/godotengine/godot-demo-projects/assets/2223172/1b08bb86-6b8b-4207-b622-eadfcaf754a0

(maybe the demo needs a better name to make it more obvious what is it about, idk)


Calinou (approved):

PR #958, opened 2023-09-21
Physics-Based Platformer 2D, fixed "Debugger Errors"

Alex2782:
Edited with Godot 4.1.1, only `Scene: Enemy`. The other files Godot has changed automatically.

The 2 'sound nodes' cause the 'can_process' error output. 

Bildschirmfoto 2023-09-21 um 20 22 56

... But if they are moved up, then no more, maybe because of `VisibleOnScreenEnabler2D`.

Bildschirmfoto 2023-09-21 um 20 27 46


Calinou (approved):

Thanks!

PR #967, opened 2023-09-23
Fixed a grammatical typo

VedantMnkr:

### DIFF
--- a/mobile/README.md
+++ b/mobile/README.md
@@ -1,6 +1,6 @@
 # Mobile Demos
 
-These demos depend on features only available in mobile phones or tablets.
+These demos depend on features only available on mobile phones or tablets.
 
 The Android IAP demo only runs on Android.
 


Calinou:
I'd say the current grammatical form, while slightly unusual, is correct.

VedantMnkr:
> I'd say the current grammatical form, while slightly unusual, is correct.

I see.

aaronfranke:
The change in this PR is better, but it's incredibly minor, so it's not something we'd want to merge on its own. We can include it as a minor fix in another PR, or maybe a PR fixing many typos.

VedantMnkr:
> The change in this PR is better, but it's incredibly minor, so it's not something we'd want to merge on its own. We can include it as a minor fix in another PR, or maybe a PR fixing many typos.

Ok

VedantMnkr:
btw, I'm new to this so should I close the pull request or let it be as it is.

Calinou:
> btw, I'm new to this so should I close the pull request or let it be as it is.

I think it's better to close this for now. You can open another PR if you find a few more *real* typos to fix in other files.

PR #955, opened 2023-09-20
'Navigation Polygon 2D' ported to Godot 4.1.1

Alex2782:
I have combined the script from the documentation with the old solution (click event with red dot)
https://docs.godotengine.org/en/latest/tutorials/navigation/navigation_introduction_2d.html#setup-for-2d-scene

https://github.com/godotengine/godot-demo-projects/assets/41921395/a21e9ab1-43e1-4c0b-820a-563a8ee23264

'NavigationPolygon' drawn again, it looked buggy.
image

PR #974, opened 2023-10-05
UI Mirroring Demo: ported to Godot 4.1

Alex2782:
v4.2.dev5.official [e3e2528ba]

https://github.com/godotengine/godot-demo-projects/assets/41921395/f38e6581-fef6-41ad-ae54-37b1d586b99a


Calinou (changes requested):
Tested locally on 4.2.beta1. I can't see Japanese characters where they should appear in the translation demo:

image

This also applies to the UI mirroring demo:

image

Calinou:
I suggest checking what https://github.com/godotengine/godot-demo-projects/pull/930 does.

Alex2782:
I opened my Godot projects that were targeted at 4.2 with version 4.1.2. No problems. I have noticed some problems with other projects as well. Errors were issued with the files in the `imported` folder. e.g. the font resources I had to import again.

my `.godot` folder: [godot.zip](https://github.com/godotengine/godot-demo-projects/files/13030461/godot.zip)
@Calinou does it work after that if you use the whole or only `imported` folder with you?

image


**Godot v4.1.2.stable** - macOS 13.6.0 - Vulkan (Forward+) - integrated Apple M1 - Apple M1 (8 Threads)
`UI Mirroring Demo` also with **v4.2.beta1**.official [b1371806a]

image


image

Alex2782:
**.gitignore**

```.gitignore
# Godot 4+ specific ignores
.godot/

.....................

# Imported translations (automatically generated from CSV files)
*.translation
```

Alex2782:
`master` Godot project 4.0 after conversion to 4.1
Many errors are output, but I can start the project without errors.

image


**delete all ignore files**
```
rm -r .godot 
rm -r .import 
```

image

Alex2782:
If I close Godot and open the project again then everything is ok, Japanese is also displayed.

Calinou:
PS: I noticed your PR adds two new files in `ui_mirroring/fonts`, `noto_font.res` and `noto_font_arabic.res`. Why are these files required when the existing TTF fonts could be reused?

bruvzg (commented):
Fonts are clearly not ported correctly:

I would suggest doing something like this (from the clean state before any changes):
- Delete all font `*.tres` files (`droid_sans.tres`, `noto_font.tres`).
- Reimport all font from the editor import tab (probably optional).
- Set fallbacks in the main font import settings and reimport it.
Screenshot 2023-10-19 at 08 22 19

- Open the scene, and in the "Fix Dependencies" dialog, select the main font instead of missing font `*.tres`.

Also, do we need to ship all the font with the demos? For the Mirroring and Translation demos system font fallback should be fine, neither require specific font dimensions to work.

Alex2782:
> Open the scene, and in the "Fix Dependencies" dialog

I will continue later, without "Fix Dependecies" Godot crashes now. I do not know the function "Fix Dependecies" yet.

image

Alex2782:
> * Reimport all font from the editor import tab (probably optional).

I had opened `Project Settings` -> `Import Defaults` -> `Importer "Font Data (Dynamic Font)"` and under `"Fallbacks"` inserted and saved the 2 font files, after that I could not continue (Crash / Godot open Project Manager). Was that correct? 

The file '`ui_mirroring.tscn`' is now 7.3 MB. If I add font for Japanese, then the scene will be even bigger. 

@smix8 says: https://github.com/godotengine/godot-demo-projects/pull/955#issuecomment-1729159005

> If we rework and update the demo maybe we should save the NavigationPolygon resource to an external file a**s a binary .res file?**
> 
> Right now it is stored embedded as plain text inside the navigation.tscn file, something that we should not teach to new users as it will bloat their scene files and make them super slow to load/save/update on large and complex navigation meshes.

@bruvzg does this PR work for you with binary 'res'-Fonts? Or is Japanese not displayed for you either? 
The label for `HELLO_TEXT` has normal English font, but still Arabic is displayed correctly? @Calinou 

> I suggest checking what #930 does.
> Add PO support in addition to CSV.

I think there is little point in trying to fix problems I can't reproduce.

bruvzg:
> does this PR work for you with binary 'res'-Fonts?

There should not be any binary (or text) font resources at all, font files should be used directly.

bruvzg:
Something like this - https://github.com/bruvzg/godot-demo-projects/commit/0401f2871460370ba487d8506990aec5b5d43c09

Alex2782:
ok I'll check later, from some tutorials I know only first convert the fonts to "tres", also already with Godot 3.2

Calinou:
> Something like this - [bruvzg@0401f28](https://github.com/bruvzg/godot-demo-projects/commit/0401f2871460370ba487d8506990aec5b5d43c09)

I've tested that commit on 4.2.beta2 but I still can't see the CJK text:

image

The UI mirroring demo update works well though.

PR #976, opened 2023-10-08
Port 2D Skeleton demo to Godot 4.1

idbrii:
References #697.

Major changes:

* Recreate tilemap so it will display correctly. New setup uses terrains and autotile mostly works when painting.
* Change stretch, window, and camera settings to frame the world.
* Copy 2d/platformer character controller code.
* Use requests and correct ids with AnimationTree.
* Rework parallax background. I'm not familiar with using this node, but I got it working. The world isn't large so the camera doesn't move much, but that's not what we're showing off in this demo.

Screenshot:
Screenshot of world

Showing character animation (captured before I got parallax background setup):
Video of character animation


Calinou (commented):

I get some warnings when opening the project in the editor though:

```
WARNING: No Bone2D children of node Chin. Cannot calculate bone length or angle reliably.
Using transform rotation for bone angle
     at: calculate_length_and_rotation (scene/2d/skeleton_2d.cpp:451)
WARNING: No Bone2D children of node RightHand. Cannot calculate bone length or angle reliably.
Using transform rotation for bone angle
     at: calculate_length_and_rotation (scene/2d/skeleton_2d.cpp:451)
WARNING: No Bone2D children of node LeftHand. Cannot calculate bone length or angle reliably.
Using transform rotation for bone angle
     at: calculate_length_and_rotation (scene/2d/skeleton_2d.cpp:451)
WARNING: No Bone2D children of node LeftFoot. Cannot calculate bone length or angle reliably.
Using transform rotation for bone angle
     at: calculate_length_and_rotation (scene/2d/skeleton_2d.cpp:451)
WARNING: No Bone2D children of node RightFoot. Cannot calculate bone length or angle reliably.
Using transform rotation for bone angle
     at: calculate_length_and_rotation (scene/2d/skeleton_2d.cpp:451)
```

Could these be resolved without affecting functionality?

idbrii:
I'll have a look at those warnings. I didn't try to fix them yet.

idbrii:
Fixed those warnings. Seems that you need to manually disable the auto calc option on leaf nodes.

Output and Debugger > Errors have no warnings or errors and it still looks right to me:

https://github.com/godotengine/godot-demo-projects/assets/43559/b80d1277-609b-4d4d-8641-3e7ec8feeb47

idbrii:
Updated the readme to "Renderer: Forward+" even though I didn't make that change because looks like that's now true and other demo updates are making the same change. I guess GLES2 was the default before and the new default is Forward+ so no change in config files is visible?

Calinou:
> Updated the readme to "Renderer: Forward+" even though I didn't make that change because looks like that's now true and other demo updates are making the same change. I guess GLES2 was the default before and the new default is Forward+ so no change in config files is visible?

GLES3 is the default in 3.x, and Forward+ in 4.x.

That said, I suggest changing this demo to use Compatibility if it doesn't need features only available in Forward+. Remember to also do this for the `.mobile` override of the Rendering Method setting in the project settings dialog?

idbrii:
Seems fine in GL Compatibility. All three modes are using GL Compatibility now and rebased on master.

Calinou (approved):
Tested on 4.2.beta3, I don't get any error messages anymore.

Thanks!

PR #987, opened 2023-10-22
Multiplayer Bomber, fixes PhysicsRayQuery

Alex2782:
v4.1.2.stable.official [399c9dc39]

If the bomb and the player are at the same position, then `intersect_ray` returns no result.
`query.hit_from_inside = true`, fixes the problem. The other changes were made automatically by Godot.

https://github.com/godotengine/godot-demo-projects/assets/41921395/fa8123dc-3ba6-4e09-b6fa-4e023d728044


Calinou (approved):
Thanks!

PR #998, opened 2023-11-17
Add mobile controls and optimized settings to Truck Town

Calinou:
On mobile platforms, the vehicle automatically accelerates. Touch the left and right edges of the screen to steer. Touch the middle of the screen to brake/reverse (this also temporarily stops acceleration).

- Fix braking not working correctly (and make breaking stronger to make vehicles easier to control at high speeds).
- Allow using the D-Pad for steering and X/Square for braking.
- Tweak exhaust particles to be more subtle and fade more smoothly.
- Change spawn position to a location that's more straightforward to drive.

Tested on 4.1.3, so it should be fine to merge now.

## Preview

*Settings have been tweaked to achieve a constant 120 FPS on a Samsung Galaxy Z Fold4's (large) main display – before tweaks, it was roughly 60 FPS on the main display and 120 FPS on the cover display. This should ensure decent performance on mid-range devices that don't target such a high framerate, but additional testing is welcome.*

truck_town webp


akien-mga (approved):
Didn't test, but the described changes and code make sense to me.

akien-mga:
Thanks!

PR #870, opened 2023-03-02
Port dodge the creeps C# to Godot 4

van800:
I have used corresponding changes from https://github.com/godotengine/godot-demo-projects/pull/866

PR #1000, opened 2023-12-13
Updated Dodge the Creeps C# to Godot mono 4.2

captain-redbeard:
This change updates the C# Dodge the Creeps project to Godot 4.2 based on the updated documentation from here: https://docs.godotengine.org/en/stable/getting_started/first_2d_game/index.html

**Version**: Godot mono 4.2

### Changes
~~- Scene files changed to lowercase.~~
- Switched to explicit types as the documentation mentions that this should normally be used. **(Needs documentation update to match code)**
~~- Removed particle effect as it's not in the documentation.~~
- Removed GD.Randomize() from Main.cs as it's not in the documentation.
- Reuse ShowMessage() method in HUD.cs. **(Needs documentation update)**

### Possible issues
~~- Godot's editor defaults to tab indention, opening this project and saving any CS file will result in the indention being converted to tabs.~~
  ~~- Should we switch to tab indention to avoid this?~~


mono/dodge_the_creeps/HUD.tscn
AThousandShips, line 1: Why did you rename these? They are named the same in the non-mono version and should be kept this way IMO

AThousandShips:
See also:
* #870

mono/dodge_the_creeps/HUD.tscn
captain-redbeard: When I make a new scene with pascal or upper case and save it the name is changed to lowercase. This seems to be the standard?

AThousandShips: This is a bug AFAIK, you need to set the name manually, the example projects should follow TitleCase

AThousandShips: Also in this case modifying the file name *while* also rewriting it majorly detatches the history of it, making it harder to track the history of the file, the file is no longer linked to the old file, so keeping it in TitleCase for now helps preserve the history of it (if we later change all the file names but just that we get no loss of history)

Calinou:
> - Godot's editor defaults to tab indention, opening this project and saving any CS file will result in the indention being converted to tabs.
>
> - Should we switch to tab indention to avoid this?

Yes, I'd switch all C# files to use tabs. The docs use spaces because we have to use spaces for indentation there, even in GDScript.

> - Removed particle effect as it's not in the documentation.

The particle version is present in the GDScript version of the demo as well, so I'd prefer to keep it for consistency.

mono/dodge_the_creeps/Main.cs
```diff
@@ -2,76 +2,63 @@
...
-
-        GetNode<AudioStreamPlayer>("Music").Stop();
-        GetNode<AudioStreamPlayer>("DeathSound").Play();
-    }
-
-    public void NewGame()
+    private void NewGame()
     {
         // Note that for calling Godot-provided methods with strings,
         // we have to use the original Godot snake_case name.
```
AThousandShips, line 13: This comment should be updated to talk about using the names

Calinou:
@captain-redbeard Bump 🙂

mono/dodge_the_creeps/Main.cs
captain-redbeard: > This comment should be updated to talk about using the names

I'm not sure what you mean?

mono/dodge_the_creeps/Main.tscn
aaronfranke, line 1: You somehow managed to copy the entirety of "House In a Forest Loop.ogg" into Main.tscn, making it 6.6 MB instead of 2 KB.

mono/dodge_the_creeps/Mob.cs
```diff
@@ -1,17 +1,17 @@
...
-    {
-        var animSprite = GetNode<AnimatedSprite2D>("AnimatedSprite2D");
-        animSprite.Playing = true;
-        string[] mobTypes = animSprite.Frames.GetAnimationNames();
-        animSprite.Animation = mobTypes[GD.Randi() % mobTypes.Length];
-    }
+	// Called when the node enters the scene tree for the first time.
+	public override void _Ready()
+	{
+		AnimatedSprite2D animatedSprite2D = GetNode<AnimatedSprite2D>("AnimatedSprite2D");
```
aaronfranke: Per [the style guide](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_style_guide.html), the type can be omitted if it is already present on the same line. Since we have `GetNode<AnimatedSprite2D>`, the `var` keyword is acceptable here.

Per [the style guide](https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_style_guide.html), use spaces in C# files to avoid conflicts with existing C# tooling.

aaronfranke (approved):
Thank you for your work on this! I spent a few hours resolving the differences between this PR and #870, ensuring best practices, no regressions, and consistency with the tutorial in the documentation.

Ultimately what I ended up with was quite different from both PRs. I have pushed the consolidations to both PRs so that you can view them, but I can only accept one PR. As an arbitrary choice, I will be accepting this PR and closing PR #870. I will squash-and-merge and include @van800 as a co-author.

PR #1001, opened 2023-12-14
Add Squash the Creeps 3D C# demo for Godot mono 4.2

captain-redbeard:
This change adds the C# Squash the Creeps project based on the documentation from here: https://docs.godotengine.org/en/stable/getting_started/first_3d_game/index.html

**Version:** Godot mono 4.2

### Changes
- Scene files changed to lowercase. (Matches documentation, doesn't match gdscript demo)
- Switched to explicit types as the documentation mentions that this should normally be used. (Needs documentation update to match code)
- Added WorldEnvironment based on gdscript demo. (This would be useful in the documentation so people can end up with a product that matches the screenshots)

### Issues
There was an issue with the shadows on the player character when using the orthogonal projection on Camera3D. The solution was to turn off Generate LODs on the glib import as sourced from here: https://www.reddit.com/r/godot/comments/170wqon/weird_shadows_in_godot_3d_tutorial/
- I don't know enough about this issue to comment if the documentation should be updated or if this is a bug?


AThousandShips:
Same as on the 2D case, the file names need to be fixed to follow the naming

aaronfranke (approved):
Thank you for this! Since this is one of the two demos used as a tutorial in the docs, having a C# version of this is quite beneficial, so I agree that we should have it. I did a comparison of the code in this PR and that of the docs, and pushed some fixes to here. The code in the docs should be updated with a few fixes from here sometime, too.

PR #1003, opened 2023-12-16
Port Dynamic Tilemap Layers 2D demo to C# Godot mono 4.2

DevDaddyDownUnder:
This change ports the Dynamic Tilemap Layers 2D demo to C#.

**Version:** Godot mono 4.2


Calinou:
Is a C# port of this demo really justified? We try to keep C# ports to a minimum, as they tend to get out of sync compared to their GDScript counterpart over time. The logic seems relatively simple and doesn't feature too many things that would vary between GDScript and C#.

C#-specific demos are most suited for features that look very different between GDScript and C#, e.g. editor plugins.

PR #1006, opened 2023-12-24
Fix dodge the creeps vertical movement

coruja182:
using rotation in the player node instead of flipping vertically, by doing so the trail is also rotated


Calinou (approved):

Thanks!

PR #1013, opened 2024-01-29
Update demo files for Godot 4.2.1

Calinou:
All demos were opened with Godot 4.2.1, with the mesh format upgraded for 3D demos.

The 3D antialiasing demo now uses Extra Cull Margin on the last object to prevent it from disappearing too early when the camera rotates (due to its use of shader-based animation).

This is a prerequisite for uploading demos to the asset library, then https://github.com/godotengine/godot-demo-projects/issues/868 can be worked on.


aaronfranke (approved):
I didn't test this but I trust it implicitly, it's a simple noisy bulk update by a highly trusted contributor.

smix8 (approved):
Loaded most projects and did not encounter any project caused error.

All errors I encounter happen regularly in the stable version so not really a thing caused by this pr or the demo.

PR #1019, opened 2024-03-05
Fix squash_the_creeps hitboxes

Nazarwadim:
I added CollisionShapes for the player and the enemy to fix the bug of the enemy passing through the player.
image
image


Calinou (changes requested):
Tested locally, the collision shapes don't rotate when the player rotates (which happens when you move):

image

It works on enemies though.

Calinou (approved):

Thanks!

PR #1020, opened 2024-03-05
Fix player and mob animations not looping in Squash the Creeps

Calinou:
This was missed in the upgrade to 4.0, which reset loop mode properties in Animation.

### Before

https://github.com/godotengine/godot-demo-projects/assets/180032/6b4fab77-3dc1-40f0-a8a1-0668d740cadd

### After *(this PR)*

*Mob animation fix isn't shown on the video, as it was recorded before I pushed an update to this PR.*

https://github.com/godotengine/godot-demo-projects/assets/180032/eabd68b6-a048-4310-8f6f-6ac7c1a07ced


Nazarwadim (changes requested):
You should do this on mob too

Calinou:
> You should do this on mob too

Done. Thanks for the reminder :slightly_smiling_face:

Nazarwadim (approved):
Now the player makes a fiu fiu fiu when moves

PR #1023, opened 2024-03-24
Add Split Screen Demo showing input handling

Sauermann:
This demo shows:
- Single World2D, that is shared among many Viewports
- Simplified Input Map, that uses the same Actions for all Split Screens
- Input event routing to different viewports based on joypad device id and dedicated keyboard keys
- Dynamic keybinding adjustment for each Split Screen

Works with 4.2.1+.

- [x] Would appreciate advisement regarding the renderer: Which renderer should be used?

grafik


Calinou:
> Would appreciate advisement regarding the renderer: Which renderer should be used?

I'd say projects should always use the most conservative rendering method they can, so they work on the highest amount of devices possible out of the box. This means using Compatibility unless you need another rendering method.

Sauermann:
I have updated:
- Use compatibility renderer
- set deadzone to 0.2
- adjust an `elif`

When running with 4.2.1, now I get the warning:
```
WARNING: Glow is not supported when using the GL Compatibility backend yet. Support will be added in a future release.
```
How can I get rid of this warning or should I ignore it?

Calinou:
> When running with 4.2.1, now I get the warning:

You can ignore this warning, it's showing up only because of the 3D editor's preview environment (you won't see it when running the project). This will also not be an issue anymore in 4.3 as glow is implemented in the Compatibility rendering method there.

aaronfranke (approved):
Thank you! I have pushed some style fixes, layout improvements, and an update to Godot 4.5.

As for the assessment of whether or not we want the demo, I think it's important to have at least one demo showing World2D sharing, and the focus of the demo itself, split screen input, is also worth showing. So let's do it!

PR #1026, opened 2024-03-29
Correct font setup in Dodge The Creeps 2D.

chrisl8:
The fonts were not working in Dodge The Creeps 2D example and there was an error being generated.

This fixes the issue based on how the tutorial explains to set up fronts now
https://docs.godotengine.org/en/latest/getting_started/first_2d_game/06.heads_up_display.html#

I believe the `Xolonium-Regular.tres` file was causing this error:
```
Class 'Font' or its base class cannot be instantiated.
  editor/plugins/editor_preview_plugins.cpp:828 - Condition "sampled_font.is_null()" is true. Returning: Ref<Texture2D>()
```

Which appears in both Godot 4.2 and 4.3, although not as readily in 4.2.

The file does not seem to be necessary so I removed it.

### DIFF
--- a/2d/dodge_the_creeps/HUD.tscn
+++ b/2d/dodge_the_creeps/HUD.tscn
@@ -1,6 +1,7 @@
-[gd_scene load_steps=4 format=3 uid="uid://ccqoreueuxdb7"]
+[gd_scene load_steps=5 format=3 uid="uid://ccqoreueuxdb7"]
 
 [ext_resource type="Script" path="res://HUD.gd" id="1"]
+[ext_resource type="FontFile" uid="uid://cit6gwe5px1q8" path="res://fonts/Xolonium-Regular.ttf" id="2_2jm3i"]
 
 [sub_resource type="InputEventAction" id="InputEventAction_fopy7"]
 action = &"start_game"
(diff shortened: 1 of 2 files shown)


chrisl8:
It looks like this might be the point where many demos lost their fonts, and I can use this to figure out which fonts to put back:

https://github.com/godotengine/godot-demo-projects/commit/1d5184e23510ffcbbd95cab6c43b5073d1dc6eae

Does that sound right?

Calinou:
Font handling was remade from scratch for Godot 4.0, so you'll need to disregard the existing `.tres` font resources. 

Depending on your needs, there are 2 approaches:

- Use the font file directly, without an intermediate `.tres` resource.
- If you need font fallbacks (e.g. for CJK characters), you typically want to create a FontVariation resource, assign it a font and assign font fallbacks there. Save the resulting FontVariation to a `.tres` resource for easier reusability across scenes.

chrisl8:
Ok, well I think this matches with the current tutorial at https://docs.godotengine.org/en/latest/getting_started/first_2d_game/06.heads_up_display.html# so it seems like the correct fix to me. It makes the fonts actually work and clears out some errors on every startup of this demo with Godot 4+

Calinou (approved):
Tested locally, it works as expected. Thanks!

PR #1027, opened 2024-03-30
Update font code for Godot 4.0

chrisl8:
This clears out some errors that happen on every load of this demo in Godot 4 onward. It also makes the fonts actually work, as they did not before.

We probably don't strictly **need** these font settings for a demo, but they were here before.

I don't think `SourceCodePro-Black.ttf` was ever used, based on looking at history. So I removed it.

Both `.tres` files are gone too, as those are not used in Godot 4.

I have updated everything that used `SourceCodePro-Bold.ttf` to the new setup.

This is the same kind of update as https://github.com/godotengine/godot-demo-projects/pull/1026


Calinou (approved):

Thanks!

Issue #1004, opened 2023-12-18
In 3d/global_illumination, the "toggle_mouse_capture" does not work


joseph-sch:
**Which demo project is affected:**

`3d/global_illumination`

**OS/device including version:**

N/R

**Issue description:**

The overlay says "**Escape or F10: Toggle mouse capture**" but `"toggle_mouse_capture"` is not mapped to any of these keys in the project settings.

The issue is in the `project.godot` file:

https://github.com/godotengine/godot-demo-projects/blob/4a4b46cbe1f5105b664b99f325984096151b75a1/3d/global_illumination/project.godot#L70-L75

The two codes `16777217` and `16777253` correspond to `0x1000001` and `0x1000025` respectively, so the first one is not a valid mapping and the second one maps to `KEY_MASK_CMD_OR_CTRL + KEY_PERCENT` (which cannot be obtained because the percent key requires pressing Shift, unless I'm mistaken).

The correct lines should be, I assume, like in other 3D projects, e.g.:

https://github.com/godotengine/godot-demo-projects/blob/4a4b46cbe1f5105b664b99f325984096151b75a1/3d/waypoints/project.godot#L59-L64

There are other mappings that appear invalid (see highlighted in yellow in screenshot), but they are not critical.

**Screenshots of issue:**

![image](https://github.com/godotengine/godot-demo-projects/assets/11660030/63b2b7e0-0971-400a-9a8f-a44f9f77e4a7)

Calinou:
This is an issue caused by upgrading the project to Godot 4, which broke a lot of key bindings. Many other demos may have similar broken binding (I suggest grepping for the keycode value across all projects' `project.godot` files).

joseph-sch:
I checked all projects **which use the `toggle_mouse_capture`** and _that_ input is incorrectly mapped only in `3d/global_illumination`.

But indeed, there seem to be other input mapping issues in many `project.godot` files. For example, these all contain `16777217` or `16777218` i.e. `0x1000001` or `0x1000002`:

![image](https://github.com/godotengine/godot-demo-projects/assets/11660030/31bbec4d-c25f-4514-b1bf-f5e11bbd791a)

PR #1029, opened 2024-04-01
Fix incorrect toggle_mouse_capture key map in 3d/global_illumination

jsjtxietian:
I noticed https://github.com/godotengine/godot-demo-projects/issues/1004 too when playing with GI, fixes https://github.com/godotengine/godot-demo-projects/issues/1004


Calinou (approved):
Thanks!

PR #1030, opened 2024-04-01
Remove MSAA to fix #1028

### LINKED ISSUE #1028: WARNING: 2D MSAA is not yet supported for GLES3. in Dynamic TileMap Layers Demo due to setting to Compatibility
The Dynamic TileMap Layers demo has been set to `renderer/rendering_method="gl_compatibility"` however, it is also set to use MSAA 2D 4x (Slow) Anti Aliasing, resulting in this warning on every run:

`WARNING: 2D MSAA is not yet supported for GLES3.`

https://github.com/godotengine/godot-demo-projects/blob/31d1c0c1122717b3ad74aeb94255bc38def52747/2d/dynamic_tilemap_layers/project.godot#L67

Is this the intended behavior?

It seems like the demo should either be set to run in Forward+ by default or MSAA 2D should be turned off to avoid confusing demo users.
```[tasklist]
### Tasks
```

chrisl8:
Remove MSAA 2D 4x (Slow) Anti Aliasing, as that is not compatible with the switch to using Compatibility mode for the demo.

Fixes #1028

### DIFF
--- a/2d/dynamic_tilemap_layers/project.godot
+++ b/2d/dynamic_tilemap_layers/project.godot
@@ -64,4 +64,3 @@ common/physics_ticks_per_second=120
 renderer/rendering_method="gl_compatibility"
 renderer/rendering_method.mobile="gl_compatibility"
 environment/defaults/default_clear_color=Color(0.156863, 0.133333, 0.25098, 1)
-anti_aliasing/quality/msaa_2d=2


Calinou (approved):
Thanks!

This can be revisited in the future if https://github.com/godotengine/godot/issues/69462 is resolved.

PR #1037, opened 2024-04-07
Add Multiple resolutions with C# demo

JeremyRodgers:
C# version of the gui/multiple_resolutions demo project.
The concepts and code in this project are foundational for many applications and users would benefit from having a C# version available.
Not tested in 3.x.


aaronfranke:
Thank you for contributing, but we have decided to reject this PR. The arguement about being foundational doesn't matter, this is a demo that showcases engine functionality, and the API does not really change much betwen languages.

PR #1038, opened 2024-04-09
Switch to single safe threaded model in TextureRD demo

clayjohn:
Avoids the bug in https://github.com/godotengine/godot/issues/90363

Multithreaded rendering is not safe in 4.2, so it should not be used in our official demos


Calinou (approved):

Thanks!

PR #1039, opened 2024-04-09
Only print MIDI device list if any are found.

chrisl8:
This is a silly "fix", but it prevents an empty `[]` from being printed every time this demo is run without a MIDI device plugged into the computer.

The MIDI devices list is printed if one is plugged into the computer, which I have tested.


Calinou (approved):
Thanks!

Issue #1040, opened 2024-04-10
2.5d Demo is not fully ported from Godot 3.5 to 4.0


chrisl8:
**Which demo project is affected:**

[2.5D Demo with GDScript](https://github.com/godotengine/godot-demo-projects/tree/master/misc/2.5d)

**OS/device including version:**

Godot v4.3.dev (1f0f81049) - Windows 10.0.22621 - GLES3 (Compatibility) - NVIDIA GeForce RTX 3080 Ti (NVIDIA; 31.0.15.5186) - AMD Ryzen 9 5900X 12-Core Processor (24 Threads)

**Issue description:**

This project is not entirely ported from Godot 3.5 to 4.0.

In Godot 3.5, there is this extra editor screen that allows you to view the scene in various perspectives within the editor:
![image](https://github.com/godotengine/godot-demo-projects/assets/6188278/626e77a2-c692-4d05-893f-28d8a7731f84)

However, this is entirely missing in the 4.0 version. Further, there are quite a few calls to functions that don't exist anymore in 4.0.

```
  res://addons/node25d/main_screen/gizmo_25d.gd:9 - Parse Error: Could not find type "Node25D" in the current scope.
  res://addons/node25d/main_screen/gizmo_25d.gd:64 - Parse Error: Identifier "Node25D" not declared in the current scope.
  res://addons/node25d/main_screen/gizmo_25d.gd:73 - Parse Error: Identifier "Node25D" not declared in the current scope.
  res://addons/node25d/main_screen/gizmo_25d.gd:73 - Parse Error: Identifier "Node25D" not declared in the current scope.
  modules\gdscript\gdscript.cpp:2859 - Failed to load script "res://addons/node25d/main_screen/gizmo_25d.gd" with error "Parse error". (User)
  res://addons/node25d/main_screen/viewport_25d.gd:80 - Parse Error: Could not find type "Node25D" in the current scope.
  modules\gdscript\gdscript.cpp:2859 - Failed to load script "res://addons/node25d/main_screen/viewport_25d.gd" with error "Parse error". (User)
  Unable to open file: res://.godot/imported/node_25d_icon.png-2ad780313818706789bbb15408797db2.ctex.
  Failed loading resource: res://.godot/imported/node_25d_icon.png-2ad780313818706789bbb15408797db2.ctex. Make sure resources have been imported by opening the project in the editor at least once.
[...]

Issue #1016, opened 2024-02-06
2.5D Demo get_editor_viewport invalid call


awardell:
**Which demo project is affected:**
misc/2.5d

**OS/device including version:**
Godot v4.2.2.rc (991454b8b) - Windows 10.0.19045 - Vulkan (Forward+) - dedicated NVIDIA GeForce GTX 1080 Ti (NVIDIA; 31.0.15.3623) - AMD Ryzen Threadripper 1950X 16-Core Processor (32 Threads)

godot-demo-projects commit c423f3bf

**Issue description:**
When opening the demo using Godot 4.2.2, receive the following error:
`res://addons/node25d/node25d_plugin.gd:13 - Invalid call. Nonexistent function 'get_editor_viewport' in base 'EditorInterface'.`

Line 13 is
`get_editor_interface().get_editor_viewport().add_child(main_panel_instance)`

`get_editor_interface()` is deprecated and should be `EditorInterface`, but that isn't the issue.
The real issue is that `get_editor_viewport()` has been split into `get_editor_viewport_2d()` and `get_editor_viewport_3d()`
I'm unsure which is appropriate, but the fixed line would look something like

`EditorInterface.get_editor_viewport_3d().add_child(main_panel_instance)`

aaronfranke:
Fixed by #1041

PR #1041, opened 2024-04-10
Update 2.5D GDScript demo to display the editor plugin in Godot 4.

chrisl8:
This works now in Godot 4, but I still get this error when it starts:

```
  .\scene/main/node.h:445 - Parameter "data.tree" is null.
  res://addons/node25d/main_screen/viewport_25d.gd:23 - Invalid access to property or key 'process_frame' on a base object of type 'null instance'.
```

I'm not sure how to fix it though. I feel like there is something hacky about this line, but I'm not sure what the correct solution is:

https://github.com/godotengine/godot-demo-projects/blob/d190e73ad3af67447bb43ebe1ae0bcd01152e149/misc/2.5d/addons/node25d/main_screen/viewport_25d.gd#L20-L22

This should fix:
https://github.com/godotengine/godot-demo-projects/issues/1040
and
https://github.com/godotengine/godot-demo-projects/issues/1016

I did hit the files I edited with the [gdformat](https://github.com/Scony/godot-gdscript-toolkit) as well. If you prefer the changes without white-space updates I can remove those changes.

UPDATE: The error mentioned above **only** seems to happen if you open the `main_screen_25d` scene in the editor, then leave it open, and close and open the editor again.
image

### DIFF
--- a/misc/2.5d/addons/node25d/main_screen/gizmo_25d.gd
+++ b/misc/2.5d/addons/node25d/main_screen/gizmo_25d.gd
@@ -22,11 +22,12 @@ var dominant_axis
 @onready var lines_root = $Lines
 @onready var lines = [$Lines/X, $Lines/Y, $Lines/Z]
 
+
 func _process(_delta):
 	if not lines:
-		return # Somehow this node hasn't been set up yet.
+		return  # Somehow this node hasn't been set up yet.
 	if not node_25d:
-		return # We're most likely viewing the Gizmo25D scene.
+		return  # We're most likely viewing the Gizmo25D scene.
 	# While getting the mouse position works in any viewport, it doesn't do
 	# anything significant unless the mouse is in the 2.5D viewport.
 	var mouse_position = get_local_mouse_position()
@@ -35,7 +36,7 @@ func _process(_delta):
 		var closest_distance = 20.0
 		dominant_axis = -1
 		for i in range(3):
-			lines[i].modulate.a = 0.8 # Unrelated, but needs a loop too.
+			lines[i].modulate.a = 0.8  # Unrelated, but needs a loop too.
 			var distance = _distance_to_segment_at_index(i, mouse_position)
 			if distance < closest_distance:
 				closest_distance = distance
@@ -70,7 +71,9 @@ func _process(_delta):
 		# Make sure the gizmo is located at the object.
 		global_position = node_25d.global_position
 		if ROUGHLY_ROUND_TO_PIXELS:
-			spatial_node.transform.origin = (spatial_node.transform.origin * Node25D.SCALE).round() / Node25D.SCALE
+			spatial_node.transform.origin = (
+				(spatial_node.transform.origin * Node25D.SCALE).round() / Node25D.SCALE
+			)
 	# Move the gizmo lines appropriately.
 	lines_root.global_position = node_25d.global_position
 	node_25d.property_list_changed_notify()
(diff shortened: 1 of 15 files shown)


misc/2.5d/addons/node25d/main_screen/gizmo_25d.gd
```diff
@@ -70,7 +71,9 @@ func _process(_delta):
 		# Make sure the gizmo is located at the object.
 		global_position = node_25d.global_position
 		if ROUGHLY_ROUND_TO_PIXELS:
-			spatial_node.transform.origin = (spatial_node.transform.origin * Node25D.SCALE).round() / Node25D.SCALE
+			spatial_node.transform.origin = (
+				(spatial_node.transform.origin * Node25D.SCALE).round() / Node25D.SCALE
+			)
```
aaronfranke, line 76: Would breaking the line up like this be sufficient to please the formatter?

```suggestion
			var pixel_pos: Vector3 = (spatial_node.transform.origin * Node25D.SCALE).round()
			spatial_node.transform.origin = pixel_pos / Node25D.SCALE
```

misc/2.5d/addons/node25d/main_screen/viewport_25d.gd
```diff
@@ -123,14 +126,15 @@ func _gui_input(event):
 
 
 func _recursive_change_view_mode(current_node):
-	if current_node.has_method("set_view_mode"):
-		current_node.set_view_mode(view_mode_index)
-	for child in current_node.get_children():
-		_recursive_change_view_mode(child)
+	if current_node:
```
aaronfranke, line 129: Prefer using guard clauses.

```suggestion
	if not current_node:
		return
```

misc/2.5d/addons/node25d/node25d_plugin.gd
```diff
@@ -5,43 +5,53 @@ const MainPanel = preload("res://addons/node25d/main_screen/main_screen_25d.tscn
...
-	add_custom_type("ShadowMath25D", "CharacterBody3D", preload("shadow_math_25d.gd"), preload("icons/shadow_math_25d_icon.png"))
+	add_custom_type(
+		"YSort25D", "Node", preload("y_sort_25d.gd"), preload("icons/y_sort_25d_icon.png")
+	)
+	add_custom_type(
+		"ShadowMath25D",
+		"CharacterBody3D",
+		preload("shadow_math_25d.gd"),
+		preload("icons/shadow_math_25d_icon.png")
+	)
```
aaronfranke, line 28: IMO this part hurts readability, I would undo the formatter's changes here.

misc/2.5d/addons/node25d/main_screen/gizmo_25d.gd
chrisl8: I just undid the formatting for these lines to put it back the way it was, if that is OK with you.

misc/2.5d/addons/node25d/node25d_plugin.gd
chrisl8: I just undid the formatting for these lines to put it back the way it was, if that is OK with you.

misc/2.5d/addons/node25d/main_screen/viewport_25d.gd
chrisl8: Done.

chrisl8:
@aaronfranke  and @Calinou  Let me know if you need me to change anything else. Also if you need this squashed.

aaronfranke (approved):
Thanks for working on this! I pushed a few more changes to fully fix the shadow, fix the gizmo, and generally clean up the code.

Calinou:
Note that I fixed incorrect static typing in a follow-up commit: https://github.com/godotengine/godot-demo-projects/commit/7223aec0012ba5f89b7f9f28a65e91abb30be0cc

PR #1049, opened 2024-04-18
change upward movement key to W instead of Z

Goldenlion5648:


Calinou (approved):

Thanks!

PR #1052, opened 2024-05-07
Fixed minor Godot 3 bit in "gui_in_3d" demo's billboard specific code

Tarik-B:
- Godot 3's SpatialMaterial has a "params_billboard_mode" property:
https://docs.godotengine.org/en/3.6/classes/class_spatialmaterial.html#class-spatialmaterial-property-params-billboard-mode
- but Godot 4's BaseMaterial3D instead has a "billboard_mode" property:
https://docs.godotengine.org/en/stable/classes/class_basematerial3d.html#class-basematerial3d-property-billboard-mode

But I guess nobody needed any convincing since that change has already been made at line 20:
`if node_quad.get_surface_override_material(0).billboard_mode == BaseMaterial3D.BillboardMode.BILLBOARD_DISABLED:`

### DIFF
--- a/viewport/gui_in_3d/gui_3d.gd
+++ b/viewport/gui_in_3d/gui_3d.gd
@@ -110,7 +110,7 @@ func _mouse_input_event(_camera: Camera3D, event: InputEvent, event_position: Ve
 
 
 func rotate_area_to_billboard():
-	var billboard_mode = node_quad.get_surface_override_material(0).params_billboard_mode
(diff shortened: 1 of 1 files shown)


aaronfranke (approved):
Thanks, good catch! Didn't test but I assume this works.

Tarik-B:
> Thanks, good catch! Didn't test but I assume this works.

Well I did test it since, as you can imagine, I stumbled upon it while trying the demo.

PR #1054, opened 2024-05-13
Remove module and albedo_mix changes from Fade Decal example.

chrisl8:
The way modulate works on decals changed for Godot 4.3. See: https://github.com/godotengine/godot/pull/89849

This caused the decals in the fade example to look bad. See: https://github.com/godotengine/godot/issues/91269

I don't see the point in either of these settings for this example, as the "fade" works fine with the decals without additional adjustments to them.

My suggestion, as provided in this PR, is to simply drop the modulate and albedo_mix adjustments. You won't hurt my feelings if you reject this and do something different though.


Calinou (approved):

Thanks!

PR #1057, opened 2024-05-20
Fixed bug in pong demo where ball does not reset to center.

grimatoma:
The initial position is set as the vector2 of Position but Position is a pointer so it updates meaning the initial position is the current position of the ball

### DIFF
--- a/mono/pong/Logic/Ball.cs
+++ b/mono/pong/Logic/Ball.cs
@@ -11,7 +11,7 @@ public partial class Ball : Area2D
 
     public override void _Ready()
     {
-        _initialPos = Position;
+        _initialPos =  new Vector2(Position.X, Position.Y);
     }
 
     public override void _Process(double delta)


mono/pong/Logic/Ball.cs
```diff
@@ -11,7 +11,7 @@ public partial class Ball : Area2D
 
     public override void _Ready()
     {
-        _initialPos = Position;
+        _initialPos =  new Vector2(Position.X, Position.Y);
```
AThousandShips, line 14:
```suggestion
        _initialPos = new Vector2(Position.X, Position.Y);
```

paulloz:
`Vector2` is a `struct`, and as such is copied on assignment. So both of those lines will yield the same result:
```cs
_initialPos = Position;
_initialPos = new Vector2(Position.X, Position.Y);
```

PR #1063, opened 2024-05-31
Use static typing in all demos

Calinou:
This leads to code that is easier to understand and runs faster thanks to GDScript's typed instructions.

The untyped declaration warning is now enabled on all projects where type hints were added. All projects currently run without any untyped declaration warnings (other script warnings have been fixed in this PR as well).

Dodge the Creeps and Squash the Creeps demos intentionally don't use type hints to match the documentation, where type hints haven't been adopted yet (given its beginner focus).

Various small visual and usability tweaks were made to many of the demos in the process.

This PR was tested on 4.2.2 (with the exception of the 2.5D demo which requires 4.3 to run), so it can be merged without breaking compatibility.

This closes https://github.com/godotengine/godot-demo-projects/issues/868.


smix8:
My question is if `:=` should really be used that often in demo projects.

Yes sometimes it is obvious what the type is, sometimes you need to scroll up, sometimes you even need to look at the documentation because you can't trust a function name's return value.

In general I prefer for tutorials and demos to have the actual type writen next to it even if it makes the scripts slightly more bloated. It removes any ambiguity, adds color highlights and spacing in the script editor, and all that makes the code reading so much easier.

smix8 (approved):
Tested the demos and works fine.

aaronfranke (approved):

adamscott:
> My question is if `:=` should really be used that often in demo projects.
> 
> Yes sometimes it is obvious what the type is, sometimes you need to scroll up, sometimes you even need to look at the documentation because you can't trust a function name's return value.
> 
> In general I prefer for tutorials and demos to have the actual type writen next to it even if it makes the scripts slightly more bloated. It removes any ambiguity, adds color highlights and spacing in the script editor, and all that makes the code reading so much easier.

This is a script editor issue, not showing the typing, not the fault of `:=`.

2d/pong/logic/ball.gd
```diff
@@ -2,17 +2,17 @@ extends Area2D
...
+var _speed := DEFAULT_SPEED
+var direction := Vector2.LEFT
 
-@onready var _initial_pos = position
+@onready var _initial_pos := position
 
-func _process(delta):
-	_speed += delta * 2
+func _process(delta: float) -> void:
+	_speed += int(delta * 2)
```
aaronfranke, line 11: @Calinou This completely breaks the Pong demo, it causes the speed to never increase.

We need to change `const DEFAULT_SPEED = 100` to `100.0` since this value is always used in a float context, and change all the types to floats.

In general this is why I prefer explicitly writing `: int` and `: float` instead of inferring int/float. This is what the style guide says, but it is not what this PR does.

Calinou: Good catch :slightly_smiling_face: I've opened a PR to fix the issue: https://github.com/godotengine/godot-demo-projects/pull/1073

PR #1064, opened 2024-06-03
Fix class name and extends

OursCodeur:
Pretty straightforward - those were swapped during conversion to strict typing and caused the 3d navigation demo to crash on startup.

### DIFF
--- a/3d/navigation/line3d.gd
+++ b/3d/navigation/line3d.gd
@@ -1,5 +1,5 @@
-class_name MeshInstance3D
-extends Line3D
+class_name Line3D
+extends MeshInstance3D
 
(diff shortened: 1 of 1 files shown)


Calinou:
Fixed manually in https://github.com/godotengine/godot-demo-projects/commit/9a0c857131b8561257d3c289a0b76a611c10dbcc, as there were several other instances of this issue. This was due to an issue in the `sd` command I used in https://github.com/godotengine/godot-demo-projects/pull/1063 to swap the order of `class_name` and `extends` to follow the GDScript style guide.

Thanks for the contribution nonetheless :slightly_smiling_face:

PR #1065, opened 2024-06-03
Fix class name and extends

OursCodeur:
Those were swapped during conversion to strict typing and caused the 2.5d demo to crash on startup.

- *Production edit: Follow-up to https://github.com/godotengine/godot-demo-projects/pull/1064.*


Calinou:
Fixed manually in https://github.com/godotengine/godot-demo-projects/commit/9a0c857131b8561257d3c289a0b76a611c10dbcc, as there were several other instances of this issues. This was due to an issue in the `sd` command I used in https://github.com/godotengine/godot-demo-projects/pull/1063 to swap the order of `class_name` and `extends` to follow the GDScript style guide.

Thanks for the contribution nonetheless :slightly_smiling_face:

PR #1068, opened 2024-06-05
Improve Material Testers demo project

Calinou:
- Add reflection probes for each test sphere, which can be toggled using a new dropdown with 3 options (no reflection probe, reflection only, reflection + ambient). The cull and reflection masks are configured on each reflection probe so that spheres don't self-reflect, but other spheres still appear in each sphere's reflections. Probe extents are made very large to prevent any visible fading, and the test beds don't receive the reflection probe's ambient or reflection.
- Improve ice material visuals by giving it some opacity.
- Use lossless compression on all textures, as their relatively low resolution and count by today's standards makes them not use much VRAM anyway.

Don't cherry-pick to 4.2, as this will only work in 4.3 since this requires https://github.com/godotengine/godot/pull/86073 to look correct.

## Preview

No Reflection Probes

Reflection Probes (Reflection only)

Reflection Probes (Reflection + Ambient)


BastiaanOlij:
Need to find some time to test this but from the screenshots this looks fantastic!

aaronfranke (approved):
Tested, looks good.

PR #1069, opened 2024-06-06
[3.x] Enable physics interpolation in all demos that use physics

Calinou:
This also improves visuals in the Voxel and Dynamic Split Screen demos, and fixes splitscreen mode in the Platformer 2D demo.

Most demos (other than physics tests) have also had their physics FPS increased to 120. This is done consistency with 4.x, but also for the benefits it still provides when interpolation are enabled (lower input lag, more stable simulation).

This can be cherry-picked to 3.5, but the benefits of 2D physics interpolation will only be seen in 3.6.


aaronfranke (approved):

lawnjelly:
Just looking through some of the physics interpolation warnings:

### 3D platformer
* In the coin scene, you should change the `Animation` playback process mode to physics.

Calinou:
> - In the coin scene, you should change the `Animation` playback process mode to physics.


Done.

2d/skeleton/project.godot
```diff
@@ -82,7 +82,9 @@ limits/debugger_stdout/max_chars_per_second=1000000
 
 [physics]
 
+common/physics_fps=120
 2d/default_gravity=900
+common/physics_interpolation=true
```
lawnjelly, line 87: Seeing [#92872](https://github.com/godotengine/godot/issues/92872) it might be good to leave interpolation off for now for this project until the bug is fixed (can't say how long this will take as I have to re-familiarize myself with the 2D skinning which is quite complex).

Calinou: I re-enabled physics interpolation in the 2D skeleton demo now that https://github.com/godotengine/godot/issues/92872 is fixed in 3.x.

lawnjelly: > I re-enabled physics interpolation in the 2D skeleton demo now that https://github.com/godotengine/godot/issues/92872 is fixed in 3.x.

Sounds sensible, I thought it was going to be more difficult to fix than it turned out to be. :grin:

lawnjelly:
> This PR also benefits from https://github.com/godotengine/godot/pull/92469, as it relies on that PR to reduce glitching when a 3D project starts (although it's hardly visible at 120 TPS in the first place).

The auto-resets is now merged as part of https://github.com/godotengine/godot/pull/92784 so afaik all the functionality should be good now for the demos. :+1: 

It also might be worth having at least one slow moving demo at a lower tick rate, just to demo that way of working, which can be good for e.g. RPGs where responsiveness is not critical.

Also as an aside, even with 120tps, bear in mind that if rendering at 60fps, I'm not sure the input gets sampled more than 60fps (without agile input) on most platforms, so the responsiveness may only be seen with high refresh monitors.

Calinou:
I've tested the demos on 3.6.stable and they work as expected.

PR #1071, opened 2024-06-08
Update Mono Pong

Insane96:
## Scene
The scene had some problems:
All the collision node were twice as small as they should have. Not sure if it's due to an update in the engine.
image
So the collisions are now bigger (2x exactly) to cover the whole sprite, and I've changed the ball collision shape to a sphere.

## Ceiling and Floor
The ceiling was just transposing the angle of the ball by 90°, I've changed it to actually bounce back.

## Paddle
The paddles have been made faster in movement (+50%) and now bounce the ball back based off the position of the collision.

## Ball
Ball now speeds up faster and picks a new random direction and angle when the game starts or resets.

### DIFF
--- a/mono/pong/Logic/Ball.cs
+++ b/mono/pong/Logic/Ball.cs
@@ -1,10 +1,12 @@
+using System;
 using Godot;
 
 public partial class Ball : Area2D
 {
+    private readonly Random _random = new();
     private const int DefaultSpeed = 100;
 
-    public Vector2 direction = Vector2.Left;
+    public Vector2 direction;
 
     private Vector2 _initialPos;
     private double _speed = DefaultSpeed;
@@ -12,18 +14,24 @@ public partial class Ball : Area2D
     public override void _Ready()
     {
         _initialPos = Position;
+        direction = this.NewRandomDirection();
     }
 
     public override void _Process(double delta)
     {
-        _speed += delta * 2;
+        _speed += delta * 5;
         Position += (float)(_speed * delta) * direction;
     }
 
     public void Reset()
     {
-        direction = Vector2.Left;
+        direction = this.NewRandomDirection();
         Position = _initialPos;
         _speed = DefaultSpeed;
+    }
+
+    public Vector2 NewRandomDirection()
+    {
+        return new Vector2((float)(this._random.NextDouble() * 2f - 1f), (float)(this._random.NextDouble() * 2f - 1f)).Normalized();
     }
 }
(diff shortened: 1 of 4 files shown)


mono/pong/Logic/Paddle.cs
paulloz, line 1: Use 4 spaces (no tabs) to indent[^1].

[^1]: https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_style_guide.html#formatting

paulloz (changes requested):

Insane96:
Applied all request changes

mono/pong/Logic/Ball.cs
```diff
@@ -1,29 +1,37 @@
+using System;
 using Godot;
 
 public partial class Ball : Area2D
 {
+    private readonly Random _random = new();
```
aaronfranke, line 6: Target-typed `new` is only available in C# 9 or later. For the sake of keeping the demo's code compatible with older C# versions, we should not use it.

Keep in mind that even if Godot supports C# 9 or later, there are older Godot versions that do not. There is value in keeping the code consistent between all versions of the demo, even with the 3.x versions. That isn't to say that we don't want to improve things over time, just avoid unnecessary changes that don't affect behavior but do negatively affect compatibility.

aaronfranke (changes requested):
None of these changes can be considered or accepted unless the same changes are made to the GDScript version of the demo.

mono/pong/Logic/Ball.cs
```diff
@@ -1,29 +1,37 @@
...
-    public Vector2 direction = Vector2.Left;
+    public Vector2 Direction { get; set; }
 
     private Vector2 _initialPos;
     private double _speed = DefaultSpeed;
 
     public override void _Ready()
     {
         _initialPos = Position;
+        Direction = this.NewRandomDirection();
```
AThousandShips, line 17: Are all the added `this` really necessary? What are they disambiguating?

paulloz: Thanks, I missed those when I reviewed.

paulloz:
> None of these changes can be considered or accepted unless the same changes are made to the GDScript version of the demo.

The purely corrective parts (area sizes in the `.tscn`, and the typo fixed in `Paddle.cs`) could. But otherwise, yes.

AThousandShips:
Even those changes to the shapes don't match the gdscript version so they should match

PR #1091, opened 2024-07-17
Add FSR 2.2, exclusive fullscreen and ultra quality SSAO and SSIL to graphics settings demo

Ryhon0:
FSR 2.2 was added in 4.2 and this demo only has the option to use FSR 1.0, this adds an option to use it.
The exclusive fullscreen option had code to be enabled but not an item in the option button, this PR adds the missing item.     
Bloom had a redundant high option, which did the exact same thing as the low option. This PR renames low to enabled and removes high.  
SSAO and SSIL quality enums also have a value for ultra quality, which weren't used. This PR adds them as options.

### DIFF
--- a/3d/graphics_settings/control.tscn
+++ b/3d/graphics_settings/control.tscn
@@ -169,12 +169,14 @@ text = "Display Filter:"
 layout_mode = 2
 size_flags_horizontal = 3
 theme_override_font_sizes/font_size = 16
-item_count = 2
+item_count = 3
 selected = 0
 popup/item_0/text = "Bilinear (Fastest)"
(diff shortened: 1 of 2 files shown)


Calinou:
> Bloom had a redundant high option, which did the exact same thing as the low option.

These are intended to be different. Low has glow bicubic scaling disabled, High has it enabled.

Calinou:
Merged manually with https://github.com/godotengine/godot-demo-projects/commit/8ab921d5b4515bed1e3c681e24c7b2f3bc8ab420 (which rebases this PR to fix conflicts). Thanks!

PR #1099, opened 2024-08-05
Add navigation mesh chunks demos

smix8:
Adds 2D and 3D demo project for how to bake navigation meshes for large world chunk systems.

Note these demos require Godot 4.3+ with the new navigation mesh border size properties.

Intended as demo part for the updated 4.3 documentation here https://docs.godotengine.org/en/latest/tutorials/navigation/navigation_using_navigationmeshes.html#baking-navigation-mesh-chunks-for-large-worlds

I had to add convoluted parser code for the bound calculation in the demo as the PR that adds it to the core engine is not merged https://github.com/godotengine/godot/pull/95137. That is why there is a skip at the begin of the `calculate_source_geometry_bounds()` function if a `get_bounds()` function is found on the source geometry objec. The intention is that users see and remember to update this in their projects and that the demo uses the way faster parsing automatically should the PR be merged before the demo gets updated.

navigation_mesh_chunks


3d/navigation_mesh_chunks/navmesh_chhunks_demo_3d.gd
```diff
@@ -0,0 +1,149 @@
...
+			chunk_region.navigation_mesh = chunk_navmesh
+			chunks_root_node.add_child(chunk_region)
+
+			chunk_id_to_region[chunk_id] = chunk_region
+
+
+static func calculate_source_geometry_bounds(p_source_geometry: NavigationMeshSourceGeometryData3D) -> AABB:
+	if p_source_geometry.has_method("get_bounds"):
+		# Godot 4.3 Patch added get_bounds() function that does the same but faster.
+		return p_source_geometry.call("get_bounds")
```
Calinou, line 99: Is this check needed, considering the demo already needs Godot 4.3 to run?

2d/navigation_mesh_chunks/navmesh_chhunks_demo_2d.gd
```diff
@@ -0,0 +1,151 @@
...
+	var parse_settings: NavigationPolygon = NavigationPolygon.new()
+	parse_settings.parsed_geometry_type = NavigationPolygon.PARSED_GEOMETRY_STATIC_COLLIDERS
+	NavigationServer2D.parse_source_geometry_data(parse_settings, source_geometry, %ParseRootNode)
+
+	# Add an outline to define the traversable surface that the parsed collision shapes can "cut" into.
+	var traversable_outline: PackedVector2Array = PackedVector2Array([
+		Vector2(0.0,0.0),
+		Vector2(1920.0,0.0),
+		Vector2(1920.0,1080.0),
+		Vector2(0.0,1080.0),
```
Calinou, line 38:
```suggestion
		Vector2(0.0, 0.0),
		Vector2(1920.0, 0.0),
		Vector2(1920.0, 1080.0),
		Vector2(0.0, 1080.0),
```

Calinou (commented):

I suggest setting **MSAA 3D** project setting to 4x in the 3D demo (like we do in many other demos), so that it benefits from antialiasing. MSAA 2D isn't implemented in Compatibility yet, so this can't be done in the 2D demo currently.

3d/navigation_mesh_chunks/navmesh_chhunks_demo_3d.gd
smix8: The function does not exist yet in 4.3 as the linked PR that adds it is not merge, and properly will not before 4.3.1 as a patch.

That is why the calculate_source_geometry_bounds() function was added in the demo that does everything in scripts.

Calinou (approved):

PR #1100, opened 2024-08-10
Fix incorrect asset library link

abcSup:
The original link is incorrect and points to the [Tween Interpolation Demo](https://godotengine.org/asset-library/asset/2733)


aaronfranke (approved):

Calinou (approved):
Thanks!

PR #1102, opened 2024-08-13
Change semi-truck in Truck Town demo to use 6DOF joint

### LINKED ISSUE #1101: Truck Town Semi-Truck physics wonky when using Godot Jolt physics engine
`3d/truck_town`

Linux

**Issue description:**
When enabling the Godot Jolt addon, the semi-truck vehicle is essentially unplayable: the trailer is moved upwards off the ground at the hinge, and the truck is moved downwards off of its wheels. Might be an issue with the joint.

Works fine with default Godot Physics.

**Screenshots of issue:**

mihe:
Fixes #1101.

This demo is currently a bit broken when using the semi-truck, as the joint between the truck and the trailer uses a `ConeTwistJoint3D` with a `swing_span` and `twist_span` of 30 and 45 _radians_ respectively. I'm assuming this was meant to be degrees at some point, but something must have changed in Godot since this demo was made, where it no longer saves these properties as degrees. This only happened to work due to `ConeTwistJoint3D` supporting multiple revolutions along the twist axis in Godot/Bullet, but which is not something supported by Jolt.

This PR fixes that by changing over to a 6DOF joint instead, which I would argue is the more intuitive setup for this particular vehicle.

I also changed/improved the collision boxes for both the truck and the trailer, as the current setup seemed to cause some issues with Jolt even after changing the joint, due to the difference in how the `hit_back_faces` parameter works for `intersect_ray`.

### DIFF
--- a/3d/truck_town/vehicles/trailer_truck.tscn
+++ b/3d/truck_town/vehicles/trailer_truck.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=24 format=3 uid="uid://drinprblemj5u"]
+[gd_scene load_steps=27 format=3 uid="uid://drinprblemj5u"]
 
 [ext_resource type="Script" path="res://vehicles/vehicle.gd" id="1_wetfm"]
 [ext_resource type="ArrayMesh" uid="uid://bqrwin8ccgptt" path="res://vehicles/meshes/wheel.res" id="2_q28iu"]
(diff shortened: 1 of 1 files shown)


aaronfranke (approved):
Makes sense and works good in testing. Interestingly, it seems to work better even in Godot Physics.

PR #1107, opened 2024-08-26
updated meshes to OBJ instead of RES for 3d/navigation

shahriarlabib000:
Version:4.3.stable
- Updated meshes to .obj
- switched to compatibility renderer


shahriarlabib000:
Might be worth mentioning that it fixes some weird artifacts in the meshes 
### Before
Screenshot_2024-08-26-23-51-48-973_org godotengine editor v4
Screenshot_2024-08-26-23-52-07-691_org godotengine editor v4

### Now
Screenshot_2024-08-26-23-53-32-005_org godotengine editor v4
Screenshot_2024-08-26-23-53-44-369_org godotengine editor v4

aaronfranke (approved):
Looks good, thanks! Note that in general, glTF (.glb or .gltf) is the preferred mesh format in Godot, but for this very simple demo OBJ may be better just so that we have some of the demos showing OBJ.

PR #1110, opened 2024-08-27
Improve Truck Town model, fix issues, use OMI glTF Physics

aaronfranke:
This PR improves the Truck Town demo, fixing many issues, and making use of the updated OMI glTF physics in Godot 4.3. Now, instead of the physics information being defined in the `.import` file, instead the physics is defined inside of the `.gltf` file itself. This means that theoretically anyone can grab the Truck Town model and use it in another project or even another game engine without setting up all the physics again.

For the sake of allowing users to learn about the glTF extensions, I kept the model as a `.gltf` text file instead of a `.glb` binary file, so that anyone can open it in a text editor.

The model is now centered on the map, such that the outer ground mesh has a rectangle cut out of its center. I moved the CSG racetrack into a separate scene to preserve the source geometry in case someone wants to edit it, but I put it in the .glb file so that I could fix some issues with the textures (see below).

This PR also fixes several problems with the meshes and materials. I fixed a lot of cases of weird texture stretching. I fixed some of the doors not being yellow (as it seems was intended). I removed the (2, 2, 2) scale on the UV map, instead baking the UV map scale into the UV map itself. Here are some comparison screenshots, where the left side is the old demo and the right side is this PR.

Screenshot 2024-08-27 at 1 51 39 PM

Screenshot 2024-08-27 at 1 52 32 PM

Screenshot 2024-08-27 at 1 52 07 PM

Screenshot 2024-08-27 at 2 05 15 PM


Calinou (commented):
Tested locally on 4.3.stable, it mostly works as expected.

The only issue I found is that the collision outside the stage is too low, so the wheels sink inside the grass:

image

This doesn't occur in the city or road.

aaronfranke:
@Calinou I can't reproduce that. The collider is located at Y=-5 and has a Y size of 10, so it should go up to the origin height, and the ground texture is slightly below the origin height.

Unrelated: I pushed a change that fixes the triangles on the undersides of each roof, and recolors all chimneys to be a dull red-brown-brick color (previously the buildings had a mix of beige and brown chimneys).

PR #1113, opened 2024-09-01
Add full license info for voxel textures

Montandalar:
Add full license info for voxel textures

Hello,

The textures in the Voxel Demo were included from [Minetest Game](https://github.com/minetest/minetest_game/) ("MTG") for [Minetest](minetest.net/) without "full" attribution of the authors. The attribution information needs to be present since the files are under CC BY-SA 3.0.

Although a general attribution was given to the "Minetest Contributors", I thought it was worth noting the original authors of each file that was used, and noting that some are actually under CC0. This info was supplied in a [README](https://github.com/minetest/minetest_game/blob/master/mods/default/README.txt).

Although CC0 is compatible with CC BY-SA 3.0 and the whole work can be received under CC BY-SA 3.0, it would be good for people to know there are parts they can reuse with less restriction. I think the current wording and linking to "MTG" makes it clear enough if someone wants to attempt such reuse.

### DIFF
--- a/3d/voxel/README.md
+++ b/3d/voxel/README.md
@@ -66,7 +66,8 @@ https://creativecommons.org/publicdomain/zero/1.0/
 
 Font is "TinyUnicode" by DuffsDevice. Copyright &copy; DuffsDevice, CC-BY (Attribution) http://www.pentacom.jp/pentacom/bitfontmaker2/gallery/?id=468
 
-### Copyright info for textures reused from Minetest Game
+### Copyright information for textures reused from Minetest Game
+
 While most textures are under CC BY-SA 3.0, some are under CC0 1.0
 
 Cisoun's texture pack (CC BY-SA 3.0):
--- a/3d/voxel/README.md
+++ b/3d/voxel/README.md
@@ -54,7 +54,102 @@ use Zylann's voxel module instead: https://github.com/Zylann/godot_voxel
 
 ## Licenses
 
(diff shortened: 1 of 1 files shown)


aaronfranke:
I don't think it's legally necessary to list the CC0 textures, but maybe it's a good idea to do it anyway. I'll wait a bit in case other contributors want to give feedback on this PR.

On a technical level, I suggest using \` characters around file names instead of using `\_`.

3d/voxel/README.md
```diff
@@ -54,7 +54,102 @@ use Zylann's voxel module instead: https://github.com/Zylann/godot_voxel
...
+ CC BY-SA 3.0 Unported (Attribution-ShareAlike)
 http://creativecommons.org/licenses/by-sa/3.0/
 
+Some textures Copyright &copy; 2010-2018 Minetest contributors,
+ CC0 1.0 "No rights reserved"
+https://creativecommons.org/publicdomain/zero/1.0/
+
 Font is "TinyUnicode" by DuffsDevice. Copyright &copy; DuffsDevice, CC-BY (Attribution) http://www.pentacom.jp/pentacom/bitfontmaker2/gallery/?id=468
+
+### Copyright info for textures reused from Minetest Game
```
Calinou, line 69:
```suggestion
### Copyright information for textures reused from Minetest Game

```

Calinou (approved):
Thanks!

PR #1115, opened 2024-09-16
Fix VRAM leak in compositor effect demo

RGDTAB:
The compositor effect demo was attempting to free data created by a RenderingDevice using the RenderingServer. RenderingDevice RIDs are not the same as RenderingServer RIDs, and can't be freed by the RenderingServer.


Calinou (approved):

PR #1122, opened 2024-10-16
Add 3d multiplayer demo

devmoreir4:
This is a template for a 3D multiplayer game developed in Godot Engine 4.3. It provides a basic structure for a multiplayer setup, where each player has a nickname displayed above their character and the option to choose from four different skins: red, green, blue, or yellow.

Preview:
Preview


Calinou:
I don't think this should be referred to as a template, as we don't expect the demo projects to be used as a base for a full-blown projects.

A project template has very different requirements to a demo (and isn't maintained in the same way as a demo).

PR #1124, opened 2024-10-22
Improve Operating System Testing demo

Calinou:
- Display more return values of Engine/OS methods such as system fonts, memory information and security information.
- Add headless support (the generated report is also printed to stdout).
  - **Note:** Querying MIDI inputs is disabled when in headless mode to work around https://github.com/godotengine/godot/issues/52821.
- Improve theming in the generated report.
- Hide video adapter type if this information can't be queried in the current rendering method.
- Update C# script for Godot 4.x platform names.

Tested on 4.3.stable.

### Stitched report

Full report stitched from multiple screenshots, so you can see it in all it sglory.

Report

### Headless support

From VS Code's integrated terminal:

image

### DIFF
--- a/misc/os_test/os_test.gd
+++ b/misc/os_test/os_test.gd
@@ -3,6 +3,10 @@ extends Node
 @onready var rtl: RichTextLabel = $HBoxContainer/Features
 @onready var csharp_test: Node = $CSharpTest
 
+# Line number for alternate line coloring. Incremented by 1 each time a line is added
+# (ignoring headers).
+var line_count := 0
+
 # Returns a human-readable string from a date and time, date, or time dictionary.
 func datetime_to_string(date: Dictionary) -> void:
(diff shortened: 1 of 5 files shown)


aaronfranke (approved):

misc/os_test/os_test.gd
```diff
@@ -38,27 +42,48 @@ func datetime_to_string(date: Dictionary) -> void:
 		})
 
 
-func scan_midi_devices() -> String:
+func scan_midi_inputs() -> String:
+	if DisplayServer.get_name() == "headless":
+		# Workaround for <https://github.com/godotengine/godot/issues/52821>.
+		return ""
```
Calinou, line 48: Note that checking the audio driver name would be preferable, as you can use the `Dummy` audio driver but keep a non-headless display driver.

PR #1127, opened 2024-10-29
Update web demo exports for Godot 4.3

Calinou:
This comes with other improvements:

- Demos now work on desktop and mobile browsers, as ETC2 import is now automatically enabled before exporting each demo.
- Progressive web app is available for each project.
  - Can be added to the home screen on a device for a closer-to-native experience.
  - Cross-origin isolation headers are automatically added, so that the export works with threads enabled despite being hosted on GitHub Pages (which doesn't send these headers).

Several demos will need further tweaks to look better when using the Compatibility rendering method. Currently, many 3D demos appear too bright due to lights with shadows enabled using sRGB blending.

Also, many demos will need to be modified to indicate which parts of them are not supported when running on the web platform, or when using the Compatibility rendering method. I'll open a separate PR to handle this.

Try the demos with this link: https://calinou.github.io/godot-demo-projects/

- This closes https://github.com/godotengine/godot-demo-projects/issues/786.


fire (approved):
I randomly selected a few to test the web demos and they seemed to run on macos firefox.

Did not test support on the phones but need help.

The builds play so cross origin is working.

@Calinou Do you think we should yolo merge it or test each by each with a matrix?

shahriarlabib000:
I think `3d/material_testers` uses `Forward+` renderer

aaronfranke (approved):

Calinou:
> I think 3d/material_testers uses Forward+ renderer. Tested on an Android device. Was stuck in the loading process for a long time, lost my patience.

Compatibility is always used on the web platform, as web browsers don't support Vulkan. [Godot doesn't support WebGPU yet](https://github.com/godotengine/godot-proposals/issues/6646), which would be required to support Forward+/Mobile on the web platform.

Alex2782:
@Calinou Is it intentional that it will also be executed in forks?



image

Calinou:
> @Calinou Is it intentional that it will also be executed in forks?
> 
> `.github/workflows/export_html5.yml.disabled` → `.github/workflows/export_web.yml` The file has been renamed (activated)
> 
> image

No, it shouldn't be executed in forks.

PR #1129, opened 2024-11-02
Update some GDScript files

dalexeev:
* Replace `not X is Y` with `X is not Y`.
  * This is supported since 4.3, so `project.godot` and `*.import` files for such projects have also been updated.
* Replace `# warning-ignore:X` with `@warning_ignore("X")` and remove unnecessary warning ignores.
* Convert comment keywords such as `NOTE` and `BUG` to uppercase (they are highlighted in the editor).

### DIFF
--- a/2d/dynamic_tilemap_layers/level/tile_map.gd
+++ b/2d/dynamic_tilemap_layers/level/tile_map.gd
@@ -42,7 +42,7 @@ func _tile_data_runtime_update(_layer: int, _coords: Vector2i, tile_data: TileDa
 
 
 func _on_secret_detector_body_entered(body: Node2D) -> void:
-	if not body is CharacterBody2D:
+	if body is not CharacterBody2D:
 		# Detect the player only.
 		return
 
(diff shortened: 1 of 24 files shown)


aaronfranke (commented):
I'm not totally happy with making a cosmetic change that makes the demos incompatible with earlier Godot releases, because it's a barrier to testing the latest demos on older Godot releases. I would personally wait a year or two for `is not` to have been out for awhile. But we can go forward with it if you prefer that.

The other changes are just good 👍

dalexeev:
> I would personally wait a year or two for `is not` to have been out for awhile.

We could put this in a separate PR if you prefer. I just thought that the master branch here is the same as in the main repository, and there is a separate branch for 4.3. At least that's what the README says.

aaronfranke (approved):

Calinou (approved):
Thanks!

Regarding version compatibility, it's fine for `master` to support only 4.3 and later (new demos can be 4.4-dev only too). We have stable branches for older releases.

PR #1132, opened 2024-11-13
Add fallbacks and mention non-working features when using Compatibility

Calinou:
- Make 3D lights with shadows darker when using Compatibility to better match the apperance of Forward+/Mobile, due to the use of sRGB blending for those lights.
- Silence some warnings when running projects with the Compatibility rendering method, such as the one about 2D MSAA not being supported.
- Rebake lightmaps in Global Illumination to improve quality and reduce light leaking with improvements from 4.3.
  - Increase probe density for dynamic objects, and mark the moving box as a dynamic object for GI purposes.

This is done to improve the experience with the web-based demos, which always run with the Compatibility rendering method.

### Forward+/Mobile

Forward+/Mobile

### Compatibility

*Example message about an unsupported feature in the bottom-left corner.*

Compatibility


aaronfranke:
Needs rebase.

Calinou:
Rebased and tested again, it works as expected. I've made use of `RenderingServer.get_current_rendering_method()` so that it follows automatic fallbacks and CLI argument overrides.

PR #1139, opened 2024-11-25
Fix physics picking in Gui in 3D Demo

Sauermann:
Physics picking requires that the SubViewport has a consistent state of the mouse-enter/exit notifications.

Godot V4.3 was changed in comparison to V4.2, so that it now requires this additional step.

While this demo is focused on Control nodes, Physics picking should also work, since the demo project is used sometimes as a baseline for physics interactions ([for example on the forum](https://forum.godotengine.org/t/area2d-not-detecting-input-through-subviewport-push-input/92571)).

### DIFF
--- a/viewport/gui_in_3d/gui_3d.gd
+++ b/viewport/gui_in_3d/gui_3d.gd
@@ -30,9 +30,13 @@ func _process(_delta: float) -> void:
 
 func _mouse_entered_area() -> void:
 	is_mouse_inside = true
+	# Notify the viewport, that the mouse is now hovering it.
+	node_viewport.notification(NOTIFICATION_VP_MOUSE_ENTER)
 
 
 func _mouse_exited_area() -> void:
+	# Notify the viewport, that the mouse is no longer hovering it.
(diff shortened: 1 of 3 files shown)


viewport/gui_in_3d/gui_3d.gd
```diff
@@ -30,9 +30,13 @@ func _process(_delta: float) -> void:
 
 func _mouse_entered_area() -> void:
 	is_mouse_inside = true
+	# Notify the viewport, that the mouse is now hovering it.
+	node_viewport.notification(NOTIFICATION_VP_MOUSE_ENTER)
 
 
 func _mouse_exited_area() -> void:
+	# Notify the viewport, that the mouse is no longer hovering it.
```
aaronfranke, line 38:
```suggestion
	# Notify the viewport that the mouse is now hovering it.
	node_viewport.notification(NOTIFICATION_VP_MOUSE_ENTER)


func _mouse_exited_area() -> void:
	# Notify the viewport that the mouse is no longer hovering it.
```

aaronfranke (approved):

PR #1140, opened 2024-11-25
Improve GUI in 3D demo

Calinou:
- Follow-up to https://github.com/godotengine/godot-demo-projects/pull/1139.

___

- Increase physics ticks per second to 120, so that UI interactions feel smoother and have lower input lag. This happens since input is governed by the physics tick due to the use of physics picking.
- Add a placeholder to the LineEdit and select a default option in the OptionButton.

## Preview

https://github.com/user-attachments/assets/1a68a9ed-c1a7-45b9-a2d4-4be7ff376f20

### DIFF
--- a/viewport/gui_in_3d/gui_panel_3d.tscn
+++ b/viewport/gui_in_3d/gui_panel_3d.tscn
@@ -59,8 +59,9 @@ horizontal_alignment = 1
 layout_mode = 2
 text = "A button!"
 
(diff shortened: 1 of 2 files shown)


Sauermann (approved):
Have tested the changes with v4.3.stable.official [77dcf97d8] and they work for me without problems. Code looks good.

aaronfranke (approved):

PR #1143, opened 2024-12-03
Add player_state dictionary to manage player state names for FSM demo

Yaxian:
Add `player_state` to manage player state.

```
static var player_state := {
	"previous": "previous",
	"jump": "jump",
	"idle": "idle",
	"move": "move",
	"stagger": "stagger",
	"attack": "attack",
	"die": "die",
	"dead": "dead",
	"walk": "walk",
}

```
I tried to use `enum`, like 

`state.gd`
```
enum BaseState {
 	PREVIOUS
}
```

`player_state.gd`

```
extends "res://state_machine/state.gd"

enum PlayerState {
	JUMP
}
```

But the function signature verification fails. Is there a good way to manage state in `state machine`, like type check, code hints?


2d/finite_state_machine/player/player_state.gd
```diff
@@ -0,0 +1,13 @@
...
+	"previous": "previous",
+	"jump": "jump",
+	"idle": "idle",
+	"move": "move",
+	"stagger": "stagger",
+	"attack": "attack",
+	"die": "die",
+	"dead": "dead",
+	"walk": "walk",
+}
```
aaronfranke: I think it would be better to use a constant, and use StringNames, and a typed Dictionary. Places like `signal finished` would need to be updated to StringName as well.

```suggestion
const PLAYER_STATE: Dictionary[StringName, StringName]= {
	&"previous": &"previous",
	&"jump": &"jump",
	&"idle": &"idle",
	&"move": &"move",
	&"stagger": &"stagger",
	&"attack": &"attack",
	&"die": &"die",
	&"dead": &"dead",
	&"walk": &"walk",
}
```

aaronfranke (approved):
I applied my suggested changes, thanks for the PR!

PR #1146, opened 2024-12-12
physics_tests: Upped the Jolt Physics limit settings so that the performance tests run without errors

jrouwe:
This change makes sure that if you switch physics to use the Jolt Physics module, that the correct settings are applied.


aaronfranke (approved):
I suppose this demo project needs higher limits than most games, so this makes sense.

Calinou (approved):

PR #1149, opened 2024-12-19
Explain how to setup gui_in_3d

idbrii:
Make it more obvious to newcomers how this demo works so they can replicate it in their projects.

Change the text to point to how this is working.

Add Editor Descriptions to make it clearer how this is setup.

image

### DIFF
--- a/viewport/gui_in_3d/gui_panel_3d.tscn
+++ b/viewport/gui_in_3d/gui_panel_3d.tscn
@@ -23,6 +23,7 @@ process_mode = 3
 script = ExtResource("1")
 
 [node name="SubViewport" type="SubViewport" parent="."]
+editor_description = "SubViewport allows us to use a ViewportTexture to display the contents of this viewport."
 gui_embed_subwindows = true
 size = Vector2i(560, 360)
 render_target_update_mode = 4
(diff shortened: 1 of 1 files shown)


idbrii:
Not sure if it's worth mentioning this relevant docs page:
https://docs.godotengine.org/en/stable/tutorials/shaders/using_viewport_as_texture.html
Could add to the readme, but it's only kind of related so maybe that's more confusing?

viewport/gui_in_3d/gui_panel_3d.tscn
```diff
@@ -23,6 +23,7 @@ process_mode = 3
 script = ExtResource("1")
 
 [node name="SubViewport" type="SubViewport" parent="."]
+editor_description = "SubViewport allows us to use a ViewportTexture to display the contents of this viewport."
```
aaronfranke, line 26:
```suggestion
editor_description = "SubViewport contents can be displayed using a ViewportTexture."
```

aaronfranke (commented):
Makes sense.

aaronfranke (approved):

Calinou (approved):
Thanks!

PR #1155, opened 2025-01-10
Update to TileMapLayer on multiplayer bomber.

waynemwashuma:
Updated the `TileMap` to a `TileMapLayer` to remove the deprecation warning when using a  `TileMap` .

Change is not compatible with version 3.x.x due to the new  `TileMapLayer`.


Calinou (approved):

Thanks!

PR #1162, opened 2025-02-06
Add a graphics tablet input demo

Calinou:
This adds a demo for pen input in Godot, which will also come in handy to test tablet-related PRs like https://github.com/godotengine/godot/pull/76120.

This uses code from https://github.com/godotengine/godot/issues/75903 for drawing (thanks @nlupugla!).

This demo was developed on 4.3, so it can be cherry-picked.

## Preview

graphics_tablet_input webp

## TODO

- [x] Fix width curve according to pen pressure. Right now, it changes the entire line's width, but it should keep the history of the pressure over the line's length and apply it correctly. I have code in place for that but it not's working correctly for some reason. Please help on this if you can :slightly_smiling_face:
  - https://github.com/godotengine/godot-proposals/issues/9241 could make this easier to implement.


bruvzg:
Tested on macOS with Wacom tablet and Apple Pencil, some notes:
- tilt registers but do nothing, it's probably should have at least some indication, at least display a tilt vector, or maybe change color.
- extremely long lines start to lag, so it's probable worth splitting the line at some point.

Calinou:
>     * tilt registers but do nothing, it's probably should have at least some indication, at least display a tilt vector, or maybe change color.

Done, but I can't test as my tablet (Wacom Intuos M 2013) doesn't support tilt.

>     * extremely long lines start to lag, so it's probably worth splitting the line at some point.

Done (it now splits every 1024 points).

bruvzg:
> I added a tilt vector visualization, but I can't test it as my tablet (Wacom Intuos M 2013) doesn't support tilt.

Seems to be working fine.

https://github.com/user-attachments/assets/d2cd9c20-7f21-4bb0-875f-a1bbc7429921

bruvzg (approved):

misc/graphics_tablet_input/graphics_tablet_input.gd
```diff
@@ -0,0 +1,189 @@
...
+const SPLIT_POINT_COUNT = 1024
+
+var stroke: Line2D
+var width_curve: Curve
+var pressures := PackedFloat32Array()
+var event_position: Vector2
+var event_tilt: Vector2
+
+var line_color := Color.BLACK
+var line_width := 3.0
```
aaronfranke, line 15: Per the [GDScript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#inferred-types), prefer writing the type explicitly when the type is not written on the same line.

```suggestion
var line_width: float = 3.0
```

misc/graphics_tablet_input/graphics_tablet_input.gd
```diff
@@ -0,0 +1,189 @@
...
+
+# If `true`, modulate line width accordding to pen pressure.
+# This is done using a width curve that is continuously recreated to match the line's actual profile
+# as the line is being drawn by the user.
+var pressure_sensitive := true
+
+var show_tilt_vector := true
```
aaronfranke, line 22:
```suggestion
var pressure_sensitive: bool = true

var show_tilt_vector: bool = true
```

aaronfranke (commented):
I don't have a graphics tablet so I haven't tested it, but from reading the code on GitHub, it looks great!

PR #1164, opened 2025-02-09
Change the way the mobs are cleared from screen

fsantos98:
This PR was created because of [#10655](https://github.com/godotengine/godot-docs/pull/10655#issuecomment-2646591574)

I think it makes more sense to free the mobs when it's game over so it's more visual to the person who is learning. 

At this point in the tutorial when we die we have a ~2 second window to be able to start the game again. In this time all the mobs on the screen are not visible anymore, so the `queue_free` action is not too obvious. 
It feels better to be able to see what this line of code is doing by removing them immediately when it's game over, this way we can visually see what the code is doing.

Before:
2143d44000dd690a8406b1ac9b6fa0ce

After:
69b004679a44531ae3bd28995287506b

### DIFF
--- a/2d/dodge_the_creeps/main.gd
+++ b/2d/dodge_the_creeps/main.gd
@@ -4,6 +4,7 @@ extends Node
 var score
 
 func game_over():
+	get_tree().call_group(&"mobs", &"queue_free")
 	$ScoreTimer.stop()
 	$MobTimer.stop()
 	$HUD.show_game_over()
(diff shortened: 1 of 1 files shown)


aaronfranke:
The existing behavior is better. It lets you see the enemies after death. Deleting everything on game over is jarring, and leaves the player wondering what happened.

skyace65:
I'm in agreement with aaronfranke. I prefer the current behavior

tetrapod00:
I'm also against the change. Without actual death animations, neither one looks very polished. But the changed behavior looks more jarring than the existing behavior

aaronfranke:
Thank you for your contribution! Unfortunately, we have decided to reject this change, and the corresponding docs PR.

PR #1165, opened 2025-02-10
Fix `look_at_from_position()` usage in Squash the Creeps (3D)

Calinou:
The mob's orientation was previously shifted according to the player's height, which could lead to collision and movement issues that were difficult to diagnose.


aaronfranke (approved):

tetrapod00 (approved):

gabriele2000:
In my modest opinion, I'd like to propose a change.

The following changed code
```gdscript
func initialize(start_position, player_position):
	# Ignore the player's height, so that the mob's orientation is not slightly
	# shifted if the mob spawns while the player is jumping.
	var target = Vector3(player_position.x, start_position.y, player_position.z)
	look_at_from_position(start_position, target, Vector3.UP)
```
could be just
```gdscript
func initialize(start_position, player_position):
	# Ignore the player's height, so that the mob's orientation is not slightly
	# shifted if the mob spawns while the player is jumping.
        # start_position.y could also be written as 0.0
	player_position = Vector3(player_position.x, start_position.y, player_position.z)
	look_at_from_position(start_position, player_position, Vector3.UP)
```

`player_position` has been already declared inside `func`, now we just need to overwrite the `player_position.y` value then process the overwrited value.
No need to declare a variable, let alone a new one, just for a small correction.

aaronfranke:
Needs a rebase.

@gabriele2000 Modifying function parameters is considered a bad practice by some, so I wouldn't encourage it in official demos, even though it would work fine either way and potentially use ever-so-slightly less memory.

aaronfranke (approved):
I resolved the conflicts, this is good to go!

PR #1168, opened 2025-02-17
Add WebXR demo

### LINKED ISSUE #1167: WebXR demo
There is no WebXR demo. The XR demos do not support WebXR.

It would be nice to have a WebXR demo which is exported to the web and includes the WebXR Polyfill and WebXR Layers Polyfill, e.g. as described in https://www.snopekgames.com/tutorial/2023/how-make-vr-game-webxr-godot-4

(In my case I wanted to use a demo project to report a WebXR rendering issue, but found there was none.)

rburing:
Minimal demo based on https://docs.godotengine.org/en/stable/classes/class_webxrinterface.html and https://www.snopekgames.com/tutorial/2023/how-make-vr-game-webxr-godot-4

- Closes https://github.com/godotengine/godot-demo-projects/issues/1167

cc @dsnopek


xr/webxr/main.gd
```diff
@@ -0,0 +1,134 @@
...
+		# This returns immediately - our _webxr_session_supported() method
+		# (which we connected to the "session_supported" signal above) will
+		# be called sometime later to let us know if it's supported or not.
+		webxr_interface.is_session_supported("immersive-vr")
+
+	$XROrigin3D/LeftController.button_pressed.connect(self._on_left_controller_button_pressed)
+	$XROrigin3D/LeftController.button_released.connect(self._on_left_controller_button_released)
+
+
+
```
dsnopek, line 36: One too many empty lines?

```suggestion


```

xr/webxr/main.gd
```diff
@@ -0,0 +1,134 @@
...
+	if not vr_supported:
+		OS.alert("Your browser doesn't support VR")
+		return
+
+	# We want an immersive VR session, as opposed to AR ('immersive-ar') or a
+	# simple 3DoF viewer ('viewer').
+	webxr_interface.session_mode = 'immersive-vr'
+	# 'bounded-floor' is room scale, 'local-floor' is a standing or sitting
+	# experience (it puts you 1.6m above the ground if you have 3DoF headset),
+	# whereas as 'local' puts you down at the ARVROrigin.
```
dsnopek, line 52: Left over Godot 3.x reference:

```suggestion
	# whereas as 'local' puts you down at the XROrigin3D.
```

xr/webxr/main.gd
```diff
@@ -0,0 +1,134 @@
...
+	# This list means it'll first try to request 'bounded-floor', then
+	# fallback on 'local-floor' and ultimately 'local', if nothing else is
+	# supported.
+	webxr_interface.requested_reference_space_types = 'bounded-floor, local-floor, local'
```
dsnopek, line 56: Given the number of people that have copy-pasted this code wholesale over the years, I've been regretting including `'bounded-floor'` in here. It's a good example to show both optional and required features, but I suspect most developers would actually have a better experience with `'local-floor'`.

But I'm not entirely sure we should change this, and if we do, it should be changed in the `WebXRInterface` docs too.

dsnopek (commented):
Thanks!

The demo itself looks great! I tested in the Meta WebXR emulator in Chrome, and in the Meta Quest browser on a Quest 3, and everything worked great. I only have a handful of nitpicky comments on the code.

The main thing I'm worried about, though, is that this doesn't include the polyfills, because it doesn't have the `export_presets.cfg`, which means this won't work on all browsers (most significantly on Firefox). This isn't ideal, especially because this will get exported and uploaded for folks to try, and may lead to confusion around whether certain browsers are supported or not.

I'm not sure how to handle this, though, because we already have the standard of not including `export_presets.cfg` with any of the demo projects

cc @Calinou

rburing:
In the latest push I added a special case for the WebXR demo export preset, to add the polyfills using `sed`.

xr/webxr/main.gd
rburing: Personally I would just keep the configuration as is for this demo (because as you say it's a good example of optional and required features), but we could modify the comments or add to them, here and/or in the WebXRInterface documentation.

dsnopek: Yeah, that makes sense! The descriptions of the reference spaces in the comment also aren't that great.

I'll take a stab at rewriting the comment, but we don't really need to fix this here and now - we can iterate on it elsewhere

xr/webxr/main.gd
```diff
@@ -0,0 +1,133 @@
...
+
+	# We want an immersive VR session, as opposed to AR ('immersive-ar') or a
+	# simple 3DoF viewer ('viewer').
+	webxr_interface.session_mode = 'immersive-vr'
+	# 'bounded-floor' is room scale, 'local-floor' is a standing or sitting
+	# experience (it puts you 1.6m above the ground if you have 3DoF headset),
+	# whereas as 'local' puts you down at the XROrigin3D.
+	# This list means it'll first try to request 'bounded-floor', then
+	# fallback on 'local-floor' and ultimately 'local', if nothing else is
+	# supported.
```
dsnopek, line 54: A first pass at rewriting this, but I'm not super happy with it:

```suggestion
	# The reference space affects what the XROrigin3D position represents, and
	# consequently what the positions of the XRCamera3D and XRController3D
	# nodes are relative to:
	# - 'bounded-floor' is for room-scale, where the XROrigin3D will be positioned
	#   in the center of your play space
	# - 'local-floor' is for most standing or sitting experiences, where you want the
	#   XROrigin3D to positioned at your real world floor, but the X and Z position
	#   will start where you were standing when entering XR
        # - 'local' puts the XROrigin3D where your head was when entering XR, and is
        #   best suited for experiences that don't have any relation to the real world floor
        #   like driving or flight simulators, or where the character you embody is of a different
        #   height than the real player
	# This list means it'll first try to request 'bounded-floor', then fallback on 'local-floor' 
	# and ultimately 'local', if nothing else is supported.
```

... or, maybe the whole list of reference spaces should be removed, and we should just have the last part "This list means it'll first try..." and have a link to some other documentation where we describe them individually?

dsnopek (approved):
Thanks!

> In the latest push I added a special case for the WebXR demo export preset, to add the polyfills using `sed`.

This solves one problem: the public upload of the demos will work on all the browsers it can.

However, if someone starts their WebXR project by copy-pasting the demo, they won't have the polyfills, which isn't ideal.

I don't necessarily think this should hold up getting a demo in there at all, though! We could continue to figure out how to improve this in a follow-up

xr/webxr/README.md
```diff
@@ -0,0 +1,7 @@
+# WebXR demo
+
+This is a minimalist demo of WebXR rendering and controller support.
```
AThousandShips:
```suggestion
This is a minimal demo of WebXR rendering and controller support.
```
Minimalist implies more than minimal

xr/webxr/main.gd
```diff
@@ -0,0 +1,133 @@
+extends Node3D
+
+var webxr_interface: XRInterface
+var vr_supported: bool = false
+
+
+func _ready() -> void:
+	$CanvasLayer/EnterVRButton.pressed.connect(self._on_enter_vr_button_pressed)
```
AThousandShips:
```suggestion
	$CanvasLayer/EnterVRButton.pressed.connect(_on_enter_vr_button_pressed)
```
Redundant, and also unsafe

xr/webxr/main.gd
```diff
@@ -0,0 +1,133 @@
...
+		webxr_interface.session_ended.connect(self._webxr_session_ended)
+		webxr_interface.session_failed.connect(self._webxr_session_failed)
+
+		webxr_interface.select.connect(self._webxr_on_select)
+		webxr_interface.selectstart.connect(self._webxr_on_select_start)
+		webxr_interface.selectend.connect(self._webxr_on_select_end)
+
+		webxr_interface.squeeze.connect(self._webxr_on_squeeze)
+		webxr_interface.squeezestart.connect(self._webxr_on_squeeze_start)
+		webxr_interface.squeezeend.connect(self._webxr_on_squeeze_end)
```
AThousandShips:
```suggestion
		webxr_interface.session_supported.connect(_webxr_session_supported)
		webxr_interface.session_started.connect(_webxr_session_started)
		webxr_interface.session_ended.connect(_webxr_session_ended)
		webxr_interface.session_failed.connect(_webxr_session_failed)

		webxr_interface.select.connect(_webxr_on_select)
		webxr_interface.selectstart.connect(_webxr_on_select_start)
		webxr_interface.selectend.connect(_webxr_on_select_end)

		webxr_interface.squeeze.connect(_webxr_on_squeeze)
		webxr_interface.squeezestart.connect(_webxr_on_squeeze_start)
		webxr_interface.squeezeend.connect(_webxr_on_squeeze_end)
```

xr/webxr/main.gd
```diff
@@ -0,0 +1,133 @@
...
+
+		# This returns immediately - our _webxr_session_supported() method
+		# (which we connected to the "session_supported" signal above) will
+		# be called sometime later to let us know if it's supported or not.
+		webxr_interface.is_session_supported("immersive-vr")
+
+	$XROrigin3D/LeftController.button_pressed.connect(self._on_left_controller_button_pressed)
+	$XROrigin3D/LeftController.button_released.connect(self._on_left_controller_button_released)
```
AThousandShips:
```suggestion
	$XROrigin3D/LeftController.button_pressed.connect(_on_left_controller_button_pressed)
	$XROrigin3D/LeftController.button_released.connect(_on_left_controller_button_released)
```

aaronfranke:
@rburing Can you address the review comments from @AThousandShips and @dsnopek?

rburing:
I addressed the style comments and updated to 4.5.

I agree the comments in `main.gd` can be improved, and that the demo should have an export preset with the polyfills already included, but these are larger changes that can be done as a follow-up (especially since we're not exactly sure yet how to make those changes). For now, I added instructions to include the polyfills to the demo's `README.md`.

aaronfranke (approved):
I pushed a commit for some style and type safety fixes. Thanks!

PR #1171, opened 2025-03-01
Fix webrtc example that relied on JSON integers

goatchurchprime:
This fixes the problem identified by https://github.com/godotengine/godot/issues/103374

The `JSON.parse_string()` creates floats instead of integers from numbers, which the example flagged as an error.  The solution is to send integers as strings instead.


aaronfranke:
Isn't this a workaround, not a solution? JSON numbers are, in fact, floats. Any check like `str(msg.type).is_valid_int()` in the `ws_webrtc.client._parse_msg()` function is just not a good check, right?

zodiepupper:
i know from my web/api experience that there isn't a consensus on which method is actually correct. because javascript treats floats and ints as both compatible with the number type, it doesn't have to account for this. 

it would be better to decode to float and then the impl will have to decide if it needs to cast that to an int. converting an int to a string would break compat with most api.

PR #1174, opened 2025-03-05
Fix compute shader so uniforms are the same

stuartcarnie:
Previously,

- set 0 and 1 were read-only, and
- set 1 was write-only

Metal's SPIR-V reflection implementation correctly distinguished these attributes, and resulted in the failure described by issue https://github.com/godotengine/godot/issues/101696

A fix for Godot's default SPIR-V reflection is found in https://github.com/godotengine/godot/pull/103614

To ensure the demo keeps working, we ensure all sets have matching formats.

### DIFF
--- a/compute/texture/water_plane/water_compute.glsl
+++ b/compute/texture/water_plane/water_compute.glsl
@@ -5,48 +5,48 @@
 layout(local_size_x = 8, local_size_y = 8, local_size_z = 1) in;
 
 // Our textures.
-layout(r32f, set = 0, binding = 0) uniform restrict readonly image2D current_image;
-layout(r32f, set = 1, binding = 0) uniform restrict readonly image2D previous_image;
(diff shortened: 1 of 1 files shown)


clayjohn (approved):
Approving of the code change. I'm not sure what the style guide says for demos, so not approving/rejecting the whitespace changes

Calinou (changes requested):
Looks good to me, but please keep the tab indentation within the shader file (it was converted to spaces).

The [Godot shader style guide](https://docs.godotengine.org/en/latest/tutorials/shaders/shaders_style_guide.html) recommends tabs for indentation, and while GLSL is not the same as Godot shaders, it's fair to assume the same recommendations apply. (We should make this explicit in the docs still.)

Calinou:
- Superseded by https://github.com/godotengine/godot-demo-projects/pull/1247. Thanks for the contribution nonetheless!

PR #1175, opened 2025-03-07
Added line to play animation at the end.

utsav132:
The '$AnimatedSprite2D.play()' is called at the beginning in the '_ready' function, when it should actually be called at the end, after selecting the animation.


Calinou (approved):
Thanks!

PR #1189, opened 2025-04-14
Improve flat terrain generation in the Voxel demo

Calinou:
This adds 2 layers of dirt below the grass layer and a layer of indestructible bedrock below it so you can't fall off the world.

## Preview

image


aaronfranke (approved):
The reason the terrain was so thin before was due to performance reasons, it was quite slow in Godot 3.x at least to have thick terrain. However if this performs well then let's do it.

PR #1193, opened 2025-04-16
Improve static checks by using pre-commit

Calinou:
The setup is taken from the main Godot repository.


aaronfranke (approved):

Calinou:
I think we should hold on merging this until https://github.com/godotengine/godot/issues/105456 is fixed, as this will cause issues otherwise. Therefore, I'll mark this as a draft.

**Edit:** https://github.com/godotengine/godot/pull/105779 aims to fix this issue.

PR #1195, opened 2025-04-18
Use Jolt Physics in all 3D demos that use physics, improve physics tests demos

Calinou:
- Add options for physics ticks per second, time scale, max physics steps per frame and physics interpolation to the 2D and 3D physics tests demos.
  - Physics ticks per second are always multiplied by time scale so that time scale does not affect the physics simulation quality.
- Enable 4× MSAA for better debug shape display. Remove meshes/lights as the debug collision fill make these unnecessary.
  - Switch to the Mobile rendering method in the 2D physics tests demo to allow for 2D MSAA, as it's not implemented in Compatibility yet.
- Improve collision shapes color in the 2D and 3D physics tests demos for better visibility. Each PhysicsBody type now has its own collision shape color.

I've tested all demos and can confirm they work as expected. I also fixed a few issues in the 3D physics tests demo (e.g. the moving platform not moving anymore after the project was upgraded, due to a broken AnimationPlayer setup).

## Preview

2D

3D


aaronfranke (approved):

3d/kinematic_character/project.godot
```diff
@@ -84,6 +84,7 @@ exit={
 
 common/physics_ticks_per_second=120
 3d/default_gravity=13.0
+3d/physics_engine="Jolt Physics"
```
Rindbee, line 87: I'm curious if this was added manually in an external editor, or automatically by changing the project settings.

After opening the project via the editor, the order changes.

```diff
diff --git a/3d/kinematic_character/project.godot b/3d/kinematic_character/project.godot
index b834bb6d..9119392c 100644
--- a/3d/kinematic_character/project.godot
+++ b/3d/kinematic_character/project.godot
@@ -83,8 +83,8 @@ exit={
 [physics]
 
 common/physics_ticks_per_second=120
-3d/default_gravity=13.0
 3d/physics_engine="Jolt Physics"
+3d/default_gravity=13.0
 common/physics_interpolation=true
 
 [rendering]
```

Calinou: This was added with search-and-replace in an external editor, which is why ordering can change on save.

When 4.5.stable is released, I'll go through all demos and resave them to avoid further diffs.

PR #1197, opened 2025-04-24
Add a simple rhythm game demo

PizzaLovers007:
Follow-up for one of the project suggestions in https://github.com/godotengine/godot/pull/105510#issuecomment-2816774232.

This demo project is a rhythm game that utilizes the audio syncing strategies from https://docs.godotengine.org/en/stable/tutorials/audio/sync_with_audio.html along with a [1€ filter](https://gery.casiez.net/1euro/) to have a smooth and accurate playback position. The project was built in 4.4.1.stable, thus not requiring the DSP time/`play_scheduled` from godotengine/godot#105510.

The song was taken from the [BPM Sync](https://github.com/godotengine/godot-demo-projects/tree/master/audio/bpm_sync) demo project, and all other assets were made by me.

Key design decisions:
* The 1€ filter operates on the delta between the audio clock and system clock, rather than the time itself. This is because the anti-jittering mechanism only works well when the filtered value changes very little.
* V-sync is disabled to allow for more precise input timings (capped at 1000 FPS). "Mailbox" would be ideal here, but this behaves like `VSYNC_ENABLED` with **Compatibility** render mode. Switching to **Forward+** for this specific reason didn't seem worthwhile since tearing is not visible due to the notes moving vertically.


Calinou:
> V-sync is disabled to allow for more precise input timings (capped at 1000 FPS). "Mailbox" would be ideal here, but this behaves like VSYNC_ENABLED with Compatibility render mode. Switching to Forward+ for this specific reason didn't seem worthwhile since tearing is not visible due to the notes moving vertically.

Note that this should no longer be necessary if https://github.com/godotengine/godot-proposals/issues/552 is implemented.

audio/rhythm_game/globals/one_euro_filter.gd
```diff
@@ -0,0 +1,43 @@
+## Implementation of the 1€ filter (https://gery.casiez.net/1euro/).
+##
+## Modification of https://github.com/patrykkalinowski/godot-xr-kit/blob/master/addons/xr-kit/smooth-input-filter/scripts/one_euro_filter.gd
```
fire, line 3:
```suggestion
## Copyright (c) 2023 Patryk Kalinowski (patrykkalinowski.com)
## SPDX-License-Identifier: MIT
##
## Implementation of the 1€ filter (https://gery.casiez.net/1euro/).
## 
## Modification of https://github.com/patrykkalinowski/godot-xr-kit/blob/master/addons/xr-kit/smooth-input-filter/scripts/one_euro_filter.gd
```

How does this look?

PizzaLovers007: Good point, I added the license (moved it out of the documentation comment into a regular comment).

fire:
From the other pull request, I really appreciated the web demo for explaining.

PizzaLovers007:
The metronome demo PR is done now 😃 I was working on cleaning things up before sending it out. #1199

fire:
I played with the demo and I didn't notice any problems. The game was pretty fun!

fire (commented):
I noticed that when toggling the checkbox the filter in the middle of the game it doesn't reset and required a reset. Not sure.

PizzaLovers007:
> I noticed that when toggling the checkbox the filter in the middle of the game it doesn't reset and required a reset. Not sure.

Do you mean the checkbox setting persists after pressing R to restart? This is intentional since I thought it would make sense to see the opposite filter setting from the beginning if you toggled first, then hit R.

It's supposed to immediately swap between the filtered and unfiltered playback position though, so if you need to hit R to restart before the setting takes effect then something is wrong.

mk56-spn:
i have some minor gripes: 

the default latency is way overdone, 

this is after all a test that will mostly be used by devs on computers, where 42 ms of audio lag is very uncommon on the sort of systems a rhythm game would be being tested on in my estimation ( other than phones) 

granted i am not recommending a value as low as mine because i am using a very low latency setup, but something like + 10 ms seems more in line  with what most people would expect
image

overall a very well synced and pleasing demo that demonstrates the massive difference a filter does ( its noticeable enough on a 60 refresh rate and helps immensely there already, but cranking it up to 144 or 360 which is not uncommon in the rhythm game play landscape shows a world of difference.

if i had any other requests it would be a scroll speed slider, given jitter is most visible at higher scroll speeds, this would make the difference more stark on 60 especially

fire (approved):

PizzaLovers007:
Thanks for testing! I'm surprised -2ms is what you felt was correct. I do agree that 42ms is odd, but after repeated tests I can get it consistently <1ms error. I lowered it to 20ms to try to strike a middle ground.

I also added some more things:
* Scroll speed slider
* Sync test chart (simple quarter notes, maybe this would help with input sync testing)
* Metronome, though "flawed" because of the lack of sound scheduling

fire:
@Calinou Can you take a look? If you approve we can merge.

PizzaLovers007:
I've added a graph inspired by ITG Simply Love's results screen:
rhythm_game

Does this look good/is intuitive?

mk56-spn:
> I've added a graph inspired by ITG Simply Love's results screen: rhythm_game
> 
> Does this look good/is intuitive? Points below the line are late hits, and above the line are early hits.

clean and simple , similar to etternas too ( i believe both are Stepmania forks)

fire:
I uploaded this project as a itchio web https://ifiregames.itch.io/godot-engine-samples-rhythm-game

Calinou (approved):

I think this is a pretty good start as it is. I can perform a polishing pass after it's merged.

My score on a first playthrough (never listened to the full song before):

Screenshot_20250522_225749

I kept the latency settings to their default. I'm on a 240 Hz display with X11 compositing disabled and it seemed suitable out of the box. It was perhaps a tad high and could be changed according to the monitor refresh rate on startup automatically, but this can be difficult to get right.

fire:
Thanks!

PR #1191, opened 2025-04-15
Add Sprite3D and AnimatedSprite3D demo

yuantianle:
Finish demoing the project and implementing Sprite3D and AnimatedSprite3D. Also, provide game visual effects using shaders.

TODO: The link to the demo on the asset library in the README needs to be added. 

Related to #1173

PR #1198, opened 2025-04-24
Fix Sprite3D shader effect bug

yuantianle:
The normal was not calculated correctly

This is for the bug fix of the merged PR for #1191 

Before: 
image
After:
image


aaronfranke (approved):
Looks good to me.

PR #1205, opened 2025-05-15
Disable GH `export_web` workflow in forks

Alex2782:
https://github.com/godotengine/godot-demo-projects/pull/1127#issuecomment-2885245893

In my fork it is skipped, or should we leave `static checks` active in forks?

Bildschirmfoto 2025-05-16 um 01 42 12

### DIFF
--- a/.github/workflows/export_web.yml
+++ b/.github/workflows/export_web.yml
@@ -9,6 +9,7 @@ env:
 
 jobs:
   export-html5:
+    if: github.repository == 'godotengine/godot-demo-projects'
     name: Export projects to Web and deploy to GitHub Pages
     runs-on: ubuntu-24.04
     container:


aaronfranke:
We should keep static checks in forks, so that people can see if their PR has formatting issues before they submit a PR.

LikeLakers2:
I would recommend against locking the workflows to the main repository.

For one, by default, workflows do not run on forks. You have to go to the actions section and tell it to allow the workflows.

Additionally, if someone did enable workflows on their fork, but doesn't want certain workflows to run on their fork, they can [disable specific workflows.](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow)

.github/workflows/static_checks.yml
```diff
@@ -3,6 +3,7 @@ on: [push, pull_request]
 
 jobs:
   format:
+    if: github.repository == 'godotengine/godot-demo-projects'
```
LikeLakers2, line 6: I would recommend against locking the formatting job to the main repository.

If you're concerned about it running on your fork when you don't want it to, [it's possible to disable specific workflows from the Actions tab](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow).

Alex2782:
On my fork, static checks were executed in the new version.

image

aaronfranke (approved):

Calinou (approved):
Thanks!

PR #1208, opened 2025-05-29
navigation_astar: Migrate to TileMapLayer and various cleanups

rudis:
Hi,

besides the TileMapLayer migration I performed a few cleanups and added some comments which might be helpful.

Best,
Simon


Calinou (approved):
Tested locally on 4.4.1.stable, it works as expected.

Thanks!

PR #1211, opened 2025-06-09
Move Compute Shader Heightmap demo to the `compute/` folder

Calinou:
- Use GitHub admonition syntax in its README.


akien-mga (approved):

akien-mga:
Thanks!

Issue #1061, opened 2024-05-30
Update plugins demo to use custom Resources


Calinou:
**Which demo project is affected:** https://github.com/godotengine/godot-demo-projects/tree/master/plugins

**Description:** The Plugins demo currently doesn't rely on custom Resources and uses Nodes instead. This was because of limitations in Godot 3.x that made it impossible to export custom Resources. However, this limitation has been lifted in Godot 4.0, so the demo should be updated to use the new recommended usage patterns.

Calinou:
- This was implemented by https://github.com/godotengine/godot-demo-projects/pull/1219, closing.

PR #1219, opened 2025-06-24
Refactor Material Creator Plugin Demo to use custom resource-based data

SarunasRamonas:
Summary
- Refactored the **Material Creator plugin** to use custom Resource classes instead of Node for material data.
Addressing Issue #1061 .
- Replaced manual JSON serialization with Godot's built-in ResourceSaver and ResourceLoader.


plugins/addons/material_creator/material_creator.gd
```diff
@@ -41,56 +41,33 @@ func apply_pressed() -> void:
...
 
 func save_file_selected(path: String) -> bool:
-	var silly_resource: Variant = _silly_resource_from_values()
-	# Make a file, store the silly material as a JSON string.
-	var file := FileAccess.open(path, FileAccess.WRITE)
-	file.store_string(silly_resource.make_json())
-
-	return true
+	var silly_resource: Resource = _silly_resource_from_values()
+	# Save the resource as a .tres file using Godot's ResourceSaver
```
AThousandShips:
```suggestion
	# Save the resource as a .tres file using Godot's ResourceSaver.
```
Per the comment style

plugins/addons/material_creator/material_creator.gd
```diff
@@ -41,56 +41,33 @@ func apply_pressed() -> void:
...
-		$VBoxContainer/AlbedoColorPicker.color = silly_resource.albedo_color
-		$VBoxContainer/MetallicSlider.value = silly_resource.metallic_strength
-		$VBoxContainer/RoughnessSlider.value = silly_resource.roughness_strength
-
-		# Return `true` to indicate success.
-		return true
-
-	# If the file does not exist, then return `false` to indicate failure.
-	return false
+	# Load the resource using Godot's ResourceLoader
```
AThousandShips:
```suggestion
	# Load the resource using Godot's ResourceLoader.
```

plugins/addons/material_creator/material_creator.gd
```diff
@@ -41,56 +41,33 @@ func apply_pressed() -> void:
...
+func _silly_resource_from_values() -> Resource:
 	# Get the values from the sliders and color picker.
 	var color: Color = $VBoxContainer/AlbedoColorPicker.color
 	var metallic: float = $VBoxContainer/MetallicSlider.value
 	var roughness: float = $VBoxContainer/RoughnessSlider.value
-	# Make a new silly resource (which in this case actually is a node) and initialize it.
-	var silly_resource: Variant = silly_material_resource.new()
-	#silly_resource.init()
-
+	# Make a new silly resource (now a Resource, not a Node)
```
AThousandShips:
```suggestion
	# Make a new silly resource (now a Resource, not a Node).
```

plugins/addons/material_creator/material_creator.gd
```diff
@@ -41,56 +41,33 @@ func apply_pressed() -> void:
...
+	var silly_resource: Resource = silly_material_resource.new()
 	# Assign the values.
 	silly_resource.albedo_color = color
 	silly_resource.metallic_strength = metallic
 	silly_resource.roughness_strength = roughness
-
 	return silly_resource
```
AThousandShips:
```suggestion

	return silly_resource
```
Unrelated and hurts readability, should be restored

plugins/addons/material_creator/material_creator.gd
```diff
@@ -41,56 +41,33 @@ func apply_pressed() -> void:
...
-	var silly_resource: Variant = _silly_resource_from_values()
-	# Make a file, store the silly material as a JSON string.
-	var file := FileAccess.open(path, FileAccess.WRITE)
-	file.store_string(silly_resource.make_json())
-
-	return true
+	var silly_resource: Resource = _silly_resource_from_values()
+	# Save the resource as a .tres file using Godot's ResourceSaver
+	var err = ResourceSaver.save(silly_resource, path)
+	return err == OK
```
AThousandShips:
```suggestion
	return ResourceSaver.save(silly_resource, path) == OK
```
No need for a temporary here IMO

plugins/addons/material_creator/material_resource.gd
```diff
@@ -1,15 +1,14 @@
...
-var metallic_strength := 0.0
-var roughness_strength := 0.0
+# Use export to make properties visible and serializable in the inspector and for resource saving/loading.
+@export var albedo_color: Color = Color.BLACK
+@export var metallic_strength: float = 0.0
+@export var roughness_strength: float = 0.0
 
 
+# Create a StandardMaterial3D from the resource's properties.
+=======
```
aaronfranke: `=======` is not valid GDScript.

SarunasRamonas: Thank you. Removed merge conflict markers leftover.

Calinou:
@SarunasRamonas The latest commit in this PR looks unrelated. Did you mean to commit it to another branch?

SarunasRamonas:
> @SarunasRamonas The latest commit in this PR looks unrelated. Did you mean to commit it to another branch?

Yes, you're right, so sorry about that. I have fixed it and reset the branch to the previous commit. Thanks for pointing it out.

plugins/addons/material_creator/material_creator.gd
```diff
@@ -41,53 +41,30 @@ func apply_pressed() -> void:
...
-		$VBoxContainer/MetallicSlider.value = silly_resource.metallic_strength
-		$VBoxContainer/RoughnessSlider.value = silly_resource.roughness_strength
-
-		# Return `true` to indicate success.
-		return true
-
-	# If the file does not exist, then return `false` to indicate failure.
-	return false
+	# Load the resource using Godot's ResourceLoader.
+	var silly_resource: Resource = ResourceLoader.load(path)
```
aaronfranke: You can't use `ResourceLoader` on resources that haven't been imported by an `EditorImportPlugin`.

The problem is that the existing `EditorImportPlugin` only recognizes `*.mtxt` files but the saver saves as `*.silly_mat`, so it never gets imported. This is a pre-existing bug but it was hidden before because it was loading directly before instead of using ResourceImporter.

Arguably what this PR does is the better behavior for the editor, however, I think that both options should be kept around supported. The code in this PR won't work for dynamic runtime loading outside of `res://`, so the existing code is better for that. There can be comments explaining why to prefer one over the other.

SarunasRamonas: Thanks for pointing that out.
At the moment the PR can't actually save/load `.silly_mat` because there's no custom loader/saver registered for that extension, and it also doesn't recognize `.mtxt`.

I'll update PR to:
- add proper support for `.silly_mat` via a custom loader/saver;
- keep `.mtxt` recognized by the importer so both source and resource paths are covered.

Appreciate feedback, I'll get this cleaned up.

SarunasRamonas:
- Add proper support for .silly_mat via a custom loader/saver;
- Keep .mtxt recognized by the importer so both source and resource paths are covered.
403a4c0

aaronfranke (approved):
Thank you for working on this. However, this demo is still in a really bad state. I considered pushing fixes to this PR, but honestly this needs significant enough changes that I'd rather open up a separate PR. This PR is still an improvement so I'll merge this and build on top of your work.

aaronfranke:
I opened PR #1261 as a follow-up to overhaul this plugin demo.

PR #1220, opened 2025-06-24
Add 3D physics interpolation demo

lawnjelly:
Demonstrate first person shooter and third person shooter cameras.

How to do FPS and TPS cameras with physics interpolation comes up regularly, and could do with a demo.

https://github.com/user-attachments/assets/fbf2a13f-a780-404b-acb1-120e88de8fc4

## Notes
* Just read all the requirements, I haven't the strength to edit any more today.
* Anyone else feel free to take this over.
* This is likely 4.5 onward (as the `SceneTreeFTI` changed after 4.4, and 4.4 is no longer recommended)


Calinou (changes requested):

It only works correctly in 4.5.beta1, not 4.4.1.stable (most things aren't interpolated there, except bullets).

Some feedback:

- The folder should be renamed to just `physics_interpolation`, since it's already in the `3d` folder.
- The 3D physics engine should be switched to Jolt Physics in the Project Settings to match existing demos.
- Bullets' collision layer should be changed (as well as the player's collision mask) so that the player doesn't collide with its bullets. This can cause it to suddenly jump if you shoot while moving forward.
- Friction on player movement should only be applied to the X and Z axes. Otherwise, the player slowly descends after a jump as if they're floating. This may require increasing gravity or reducing jump height to compensate (possibly both).
- There should be a shortcut action to toggle physics interpolation, so that you can see its effect in action. It would remain enabled by default nonetheless.

There are a few other changes I'd like to make, but I can work on those in a future PR after this one is merged.

lawnjelly:
> There are a few other changes I'd like to make. If you want, I can open a new PR based on this one.

Yeah that would be great! :+1: 

The particle texture BTW is from one of the other demos as I just needed a texture. That can be anything sensible.

PR #1221, opened 2025-06-24
Update NavigationChunk demos

smix8:
Updates NavigationChunk demos for Godot 4.4 that has the newer `get_bounds()` function and new NavigationAgent no post-processing option.


Calinou (approved):
Tested locally on 4.4.1, it works as expected. Code looks good to me.

Thanks!

Issue #1166, opened 2025-02-16
Voxel demo doesn't scale acceleration or friction by delta


furrykef:
**Which demo project is affected:**
`godot-demo-projects/3d/voxel`

**OS/device including version:**
any

**Issue description:**
In `player/player.gd`, constants such as `MOVEMENT_SPEED_GROUND` and `MOVEMENT_FRICTION_GROUND` are multiplied without being scaled by `delta` (e.g., [here](https://github.com/godotengine/godot-demo-projects/blob/fdb2f50a2088b91d23378009a68c05a8f0a744b9/3d/voxel/player/player.gd#L82)). That means the amount of acceleration or friction will be dependent on the physics time step. The player character will accelerate twice as fast at 120 ticks per second than at 60, for example.

The correct way to do it is to multiply by `MOVEMENT_SPEED_GROUND**delta`, etc. (notice that `delta` is an exponent). This will require changing the constants to reflect the scaling; since the project's physics time step is currently set to 120 ticks per second, the constants would have to be multiplied by 120 to give the same behavior.

Calinou:
Feel free to open a pull request for this 🙂 

See also https://github.com/godotengine/godot-proposals/issues/9803.

PR #1224, opened 2025-06-26
Fix tickrate-dependent physics in Voxel demo

Calinou:
- Fix mouse sensitivity being dependent on window size by using `InputEventMouseMotion.screen_relative`.

Tested on 4.4.1.stable, so this can be cherry-picked.

- This closes https://github.com/godotengine/godot-demo-projects/issues/1166.


3d/voxel/project.godot
```diff
@@ -148,8 +148,8 @@ pick_block={
 [physics]
 
 common/physics_ticks_per_second=120
-3d/default_gravity=20.0
 3d/physics_engine="Jolt Physics"
+3d/default_gravity=20.0
```
aaronfranke, line 152: I'd love to find a solution that doesn't involve gravity being 20 m/s², but this is already a thing from a previous update, so it doesn't have to block this PR.

Calinou: Real world gravity often doesn't make for satisfying character controllers, so you often have to turn it way up in practice (and increase jump height to compensate).

We could increase gravity specifically in the character controller if we don't want to change this project setting.

aaronfranke: I opened PR #1253 to solve this.

PR #1225, opened 2025-06-26
Add a camera feed demo in misc

shiena:
This is a demo project for camera feed and camera server.

## Supported Platforms
- macOS (4.0 or later)
- Linux (4.4 or later)
- Android (4.5 or later)
- iOS (4.7 or later)

## Limitations

Pull requests have been proposed for Windows, Web and iOS.

- ~~https://github.com/godotengine/godot/pull/105476~~
- https://github.com/godotengine/godot/pull/108538
- https://github.com/godotengine/godot/pull/106784

- https://github.com/godot-sdk-integrations/godot-ios-plugins/pull/89

An alternative implementation using PipeWire for Linux.

- https://github.com/godotengine/godot/pull/109500

macOS is incomplete due to bugs.

- https://github.com/godotengine/godot/pull/106777

Linux is incomplete due to bugs.

- https://github.com/godotengine/godot/pull/113358
- https://github.com/godotengine/godot/pull/113359
- https://github.com/godotengine/godot/pull/119852

## Preview
camera_feed


Calinou (changes requested):
Note that the demo requires 4.5 (even on unsupported platforms), as it references the `monitoring_feeds` property which doesn't exist on 4.4.1. This is not essential to fix to merge (it's fine for this demo to be 4.5+), but I thought I'd mention nonetheless.

Some feedback:

- The demo should be switched to the Compatibility rendering method if it doesn't use features exclusive to Forward+/Mobile.
- The demo's folder should be renamed from `camerafeed` to `camera_feed`.
- `application/run/low_processor_mode` should be enabled in the Project Settings to reduce CPU/GPU utilization when idle.

shiena:
@Calinou 
Updated based on review feedback:
- [x] Add require 4.5
- [x] Switch to the Compatibility rendering
- [x] Rename from `camerafeed` to `camera_feed`
- [x] Enable `application/run/low_processor_mode`

Calinou:
The demo should use a lower window size on desktop platforms. Otherwise, on a 1080p display, it'll look like this when run from the editor with game window embedding enabled:

image

You can do this by changing **Window Width Override** and **Window Height Override** to `480` and `854` respectively. This will allow it to fit on a 1080p display. (The setting is ignored on mobile platforms.)

shiena:
- [x] Changed **Window Width Override** and **Window Height Override**

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,180 @@
...
+const defaultWebResolution: Dictionary = {
+	"width": 640,
+	"height": 480
+}
+
+func _ready() -> void:
+	_adjust_ui()
+	_reload_camera_list()
+
+func _adjust_ui() -> void:
```
AThousandShips, line 53:
```suggestion

func _adjust_ui() -> void:
```
Leave two empty lines between methods, for all

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,180 @@
...
+	camera_display.size = camera_display.get_parent_area_size() - Vector2.ONE * 40
+	camera_preview.custom_minimum_size = camera_display.size
+	camera_preview.position = camera_display.size / 2
+
+func _reload_camera_list() -> void:
+	camera_list.clear()
+	format_list.clear()
+
+	var os_name := OS.get_name()
+	# Request camera permission on mobile
```
AThousandShips:
```suggestion
	# Request camera permission on mobile.
```
Per the comment style, for all

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,180 @@
...
+@onready var camera_list := $DrawerContainer/Drawer/DrawerContent/VBoxContainer/CameraList
+@onready var format_list := $DrawerContainer/Drawer/DrawerContent/VBoxContainer/FormatList
+@onready var start_or_stop_button := $DrawerContainer/Drawer/DrawerContent/VBoxContainer/ButtonContainer/StartOrStopButton
+@onready var reload_button := $DrawerContainer/Drawer/DrawerContent/VBoxContainer/ButtonContainer/ReloadButton
+
+var camera_feed: CameraFeed
+
+const defaultWebResolution: Dictionary = {
+	"width": 640,
+	"height": 480
```
AThousandShips:
```suggestion
	"height": 480,
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,180 @@
...
+	if os_name in ["Android", "iOS"]:
+		var permissions = OS.get_granted_permissions()
+		if not "CAMERA" in permissions:
+			if not OS.request_permission("CAMERA"):
+				print("CAMERA permission not granted")
+				return
+
+	if not CameraServer.camera_feeds_updated.is_connected(_on_camera_feeds_updated):
+		CameraServer.camera_feeds_updated.connect(_on_camera_feeds_updated)
+	if CameraServer.monitoring_feeds:
```
AThousandShips:
```suggestion

	if CameraServer.monitoring_feeds:
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,180 @@
...
+		CameraServer.monitoring_feeds = false
+		await get_tree().process_frame
+	CameraServer.monitoring_feeds = true
```
AThousandShips, line 111:
```suggestion

	CameraServer.monitoring_feeds = true
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,187 @@
...
+
+	if not camera_feed.frame_changed.is_connected(_on_frame_changed):
+		camera_feed.frame_changed.connect(_on_frame_changed, ConnectFlags.CONNECT_ONE_SHOT | ConnectFlags.CONNECT_DEFERRED)
+	# Start the feed
+	camera_feed.feed_is_active = true
+
+func _on_frame_changed() -> void:
+	var datatype := camera_feed.get_datatype() as CameraFeed.FeedDataType
+	var preview_size := Vector2.ZERO
+	var mat: ShaderMaterial = camera_preview.material
```
AThousandShips, line 358:
```suggestion

	var mat: ShaderMaterial = camera_preview.material
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,187 @@
...
+	var rgb_texture: CameraTexture = mat.get_shader_parameter("rgb_texture")
+	var y_texture: CameraTexture = mat.get_shader_parameter("y_texture")
+	var cbcr_texture: CameraTexture = mat.get_shader_parameter("cbcr_texture")
+	var ycbcr_texture: CameraTexture = mat.get_shader_parameter("ycbcr_texture")
+	rgb_texture.which_feed = CameraServer.FeedImage.FEED_RGBA_IMAGE
```
AThousandShips, line 364:
```suggestion

	rgb_texture.which_feed = CameraServer.FeedImage.FEED_RGBA_IMAGE
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,187 @@
...
+	y_texture.which_feed = CameraServer.FeedImage.FEED_Y_IMAGE
+	cbcr_texture.which_feed = CameraServer.FeedImage.FEED_CBCR_IMAGE
+	ycbcr_texture.which_feed = CameraServer.FEED_YCBCR_IMAGE
+	match datatype:
```
AThousandShips, line 372:
```suggestion

	match datatype:
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,187 @@
...
+			preview_size = y_texture.get_size()
+		CameraFeed.FeedDataType.FEED_YCBCR:
+			ycbcr_texture.camera_feed_id = camera_feed.get_id()
+			mat.set_shader_parameter("ycbcr_texture", ycbcr_texture)
+			mat.set_shader_parameter("mode", 2)
+			preview_size = ycbcr_texture.get_size()
+		_:
+			print("YCbCr format not fully implemented yet")
+			return
+	var white_image := Image.create(int(preview_size.x), int(preview_size.y), false, Image.FORMAT_RGBA8)
```
AThousandShips, line 400:
```suggestion

	var white_image := Image.create(int(preview_size.x), int(preview_size.y), false, Image.FORMAT_RGBA8)
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,187 @@
...
+	white_image.fill(Color.WHITE)
+	camera_preview.texture = ImageTexture.create_from_image(white_image)
+	var rot := camera_feed.feed_transform.get_rotation()
+	var degree := roundi(rad_to_deg(rot))
+	camera_preview.rotation = rot
+	camera_preview.custom_minimum_size.y = camera_display.size.y
+	if degree % 180 == 0:
```
AThousandShips:
```suggestion

	if degree % 180 == 0:
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,187 @@
...
+		camera_display.ratio = preview_size.x / preview_size.y
+	else:
+		camera_display.ratio = preview_size.y / preview_size.x
+	start_or_stop_button.text = "Stop"
```
AThousandShips, line 405:
```suggestion

	start_or_stop_button.text = "Stop"
```

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,187 @@
...
+			mat.set_shader_parameter("ycbcr_texture", ycbcr_texture)
+			mat.set_shader_parameter("mode", 2)
+			preview_size = ycbcr_texture.get_size()
+		_:
+			print("YCbCr format not fully implemented yet")
+			return
+	var white_image := Image.create(int(preview_size.x), int(preview_size.y), false, Image.FORMAT_RGBA8)
+	white_image.fill(Color.WHITE)
+	camera_preview.texture = ImageTexture.create_from_image(white_image)
+	var rot := camera_feed.feed_transform.get_rotation()
```
AThousandShips:
```suggestion

	var rot := camera_feed.feed_transform.get_rotation()
```

shiena:
@AThousandShips Address review comments.

Calinou:
I've tested the demo on macOS with 4.5.stable and I noticed that while the camera name appears, I can't choose any format:

Screenshot 2025-09-24 at 17 36 37

After clicking **Reload**, I see a permission prompt. When I accept it, the LED next to the webcam lights up, but I still don't see anything. I don't see any errors in the terminal. This occurs regardless of how I'm running Godot (it's the official 4.5 .app bundle, I've tried running it from a terminal but also from Spotlight). Maybe I'm missing an entitlement? cc @bruvzg

shiena:
@Calinou 
Format is currently not supported. You will be able to select from a list once https://github.com/godotengine/godot/pull/106777 is merged.
image

bruvzg:
Current entitlements should be fine, but if you are using multiple bundles with the same id at the same (and all versions of Godot have the same id) permission can sometimes stop working. Removing it from the allowed app list in the system settings (minus button) completly and restarting the app should fix it.

aaronfranke:
Thank you for this demo, I've added the "waiting for Godot" label because I think it's probably worth waiting for more platforms to have camera feeds supported in the engine's code before merging, otherwise users will get confused and/or annoyed wondering why it doesn't work.

CsloudX:
Godot4.5 stable was released, can merge this demo now?

Calinou:
> Godot4.5 stable was released, can merge this demo now?

As seen in OP, pull requests for Windows, iOS and Web support have yet to be merged. I'd wait until the Windows PR is merged at least, as well as the PR that fixes issues in the current macOS implementation.

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,320 @@
...
+		aspect_container.ratio = preview_size.x / preview_size.y
+	else:
+		aspect_container.ratio = preview_size.y / preview_size.x
+
+
+func _get_preview_size(mat: ShaderMaterial) -> Vector2:
+	var datatype := camera_feed.get_datatype() as CameraFeed.FeedDataType
+	match datatype:
+		CameraFeed.FeedDataType.FEED_RGB:
+			var texture: CameraTexture = mat.get_shader_parameter("rgb_texture")
```
AThousandShips:
```suggestion
			var texture: CameraTexture = mat.get_shader_parameter(&"rgb_texture")
```
For all `get_shader_parameter`, it takes `StringName`

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,320 @@
...
+	cbcr_texture.which_feed = CameraServer.FeedImage.FEED_CBCR_IMAGE
+	ycbcr_texture.which_feed = CameraServer.FEED_YCBCR_IMAGE
+
+	var datatype := camera_feed.get_datatype() as CameraFeed.FeedDataType
+	var preview_size := Vector2.ZERO
+
+	match datatype:
+		CameraFeed.FeedDataType.FEED_RGB:
+			rgb_texture.camera_feed_id = camera_feed.get_id()
+			mat.set_shader_parameter("rgb_texture", rgb_texture)
```
AThousandShips:
```suggestion
			mat.set_shader_parameter(&"rgb_texture", rgb_texture)
```
And for all these too

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,320 @@
...
+		return
+	if _cached_formats.is_empty():
+		return
+
+	var mat: ShaderMaterial = camera_preview.material
+	if not mat:
+		return
+
+	var preview_size := _get_preview_size(mat)
+	if preview_size.round() <= Vector2.ZERO:
```
AThousandShips: This might not be desired, this will not hit if the rounded value is `(1, 0)` for example, maybe instead:
```suggestion
	if preview_size.round().x <= 0 or preview_size.round().y <= 0:
```
Or similar

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,320 @@
...
+		CameraFeed.FeedDataType.FEED_YCBCR:
+			ycbcr_texture.camera_feed_id = camera_feed.get_id()
+			mat.set_shader_parameter("ycbcr_texture", ycbcr_texture)
+			mat.set_shader_parameter("mode", ShaderMode.YCBCR)
+			preview_size = ycbcr_texture.get_size()
+		_:
+			print("Skip formats that are not supported.")
+			return
+
+	if preview_size.round() <= Vector2.ZERO:
```
AThousandShips: Same here

shiena: Applied all review comments.

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,320 @@
...
+
+
+func _reload_camera_list() -> void:
+	camera_list.clear()
+	format_list.clear()
+
+	# Request camera permission on Android
+	if OS.get_name() == "Android":
+		if not "CAMERA" in OS.get_granted_permissions():
+			if not OS.request_permission("CAMERA"):
```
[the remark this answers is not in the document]
The-Cyber-Captain: Or, obviously, `assert(signaled_result[1] == true, "CAMERA permission not granted. Bailing out.")` 🤣 
I switched from warning to assert at the last second there, after converting everything to international American spelling as well.

Right. Merry Xmas y'all! 🎅

misc/camera_feed/camerafeed.gd
```diff
@@ -0,0 +1,426 @@
...
+	get_tree().root.content_scale_factor = scale_factor
+
+
+func _reload_camera_list() -> void:
+	camera_list.clear()
+	format_list.clear()
+
+	# Request camera permission on Android
+	if OS.get_name() == "Android":
+		if not "CAMERA" in OS.get_granted_permissions():
```
AThousandShips, line 99:
```suggestion
		if "CAMERA" not in OS.get_granted_permissions():
```

The-Cyber-Captain: Android permissions - and our current API for them - are... awkward beasts. I attempted to clarify the docs on them a while back, but basically it just flagged the API as a legacy problem. https://github.com/godotengine/godot/pull/105364

TL;DR `"CAMERA" not in OS.get_granted_permissions()` will always return true. **The string needs to be "android.permission.CAMERA"**. Technically, request_permission is also documented as requiring the full permission... but because of a bug (or implementation) it currently passes without. 🤷

All that said, this approach itself - which does give us a great little demo in general! - excludes the ability to use one-click deploy under Android; given the perms will always be wiped and there's no time given for the user to accept them. I would recommend also switching the logic to leverage something like

	if OS.get_name() == "Android":
		if "android.permission.CAMERA" not in OS.get_granted_permissions():
			OS.request_permission("android.permission.CAMERA")
			var signaled_result = await get_tree().on_request_permissions_result
			assert(signaled_result[1] == true, "CAMERA permission not granted. Bailing out.")
			
It's a little janky - it should really be a loop in case some other permission is requested / granted first - but solid enough here for the demo purposes... under the current permissions API constraints. (IMO it's super-important that we try keep content like the demos approachable / accessible so as not to scare off folks new to the features or platform.)

shiena: Thanks for the detailed feedback! You're absolutely right about the permission handling issues.

I've updated the implementation with the following changes:

1. Fixed permission string: Now using "android.permission.CAMERA" for get_granted_permissions() check
2. Added permission request UI: Instead of silently failing, the app now shows a dedicated permission panel with a "Grant Permission" button
3. Proper async handling: Using await get_tree().on_request_permissions_result with a loop to handle cases where other permissions might be signaled first:

This should work properly with one-click deploy now, as the user is given time to accept the permission before the camera initialization proceeds.

PR #1229, opened 2025-07-07
Add FoldableContainer to the Control Gallery demo

Calinou:
This bumps the demo requirement to Godot 4.5, as FoldableContainer is only present in 4.5 and later.

## Preview

control_gallery webp


skyace65:
Thanks!

PR #1231, opened 2025-07-13
Simplify mesh custom drawing demo

Rubonnek:
I've removed the comment about the `MultiMesh` crash and opened https://github.com/godotengine/godot/issues/108566.

### DIFF
--- a/2d/custom_drawing/meshes.gd
+++ b/2d/custom_drawing/meshes.gd
@@ -3,7 +3,7 @@
 extends Panel
 
 # You must hold a reference to the Resources either as member variables or within an Array or Dictionary.
-# Otherwise they get freed automatically and the renderer won't be able to draw them.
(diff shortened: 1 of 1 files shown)


2d/custom_drawing/meshes.gd
```diff
@@ -2,55 +2,30 @@
...
-var use_antialiasing := false
-
-
-func _draw() -> void:
-	var margin := Vector2(300, 70)
-
-	var offset := Vector2()
-	var text_mesh := TextMesh.new()
+# You must hold a reference to the Resources either as member variables or within an Array or Dictionary.
+# Otherwise they get freed automatically and the renderer won't be able to draw them.
```
Calinou:
```suggestion
# Otherwise, they get freed automatically and the renderer won't be able to draw them.
```

Calinou (approved):

PR #1236, opened 2025-08-04
Minor spelling fixes in strings, comments and non-code files.

chrisl8:
This is a set of hand picked and entered (no LLM's here) spelling fixes in only strings, comments and non-code files.

No functional changes should exist here: I avoided changing any variables or other code based spellings.

These should be non-controversial: I didn't get into yak shaving about hyphenated words or industry terms. Just the "normal words", and I tried to avoid changing anything that was just an acceptable alternate based on national origin (i.e. US vs British).

Credit to `cspell`: I used `cspell` to find these to start with, but the changes were hand picked based on being good targets and then hand-fixed to make sure they made sense in context and had no affect on code operation.

As justification: Finding a lot of spelling mistakes in comments and descriptions causes some issues for some people such as:
1. Because of how some people read, it can trip them up and slow them down.
2. It can make what is otherwise high quality code feel "amateurish" even though it is not.

This is by no means exhaustive, but hopefully meaningful.

_This is all because I ran into one spelling error and wanted to fix it, but I felt it would waste everyone's time to make a PR for one fix. hopefully this is worth the time._

### DIFF
--- a/3d/ik/addons/sade/ik_fabrik.gd
+++ b/3d/ik/addons/sade/ik_fabrik.gd
@@ -5,7 +5,7 @@ extends Node3D
 
 # The delta/tolerance for the bone chain (how do the bones need to be before it is considered satisfactory)
 const CHAIN_TOLERANCE = 0.01
-# The amount of interations the bone chain will go through in an attempt to get to the target position
+# The amount of iterations the bone chain will go through in an attempt to get to the target position
 const CHAIN_MAX_ITER = 10
 
 @export var skeleton_path: NodePath:
@@ -87,7 +87,7 @@ var chain_origin = Vector3()
 # The combined length of every bone in the bone chain
 var total_length = INF
 # The amount of iterations we've been through, and whether or not we want to limit our solver to CHAIN_MAX_ITER
-# amounts of interations.
+# amounts of iterations.
 @export var chain_iterations: int = 0
 @export var limit_chain_iterations: bool = true
 # Should we reset chain_iterations on movement during our update method?
@@ -151,7 +151,7 @@ func _ready():
 	update_mode = update_mode
 
 
-# Various upate methods
+# Various update methods
 func _process(_delta):
 	if reset_iterations_on_update:
 		chain_iterations = 0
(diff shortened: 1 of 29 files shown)


3d/ik/addons/sade/ik_fabrik.gd
```diff
@@ -5,7 +5,7 @@ extends Node3D
 
 # The delta/tolerance for the bone chain (how do the bones need to be before it is considered satisfactory)
 const CHAIN_TOLERANCE = 0.01
-# The amount of interations the bone chain will go through in an attempt to get to the target position
+# The amount of iterations the bone chain will go through in an attempt to get to the target position
```
AThousandShips:
```suggestion
# The number of iterations the bone chain will go through in an attempt to get to the target position
```

3d/ik/addons/sade/ik_fabrik.gd
```diff
@@ -87,7 +87,7 @@ var chain_origin = Vector3()
 # The combined length of every bone in the bone chain
 var total_length = INF
 # The amount of iterations we've been through, and whether or not we want to limit our solver to CHAIN_MAX_ITER
-# amounts of interations.
+# amounts of iterations.
```
AThousandShips:
```suggestion
# iterations.
```

compute/texture/water_plane/water_compute.glsl
```diff
@@ -24,7 +24,7 @@ void main() {
 
 	ivec2 uv = ivec2(gl_GlobalInvocationID.xy);
 
-	// Just in case the texture size is not divisable by 8.
+	// Just in case the texture size is not devisable by 8.
```
AThousandShips: This is incorrect, it's "able to be divided", which is "divisible", not "possible to devised or invented"
```suggestion
	// Just in case the texture size is not divisible by 8.
```

gui/gd_paint/tools_panel.gd
```diff
@@ -68,7 +68,7 @@ func button_pressed(button_name: String) -> void:
 		paint_control.brush_shape = paint_control.BrushShape.CIRCLE
 		shape_name = "Circle"
 
-	# If a opperation button is pressed
+	# If an operation button is pressed
```
AThousandShips:
```suggestion
	# If an operation button is pressed.
```
While we're here

aaronfranke (approved):
Looked over the changes, all seem correct now that ATS's patch has been applied.

PR #1085, opened 2024-07-09
Add Navigation AStar Hexagonal demo

m21-cerutti:
Demo for showing how to use AStar2D with hexagonal map with minimal code.

Works with v4.2.2.stable.official [15073afe3].

PR #1239, opened 2025-08-25
hexagonal_map: Add turn-based movement mode

idbrii:
Games with hexagonal maps usually want characters to move between tiles on the grid, so I added another movement option that makes the troll from tile to tile like a turn-based hex game. It still uses WASD controls (you need to press multiple keys to hit all 6 directions). Works well on gamepad too.

Additionally, the current `move_and_slide` troll doesn't really interact with the world, so I made it stop before leaving the grid. This also acts as a demonstration of how to get the tile id for interacting with objects represented by valid tiles.

I also converted TileMap to TileMapLayer.

No untyped_declaration warnings.

Demo:

https://github.com/user-attachments/assets/845a8c95-fc89-4242-a1b8-664f9015075f

### DIFF
--- a/2d/hexagonal_map/troll_tile.gd
+++ b/2d/hexagonal_map/troll_tile.gd
@@ -64,9 +64,23 @@ func _input_dir_to_neighbor(input_dir: Vector2) -> int:
 	return CELL_NEIGHBOR_INVALID
 
 
+func _input(event: InputEvent) -> void:
+	if event is InputEventMouseMotion:
+		var camera = get_viewport().get_camera_2d()
+		var offset: Vector2 = camera.get_global_mouse_position() - global_position
+		# Ignore input on top of our troll because it makes the pointer unpredictable.
+		const DEADZONE_SQ = 25.0 * 25.0
+		if offset.length_squared() > DEADZONE_SQ:
+			# Normalize to indicate full movement towards the mouse.
+			handle_input(offset.normalized())
+
+
 func _process(_dt: float) -> void:
 	var motion: Vector2 = Input.get_vector(&"move_left", &"move_right", &"move_up", &"move_down")
+	handle_input(motion)
 
+
+func handle_input(motion: Vector2):
 	# InputMap has a hefty deadzone (aiming is discrete) so can just check for zero.
 	if not motion.is_zero_approx():
 		# If your perspective was closer to isometric, you should rotate the input
--- a/2d/hexagonal_map/project.godot
+++ b/2d/hexagonal_map/project.godot
@@ -64,6 +64,7 @@ move_confirm={
 "deadzone": 0.2,
 "events": [Object(InputEventJoypadButton,"resource_local_to_scene":false,"resource_name":"","device":-1,"button_index":0,"pressure":0.0,"pressed":true,"script":null)
 , Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":32,"key_label":0,"unicode":32,"location":0,"echo":false,"script":null)
+, Object(InputEventMouseButton,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"button_mask":1,"position":Vector2(171, 18),"global_position":Vector2(185, 80),"factor":1.0,"button_index":1,"canceled":false,"pressed":true,"double_click":false,"script":null)
 ]
 }
 toggle_movement_method={


2d/hexagonal_map/gridlayer.gd
```diff
@@ -0,0 +1,33 @@
...
+
+
+func get_cell_half_size() -> Vector2:
+	return tile_set.tile_size / 2
+
+
+# Returns a global position that's the bottom corner of a tile.
+# https://www.reddit.com/r/godot/comments/fuejci/having_trouble_with_snapping_to_isometric_grid/fmcl07f/
+func snap_global_to_cell(world_pos: Vector2) -> Vector2:
+	# map_to_local says it's the centre, but it ignores texture_origin.
```
AThousandShips:
```suggestion
	# map_to_local says it's the center, but it ignores texture_origin.
```
Please use US English

2d/hexagonal_map/gridlayer.gd
```diff
@@ -0,0 +1,33 @@
...
+
+func global_to_cell_pos(world_pos: Vector2) -> Vector2i:
+	return local_to_map(to_local(world_pos))
+
+
+func cell_to_global_pos(cell_pos: Vector2i) -> Vector2:
+	return to_global(map_to_local(cell_pos))
+
+
+# Change visual appearance of tile at position v. Select a TileMapLayer and
```
AThousandShips:
```suggestion
# Change the visual appearance of tile at position v. Select a TileMapLayer and
```

2d/hexagonal_map/troll_tile.gd
```diff
@@ -0,0 +1,113 @@
...
+
+# Map the input to a world-space position.
+func input_dir_to_dest(input_dir: Vector2) -> Vector2:
+	var neighbor := _input_dir_to_neighbor(input_dir)
+	if neighbor != -1:
+		# neighbor is a CellNeighbor.
+		var current_cell : Vector2i = gridlayer.global_to_cell_pos(global_position)
+		var cell := gridlayer.get_neighbor_cell(current_cell, neighbor)
+		return gridlayer.cell_to_global_pos(cell)
+	return global_position
```
AThousandShips, line 29:
```suggestion

	return global_position
```

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,113 @@
...
+		return TileSet.CellNeighbor.CELL_NEIGHBOR_BOTTOM_LEFT_SIDE
+	if matches_grid_angle(input_dir, Vector2.DOWN.rotated(HEX_DELTA * 2)):
+		return TileSet.CellNeighbor.CELL_NEIGHBOR_TOP_LEFT_SIDE
+	if matches_grid_angle(input_dir, Vector2.DOWN.rotated(HEX_DELTA * 3)):
+		return TileSet.CellNeighbor.CELL_NEIGHBOR_TOP_SIDE
+	if matches_grid_angle(input_dir, Vector2.DOWN.rotated(HEX_DELTA * 4)):
+		return TileSet.CellNeighbor.CELL_NEIGHBOR_TOP_RIGHT_SIDE
+	if matches_grid_angle(input_dir, Vector2.DOWN.rotated(HEX_DELTA * 5)):
+		return TileSet.CellNeighbor.CELL_NEIGHBOR_BOTTOM_RIGHT_SIDE
+	printt("_input_dir_to_neighbor: failed to find a match.")
```
AThousandShips:
```suggestion
	print("_input_dir_to_neighbor: Failed to find a match.")
```
No need to use `printt` with a single argument

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,113 @@
...
+	return -1
+
+
+func _process(_dt: float) -> void:
+	var motion := Input.get_vector("move_left", "move_right", "move_up", "move_down")
+
+	# If your perspective was closer to isometric, you should rotate the input
+	# to match the iso projection so you correctly interpret diagonal inputs:
+	# var iso_motion := motion.rotated(TAU * 1/10)
```
AThousandShips:
```suggestion
	#var iso_motion := motion.rotated(TAU * 1 / 10)
```
No space before commented out code

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,113 @@
...
+		var has_tile := dest_tile >= 0
+
+		var r_dot := delta.dot(Vector2.RIGHT)
+		var flip_h := r_dot <= 0
+		if absf(r_dot) > 0 and flip_h != current_facing_flip:
+			current_facing_flip = flip_h
+			for sprite in $Visual.get_children():
+				sprite.flip_h = flip_h
+
+
```
AThousandShips:
```suggestion
```

2d/hexagonal_map/trolltoggle.gd
```diff
@@ -0,0 +1,25 @@
...
+
+func spawn_next_troll() -> void:
+    troll = troll_list[next_idx]
+    add_child(troll)
+    next_idx += 1
+    next_idx = next_idx % troll_list.size()
+
+
+func _process(_dt : float) -> void:
+    if Input.is_action_just_pressed("toggle_movement_method"):
```
AThousandShips:
```suggestion
    if Input.is_action_just_pressed(&"toggle_movement_method"):
```

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,113 @@
...
+	if matches_grid_angle(input_dir, Vector2.DOWN.rotated(HEX_DELTA * 4)):
+		return TileSet.CellNeighbor.CELL_NEIGHBOR_TOP_RIGHT_SIDE
+	if matches_grid_angle(input_dir, Vector2.DOWN.rotated(HEX_DELTA * 5)):
+		return TileSet.CellNeighbor.CELL_NEIGHBOR_BOTTOM_RIGHT_SIDE
+	printt("_input_dir_to_neighbor: failed to find a match.")
+	return -1
+
+
+func _process(_dt: float) -> void:
+	var motion := Input.get_vector("move_left", "move_right", "move_up", "move_down")
```
AThousandShips:
```suggestion
	var motion := Input.get_vector(&"move_left", &"move_right", &"move_up", &"move_down")
```

2d/hexagonal_map/realtime_troll.gd
```diff
@@ -0,0 +1,31 @@
...
+
+
+@export var gridlayer : TileMapLayer
+
+
+func _physics_process(_delta: float) -> void:
+	var prev_position := global_position
+
+	var motion := Vector2()
+	motion.x = Input.get_axis("move_left", "move_right")
```
AThousandShips:
```suggestion
	motion.x = Input.get_axis(&"move_left", &"move_right")
```

2d/hexagonal_map/realtime_troll.gd
```diff
@@ -0,0 +1,31 @@
...
+	motion.y = Input.get_axis("move_up", "move_down")
```
AThousandShips:
```suggestion
	motion.y = Input.get_axis(&"move_up", &"move_down")
```

2d/hexagonal_map/gridlayer.gd
```diff
@@ -0,0 +1,33 @@
+extends TileMapLayer
+
+
+func get_cell_half_size() -> Vector2:
+	return tile_set.tile_size / 2
```
aaronfranke: This function is unused.

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,121 @@
+extends Node2D
+
+const CELL_NEIGHBOR_INVALID : int = -1
+
+@export var gridlayer : TileMapLayer
+@export var move_duration : float = 0.5
+
+var block_input := false
+var current_facing_flip := false
```
aaronfranke:
```suggestion
const CELL_NEIGHBOR_INVALID: int = -1

@export var gridlayer: TileMapLayer
@export var move_duration: float = 0.5

var block_input: bool = false
var current_facing_flip: bool = false
```

2d/hexagonal_map/trolltoggle.gd
```diff
@@ -0,0 +1,29 @@
+extends Node2D
+
+var next_idx : int = 0
+var troll : Node2D
+
+@onready var troll_list := get_children()
```
aaronfranke: Prefer using full names instead of abbreviations

```suggestion
var next_index: int = 0
var troll: Node2D

@onready var troll_list: Array[Node] = get_children()
```

2d/hexagonal_map/trolltoggle.gd
```diff
@@ -0,0 +1,29 @@
...
+
+
+func spawn_next_troll() -> void:
+    troll = troll_list[next_idx]
+    add_child(troll)
+    next_idx += 1
+    next_idx = next_idx % troll_list.size()
+
+
+func _process(_dt : float) -> void:
```
aaronfranke: Godot's default argument name is `delta`, so the unused equivalent is `_delta`. There's no need to shorten this. Especially for an official demo, we don't want to confuse users.

```suggestion
func _process(_delta: float) -> void:
```

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,121 @@
...
+		var cell := gridlayer.get_neighbor_cell(current_cell, neighbor)
+		return gridlayer.cell_to_global_pos(cell)
+	return global_position
+
+
+func _matches_grid_angle(input_dir: Vector2, direction: Vector2) -> bool:
+	# Divide circle of possible inputs into one segment for each side of our
+	# grid shape.
+	const SEGMENT := TAU/6 * 0.5
+	var a : float = input_dir.angle_to(direction)
```
aaronfranke:
```suggestion
	const SEGMENT: float = TAU / 6.0 * 0.5
	var a: float = input_dir.angle_to(direction)
```

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,121 @@
...
+
+func _process(_dt: float) -> void:
+	var motion := Input.get_vector("move_left", "move_right", "move_up", "move_down")
+
+	# If your perspective was closer to isometric, you should rotate the input
+	# to match the iso projection so you correctly interpret diagonal inputs:
+	# var iso_motion := motion.rotated(TAU * 1/10)
+	var iso_motion := motion
+	# Use a hefty deadzone since movement is discrete.
+	const DEADZONE_SQ = 0.3*0.3
```
aaronfranke:
```suggestion
	const DEADZONE_SQ = 0.3 * 0.3
```

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,121 @@
...
+	var want_movement := iso_motion.length_squared() > DEADZONE_SQ
+	$DestinationPointer.visible = want_movement
+	if want_movement:
+		var dest : Vector2 = _input_dir_to_dest(iso_motion)
+		var delta : Vector2 = dest - global_position
+
+		$DestinationPointer.rotation = -delta.angle_to(Vector2.RIGHT)
```
aaronfranke: Don't use `delta` for anything other than the argument of `_process` or `_physics_process` to avoid confusing users. `$DestinationPointer` should be cached as `@onready var destination_pointer = $DestinationPointer`. `-something.angle_to(Vector2.RIGHT)` can be simplified to `something.angle()`.

```suggestion
		var destination: Vector2 = _input_dir_to_dest(iso_motion)
		var displacement: Vector2 = destination - global_position

		destination_pointer.rotation = displacement.angle()
```

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,121 @@
...
+
+		if block_input:
+			return
+
+		dest = gridlayer.snap_global_to_cell(dest)
+		var dest_tile : int = gridlayer.get_world_tile(dest)
+		var has_tile := dest_tile >= 0
+
+		var r_dot := delta.dot(Vector2.RIGHT)
+		var flip_h := r_dot <= 0
```
aaronfranke: `something.dot(Vector2.RIGHT)` can be simplified to `something.x`

```suggestion
		var destination_tile: int = grid_layer.get_world_tile(destination)
		var has_tile: bool = destination_tile >= 0
		var flip_h: bool = displacement.x <= 0
```

2d/hexagonal_map/turnbased_troll.gd
```diff
@@ -0,0 +1,121 @@
...
+			t = t.from_current()
+			t = t.as_relative()
+			t = t.set_ease(Tween.EASE_IN_OUT)
+			t = t.set_trans(Tween.TRANS_SINE)
+
+		else:
+			# Bounce off collision to give feedback that we recognized the input but it was invalid.
+			var duration := move_duration * 0.1
+			var current_pos := global_position
+			var t := tween.tween_property(self, "global_position", delta * 0.1, duration)
```
aaronfranke: Generally try to avoid single letter variables, but in this case `t` is a good name because it is used so many times and it is a variable local to this scope, so the mental overhead is extremely low.

```suggestion
			var duration: float = move_duration * 0.1
			var current_pos: Vector2 = global_position
			var t: PropertyTweener = tween.tween_property(self, "global_position", delta * 0.1, duration)
```

aaronfranke (changes requested):
I would not call these turn-based and real-time movement. These should be called tile-based and free movement. You can have movement be done on tiles without being turns, and being real-time doesn't inherently indicate that movement is freeform.

The code needs to follow the GDScript style guide: https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html

All files need to be `snake_case`, so for example, `gridlayer.gd` needs to be renamed to `grid_layer.gd`.

aaronfranke:
See also PR #1085 which adds hexagonal navigation, I wonder if all of these should be in one demo.

2d/hexagonal_map/turnbased_troll.gd
idbrii: Can remove the spaces, but the style guide says to [prefer inference for assignment](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html#inferred-types):

> Prefer := when the type is written on the same line as the assignment, otherwise prefer writing the type explicitly.

aaronfranke: It says "when the type is written on the same line as the assignment".

The line `var block_input := false` does not contain the text `bool` already.

2d/hexagonal_map/trolltoggle.gd
idbrii: [The docs say](https://docs.godotengine.org/en/stable/classes/class_stringname.html) this conversion happens automatically at compile time.

>You will usually pass a [String](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string) to methods expecting a StringName and it will be automatically converted (often at compile time), but in rare cases you can construct a StringName ahead of time with the StringName constructor or, in GDScript, the literal syntax &"example".

Using the `&` seems confusing to users when it's the same as a simple string.

Looks like the demos are mostly using `&` (75%), so I'll bring that up as a separate issue and stick with convention here.

2d/hexagonal_map/turnbased_troll.gd
idbrii: Oh. I read that to mean "when you can see the type in the assignment". Will change.

idbrii:
Applied all the style changes and renaming to "tile-based and free movement".

I believe I correctly merged with #1258 (kept uid from master).

> See also PR https://github.com/godotengine/godot-demo-projects/pull/1085 which adds hexagonal navigation, I wonder if all of these should be in one demo.

I agree it'd make sense to combine them in the future. Maybe once this PR is merged that PR could add their features to this demo?

aaronfranke:
With the current behavior of the demo, I find it very difficult to move up-left or down-right. If I try to move down-right, and I don't press the keys at *exactly* the same time, I end up moving down or right. I think it should not try to move when the input direction is close to being in-between two others, such as a right input being ambiguous between up-right and down-right, and instead a right input should wait until an up or down input is received.

idbrii:
@aaronfranke  What if instead, you select a direction and then press a key to confirm movement? I think that's more classic hex movement behaviour. Mouse picking an adjacent destination would also be more typical (but maybe for a future PR).

I wasn't sure how much complexity is okay in these demos, but if that's fine then I think the more expected movement for hex is a better choice.

idbrii:
@aaronfranke I've updated the tile troll movement so you press Space/A to confirm movement.

idbrii:
I added mouse aiming too. It was easier than expected!

2d/hexagonal_map/troll_tile.gd
```diff
@@ -0,0 +1,136 @@
...
+			t = t.from_current()
+			t = t.as_relative()
+			t = t.set_ease(Tween.EASE_IN_OUT)
+			t = t.set_trans(Tween.TRANS_SINE)
+
+		else:
+			# Bounce off collision to give feedback that we recognized the input but it was invalid.
+			var duration: float = move_duration * 0.1
+			var current_pos: Vector2 = global_position
+			var t: PropertyTweener = tween.tween_property(self, "global_position", displacement * 0.1, duration)
```
AThousandShips:
```suggestion
			var t: PropertyTweener = tween.tween_property(self, ^"global_position", displacement * 0.1, duration)
```

2d/hexagonal_map/troll_tile.gd
```diff
@@ -0,0 +1,136 @@
...
+			t = t.from_current()
+			t = t.as_relative()
+			t = t.set_ease(Tween.EASE_IN_OUT)
+			t = t.set_trans(Tween.TRANS_SINE)
+			t = tween.chain().tween_property(self, "global_position", current_pos, duration)
```
AThousandShips:
```suggestion
			t = tween.chain().tween_property(self, ^"global_position", current_pos, duration)
```

PR #1242, opened 2025-09-05
Add a custom logging demo

Calinou:
- Demo counterpart of https://github.com/godotengine/godot-docs/pull/11266.
- Follow-up to https://github.com/godotengine/godot/pull/91006.

This demo showcases a custom logger implementation, which runs in parallel to the built-in logging facilities (including file logging). The custom logger displays all messages printed by the engine in an in-game console.

## Preview

custom_logging


misc/custom_logging/main.tscn
```diff
@@ -0,0 +1,120 @@
...
+anchor_top = 1.0
+anchor_bottom = 1.0
+offset_left = 784.0
+offset_top = -103.0
+offset_right = 981.0
+offset_bottom = -80.0
+grow_vertical = 0
+tooltip_text = "Flushing standard output on print allows prints to be visible immediately
+in log files, even if the engine crashes or is killed by the user.
+This has a performance impact when prints are frequently performed.
```
AThousandShips:
```suggestion
This has a performance impact when printing frequently.
```

misc/custom_logging/README.md
```diff
@@ -0,0 +1,16 @@
+# Custom Logging
+
+This demo showcases a custom logger implementation, which runs in parallel
+to the built-in logging facilities (including file logging). The custom logger
```
AThousandShips:
```suggestion
with the built-in logging facilities (including file logging). The custom logger
```
Or "runs parallel to" without "in"

misc/custom_logging/custom_logger_ui.gd
```diff
@@ -0,0 +1,56 @@
...
+		for backtrace in script_backtraces:
+			script_backtraces_text += backtrace.format(trace_indent - 3) + "\n"
+
+		CustomLoggerUI.get_node("Panel/RichTextLabel").text += "%s %s %s[/color]\n[color=#999]%s[/color]\n[color=#999]%s[/color]" % [
+			prefix,
+			code,
+			rationale,
+			trace,
+			script_backtraces_text,
+		]
```
AThousandShips:
```suggestion
				prefix,
				code,
				rationale,
				trace,
				script_backtraces_text,
			]
```

mhilbrunner (approved):

AThousandShips (approved):

mihe:
`CustomLogger` is not taking into account that `_log_error` and `_log_message` can be called from non-main threads (and multiple threads at the same time) so will most likely end up with a race condition when appending to the `RichTextLabel`, or perhaps tripping some thread guard error somewhere.

I guess deferring the append to the main thread would be the best option in this case, ideally with some comment explaining the thread-safety gotcha.

EDIT: It might be a good idea to add a `OS.remove_logger` call somewhere in there as well, just to show that it exists, since the user might "leak" stuff otherwise.

Calinou:
> I guess deferring the append to the main thread would be the best option in this case

That makes sense, but I'm not sure how to do it in GDScript. Is it just a matter of using `call_deferred()`?

mihe:
> Is it just a matter of using `call_deferred()`?

Yep!

Calinou:
`call_deferred()` is now used, and I've made use of `OS.remove_logger()` too. I've added a comment at the top of the CustomLogger class to emphasize the importance of thread safety.

PR #1241, opened 2025-08-26
Fix water plane writable uniform set

beicause:
Fixes https://github.com/godotengine/godot/issues/109962

PR #1247, opened 2025-09-13
Fix uniform sets in "Compute Texture"

0x1hue:
Fix uniform sets that were hardcoded as `0`, which seems to work with `v4.4` but has broken this demo in `v4.5`.

Still works with `v4.4.1` ✅ 

Also removed the unnecessary texture usage bits.

---

```
  ERROR: Uniforms supplied for set (2):
  ERROR: Set: 0 Binding: 0 Type: Image Writable: N Length: 1
  ERROR: are not the same format as required by the pipeline shader. Pipeline shader requires the following bindings:
  ERROR: Set: 0 Binding: 0 Type: Image Writable: N Length: 1
  ERROR: Set: 1 Binding: 0 Type: Image Writable: N Length: 1
  ERROR: Set: 2 Binding: 0 Type: Image Writable: Y Length: 1
```

### DIFF
--- a/compute/texture/water_plane/water_plane.gd
+++ b/compute/texture/water_plane/water_plane.gd
@@ -143,13 +143,13 @@ var pipeline: RID
 var texture_rds: Array[RID] = [RID(), RID(), RID()]
 var texture_sets: Array[RID] = [RID(), RID(), RID()]
 
-func _create_uniform_set(texture_rd: RID) -> RID:
+func _create_uniform_set(texture_rd: RID, uniform_set: int) -> RID:
 	var uniform := RDUniform.new()
 	uniform.uniform_type = RenderingDevice.UNIFORM_TYPE_IMAGE
 	uniform.binding = 0
 	uniform.add_id(texture_rd)
 	# Even though we're using 3 sets, they are identical, so we're kinda cheating.
-	return rd.uniform_set_create([uniform], shader, 0)
+	return rd.uniform_set_create([uniform], shader, uniform_set)
 
 
 func _initialize_compute_code(init_with_texture_size: Vector2i) -> void:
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Tested locally on 4.4.1.stable and 4.5.rc2, it works as expected.

clayjohn:
I think this supersedes https://github.com/godotengine/godot-demo-projects/pull/1241 and https://github.com/godotengine/godot-demo-projects/pull/1174

Calinou:
> I think this supersedes #1241 and #1174

I can confirm this, so I went and closed the other PRs.

compute/texture/water_plane/water_plane.gd
```diff
@@ -143,13 +143,13 @@ var pipeline: RID
 var texture_rds: Array[RID] = [RID(), RID(), RID()]
 var texture_sets: Array[RID] = [RID(), RID(), RID()]
 
-func _create_uniform_set(texture_rd: RID) -> RID:
+func _create_uniform_set(texture_rd: RID, uniform_set: int) -> RID:
 	var uniform := RDUniform.new()
 	uniform.uniform_type = RenderingDevice.UNIFORM_TYPE_IMAGE
 	uniform.binding = 0
 	uniform.add_id(texture_rd)
 	# Even though we're using 3 sets, they are identical, so we're kinda cheating.
```
beicause, line 151: The comment can be removed

compute/texture/water_plane/water_plane.gd
```diff
@@ -215,9 +210,15 @@ func _render_process(with_next_texture: int, wave_point: Vector4, tex_size: Vect
...
-	var previous_set := texture_sets[(with_next_texture - 2) % 3]
+	# Figure out which texture to assign to which set.
+	var current_texture_rd := texture_rds[(with_next_texture - 1) % 3]
+	var previous_texture_rd := texture_rds[(with_next_texture - 2) % 3]
+	var next_texture_rd := texture_rds[with_next_texture]
+
+	# Create our uniform sets so we can use these textures in our shader.
+	var current_set := _create_uniform_set(current_texture_rd, 0)
+	var previous_set := _create_uniform_set(previous_texture_rd, 1)
+	var next_set := _create_uniform_set(next_texture_rd, 2)
```
beicause, line 221: This seems to create uniforms every frame, no? I think we just need 9 uniforms and should cache them. Or use `UniformSetCacheRD`

Calinou: Feel free to open a PR for this :slightly_smiling_face: 

Same for the above comment.

Issue #1176, opened 2025-03-17
Update demos to use TileMapLayer instead of the deprecated TileMap


TheLoneInitiate:
Please update this to show how it is used with the new TileMapLayer

Calinou:
https://github.com/godotengine/godot-demo-projects/pull/1133 updated the Dynamic TileMap Layers demo, but many other demos remain to be updated.

Up-to-date projects:

```
~/Documents/Git/godotengine/godot-demo-projects master*
❯ rg 'type="TileMapLayer"'
networking/multiplayer_bomber/world.tscn
11:[node name="Layer0" type="TileMapLayer" parent="."]
2d/dynamic_tilemap_layers/world.tscn
24:[node name="Ground" type="TileMapLayer" parent="."]
29:[node name="Secret" type="TileMapLayer" parent="."]
2d/platformer/level/level.tscn
69:[node name="TileMapLayer" type="TileMapLayer" parent="."]
~/Documents/Git/godotengine/godot-demo-projects master*
```

Outdated projects:

```
❯ rg 'type="TileMap"'     
2d/navigation_astar/game.tscn
31:[node name="TileMap" type="TileMap" parent="."]
2d/kinematic_character/world.tscn
125:[node name="TileMap" type="TileMap" parent="."]
2d/skeleton/level.tscn
12:[node name="TileMap" type="TileMap" parent="."]
2d/role_playing_game/grid_movement/exploration.tscn
33:[node name="Ground" type="TileMap" parent="."]
40:[node name="Pathways" type="TileMap" parent="."]
627:[node name="Grid Walkable Floor" type="TileMap" parent="."]
633:[node name="Grid" type="TileMap" parent="." node_paths=PackedStringArray("dialogue_ui")]
2d/isometric/dungeon.tscn
424:[node name="Floor" type="TileMap" parent="."]
530:[node name="Walls" type="TileMap" parent="."]
795:[node name="HighWalls" type="TileMap" parent="."]
803:[node name="HighWalls2" type="TileMap" parent="."]
2d/hexagonal_map/map.tscn
8:[node name="TileMap" type="TileMap" parent="."]
2d/physics_platformer/stage.tscn
132:[node name="TileMap" type="TileMap" parent="."]
2d/physics_tests/tests/functional/test_character_tilemap.tscn
135:[node name="TileMap" type="TileMap" parent="."]
```

Calinou:
[...]

Issue #1114, opened 2024-09-15
Some character animations are not displayed in JRPG Demo


qezz:
**Which demo project is affected:**

JRPG Demo, `2d/role_playing_game`

**OS/device including version:**

Reproduced on
- NixOS Linux, `v4.3.stable.nixpkgs [77dcf97d8]`
- macOS Snoma 14.6, `v4.3.stable.official [77dcf97d8]`

**Issue description:**

The main character's `idle` and `walk` animations are not shown neither in the Editor, nor in the running game. The `bump` animation works both in the AnimationTree in the Editor, and in the running game.

Same issue with the PlayerOpponent, which to my understanding uses only `idle` state.

The animations work in https://godotengine.github.io/godot-demo-projects/2d/role_playing_game/, though I don't know how exactly those were built.

For the context, the original project is created with Godot 4.2, and migrates to 4.3 on import. Migration diff in the collapsed section below.

<details><summary> `$ git diff -- 2d/role_playing_game` </summary>

```diff
diff --git a/2d/role_playing_game/project.godot b/2d/role_playing_game/project.godot
index c33b19a1..7d0c52d6 100644
--- a/2d/role_playing_game/project.godot
+++ b/2d/role_playing_game/project.godot
@@ -16,7 +16,7 @@ and GDScript. It also includes a simple JRPG-style dialogue and
 battle system on top of it."
 config/tags=PackedStringArray("2d", "demo", "gui", "official", "tilemap")
 run/main_scene="res://game.tscn"
-config/features=PackedStringArray("4.2")
+config/features=PackedStringArray("4.3")
 config/icon="res://icon.svg"
 
 [debug]
@@ -38,32 +38,32 @@ import/blender/enabled=false
 
 move_up={
 "deadzone": 0.2,
-"events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":4194320,"key_label":0,"unicode":0,"echo":false,"script":null)
[...]

PR #1248, opened 2025-09-13
Get role_playing_game working again: Gamepad, TileMapLayer, fix bugs

idbrii:
The 2d rpg had a bunch of bugs and some awkwardness setup that wouldn't work well in a larger game.

Bug fixes:
* Player and opponent are now visible. (#1114)
* Fade to black works between explore and combat.

Improvements:
* Now playable with gamepad.
* Uses TileMapLayer (#1176).
* Uses SpriteFrames to allow different art "builds" per character instead of changing textures in AnimationPlayer tracks.
* Separate Walker and Player classes so more than one thing can walk.
  * I don't love this inheritance tree, but opponents share player's visuals but didn't share code to configure those visuals. This seemed like a reasonable split.

I made these changes using Godot v4.4.1.stable.official [49a5bc7b6].

# Screenshots

image

https://github.com/user-attachments/assets/21b6850e-0096-4b41-b058-ed3de5eb6d05


Calinou (commented):
Tested locally on 4.5.rc2, it mostly works as expected.

I noticed one regression:

- UI elements (buttons) can't be interacted with the mouse anymore, in both the main and fighting scenes. Keyboard interactions still work. I tested this in both 4.4.1.stable and 4.5.rc2.

Another issue that's already in `master`, but now would be a good opportunity to fix it:

- Character animations don't loop in the fighting scene. They will only play once after each action:

https://github.com/user-attachments/assets/b6c2f9e0-6114-4d21-86c6-e1eda7275754

idbrii:
Fixed:

* buttons are now clickable with the mouse anymore, in both the main and fighting scenes.
* Character animations now loop in the fighting scene.

I debated whether the combat characters should use the same visual setup as the exploration characters, but decided that most rpgs have very different actions and appearances between the modes and didn't feel motivated to do that refactoring so I made a more minimal change.

https://github.com/user-attachments/assets/1d5c186a-ffc2-44d5-9423-106e262ab95b

I also removed this comment since I could not get the error to fire:

	# FIXME: Error calling from signal 'combat_finished' to callable:
	# 'Node(game.gd)::_on_combat_finished': Cannot convert argument 1 from Object to Object.

PR #1253, opened 2025-10-02
Make Voxel demo player responsive without using unrealistic global gravity

aaronfranke:
Follow-up to PR #1224.

This PR makes the gravity acting on the player stronger when the player has an upward velocity or a slow downward velocity, allowing for the jumping to feel nicer without increasing the global gravity.

This PR also implements basic sprinting, uses the new `get_gravity()` function, and updates the demo to Godot 4.5.


3d/voxel/player/player.gd
```diff
@@ -95,10 +96,17 @@ func _physics_process(delta: float) -> void:
 
 	if crouching:
 		movement *= MOVEMENT_SPEED_CROUCH_MODIFIER
+		sprinting = false
+	if sprinting:
+		movement *= MOVEMENT_SPEED_SPRINT_MODIFIER
+		camera.fov = lerpf(camera.fov, 100.0, delta * 10)
+	else:
+		camera.fov = lerpf(camera.fov, 80.0, delta * 10)
```
Calinou: Use FPS-independent lerp (in case physics ticks is changed):

```suggestion
	if crouching:
		movement *= MOVEMENT_SPEED_CROUCH_MODIFIER
		sprinting = false
	if sprinting:
		movement *= MOVEMENT_SPEED_SPRINT_MODIFIER
		camera.fov = lerpf(camera.fov, 100.0, 1.0 - exp(-delta * 10.0))
	else:
		camera.fov = lerpf(camera.fov, 80.0, 1.0 - exp(-delta * 10.0))
```

Note that the multiplier may have to be changed to get a similar result to before.

Also, this FOV seems on the high side, considering we were using a FOV of 74° when not sprinting before (Godot's default is 75°).

aaronfranke: The higher FOV when sprinting made me realize that the non-sprinting FOV felt a tad low, but it doesn't really matter, also I'm biased.

PR #1257, opened 2025-10-02
Switch back to Jolt Physics in Voxel demo

mihe:
As mentioned in #1252, mutating compound shapes is quite slow when using Jolt Physics, due to design choices I made at one point. However, this only applies so long as the body in question is actually in the scene tree. If you remove the body from the scene tree you only pay for the expensive rebuild once the body enters the scene tree again.

So this PR does exactly that, by first moving the `Chunk._ready` logic to `Chunk._init` instead, which means we're no longer generating the initial colliders _after_ we've entered the scene tree, which I would argue is probably a better fit anyway. It also changes `Chunk.regenerate` to temporarily remove the body from the scene tree, with a comment explaining why.

On my machine (9950X3D) this mostly resolves the performance discrepancy between the two physics engines, although I feel like there's still a tiny stutter when breaking blocks with Jolt, which might warrant taking a closer look at.

I also threw in some other minor changes:

- Added more static typing (mostly for ease of jumping through the code).
- Removed a seemingly redundant (outdated?) hack for empty chunks.
- Added sharing of the underlying `BoxShape3D` between all chunk colliders.
- Replaced `Chunk`'s use of `Thread` with the more appropriate `WorkerThreadPool`.


aaronfranke (approved):
Makes sense, looks good, runs well. Thanks!

Issue #1269, opened 2025-10-30
Compute: a nasty memory leak on Metal in current 4.5.1


iamvoidcoder:
### Tested versions

- Reproducible in: 4.5.1.stable

### System information

Godot v4.5.1.stable - macOS Sequoia (15.7.1) - Multi-window, 3 monitors - Metal (Forward+) - integrated Apple M2 Ultra (Apple8) - Apple M2 Ultra (24 threads) - 64.00 GiB memory

### Issue description

There appears to be a rather nasty memory leak in the current top-of-the-tree build 4.5.1.stable. At least on MacOS / Metal, not sure about other platforms. I was messing with some compute stuff and discovered a steady mystery RAM leak of about ~0.25MB/second or ~1MB every 4 seconds. 

I've spent some good time digging the project to make sure I'm not leaking anything like orphan node instances or dynamic resources etc. But it is happening even on a static scene with zero activity other than compute task being dispatched every frame on the render thread. 

Eventually I stumbled on some official Godot demo/test projects on Github and tried the Compute Texture one from [>>> here <<<](https://github.com/godotengine/godot-demo-projects/tree/master/compute/texture).

Without any modifications to stock Compute Texture demo project this is what I get after bout ~5 minutes of run:

<img width="1219" height="288" alt="Image" src="https://github.com/user-attachments/assets/cdf2c75d-c3d0-4831-a909-3f97835efdd5" />

### Steps to reproduce

Open Godot [Compute Texture](https://github.com/godotengine/godot-demo-projects/tree/master/compute/texture) demo project on MacOS and run. Attaching it blow as there seems to be no way to submit the bug without attaching a repro project.

### Minimal reproduction project (MRP)

[texture.zip](https://github.com/user-attachments/files/23238890/texture.zip)

clayjohn:
What changes did you make to the project? Just a quick look and I can see it is leaking 3 uniform sets per frame here:

https://github.com/godotengine/godot-demo-projects/blob/6adcbc912e20ac163d9e7d37446ba189cb4a81b6/compute/texture/water_plane/water_plane.gd#L152

Looking around it seems like the leak has been known about for some time too https://github.com/godotengine/godot-demo-projects/pull/1247/files#r2347714761

[...]

PR #1275, opened 2025-11-06
Fix memory leak in "Compute Texture"

1hue:
Fix a memory leak resulting from improper uniform set handling introduced in #1247.

Follow the recommendation in https://github.com/godotengine/godot-demo-projects/issues/1269#issuecomment-3478434095.

Compute Texture memory profiling

Also address warning:
```
The local function parameter "input_event" is shadowing an already-declared signal in the base class "CollisionObject3D".
```

---

Tested in:
- [x] v4.5.1.stable
- [x] v4.4.1.stable

---

Closes:
- #1269

### DIFF
--- a/compute/texture/water_plane/water_plane.gd
+++ b/compute/texture/water_plane/water_plane.gd
@@ -219,7 +219,7 @@ func _render_process(with_next_texture: int, wave_point: Vector4, tex_size: Vect
 	@warning_ignore("integer_division")
 	var y_groups := (tex_size.y - 1) / 8 + 1
 
-	# Figure out which uniforms to assign.
+	# Figure out which texture to assign to which set.
 	var current_set := texture_sets[with_next_texture * 3]
 	var previous_set := texture_sets[with_next_texture * 3 + 1]
 	var next_set := texture_sets[with_next_texture * 3 + 2]


compute/texture/water_plane/water_plane.gd
```diff
@@ -242,5 +249,9 @@ func _free_compute_resources() -> void:
 		if texture_rds[i]:
 			rd.free_rid(texture_rds[i])
 
+	for i in 9:
+		if texture_sets[i] and texture_sets[i].is_valid():
```
1hue, line 253: Not sure if we actually need to do this, as sets gets cleaned up alongside the shader? But maybe good for clarity?

1hue:
Is this ok to merge? @Calinou

compute/texture/water_plane/water_plane.gd
```diff
@@ -55,16 +55,16 @@ func _exit_tree() -> void:
 	RenderingServer.call_on_render_thread(_free_compute_resources)
 
 
-func _unhandled_input(input_event: InputEvent) -> void:
+func _unhandled_input(event: InputEvent) -> void:
```
AThousandShips:
```suggestion
func _unhandled_input(input_event: InputEvent) -> void:
```
This reverts a previous PR, please restore

AThousandShips:
```suggestion
func _unhandled_input(input_event: InputEvent) -> void:
```
This reverts a previous PR, please restore

1hue: You're right, this shouldn't have been renamed

PR #1279, opened 2025-11-14
Improve GUI in 3D demo project

rpaciorek:
- fixed GUI input issue in billboard mode (by use custom billboard calculation also for mesh node)
- added settings for gui_3d script gui_panel_3d.tscn allows configure gui in 3d without modifying this scene (and easily reuse it)

I'm not sure if this approach (with configuration via setters in `gui_3d.gd`) isn't too complicated for a demo ... but:
1. managing the billboard at the script level (and not the material level) allows solving the input synchronization problem by applying rotation to the entire node (and not just collisions) ... this shouldn't cause a significant overhead because these calculations were performed and applied by this script anyway ...
2. separates the configuration from the logic of the 3D GUI object, allowing you to easily reuse this solution in your project

### DIFF
--- a/viewport/gui_in_3d/gui_3d.gd
+++ b/viewport/gui_in_3d/gui_3d.gd
@@ -1,8 +1,52 @@
 extends Node3D
 
+@export var gui_scene : PackedScene :
+	set(new_scene):
+		gui_scene = new_scene
+		if node_viewport:
+			if _gui_scene_instance:
+				node_viewport.remove_child(_gui_scene_instance)
+				_gui_scene_instance.queue_free()
+			_gui_scene_instance = gui_scene.instantiate()
+			node_viewport.add_child(_gui_scene_instance)
+	get():
+		return gui_scene
+
+
+@export var screen_size_3d := Vector2(1, 1) :
+	set(new_size):
+		screen_size_3d = new_size
+		if node_collision:
+			node_quad.mesh.size = screen_size_3d
+			node_collision.shape.size.x = screen_size_3d.x
+			node_collision.shape.size.y = screen_size_3d.y
+	get():
+		return screen_size_3d
+
+
+@export var screen_size_2d := Vector2(512, 512) :
+	set(new_size):
+		screen_size_2d = new_size
+		if node_viewport:
+			node_viewport.size = screen_size_2d
+	get():
+		return screen_size_2d
+
+
+@export var billboard_mode : BaseMaterial3D.BillboardMode :
+	set(new_mode):
+		billboard_mode = new_mode
+		if billboard_mode != BaseMaterial3D.BillboardMode.BILLBOARD_DISABLED:
+			# run billboard specific code only if billboard mode is set
+			set_process(true)
+		else:
+			set_process(false)
+	get():
+		return billboard_mode
+
 
 ## Used for checking if the mouse is inside the Area3D.
-var is_mouse_inside: bool = false
+var is_mouse_inside := false
 
 ## The last processed input touch/mouse event. Used to calculate relative movement.
 var last_event_pos2D := Vector2()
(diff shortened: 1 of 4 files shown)


viewport/gui_in_3d/gui_3d.gd
```diff
@@ -118,21 +165,20 @@ func _mouse_input_event(_camera: Camera3D, input_event: InputEvent, event_positi
 
 
 func rotate_area_to_billboard() -> void:
-	var billboard_mode: BaseMaterial3D.BillboardMode = node_quad.get_surface_override_material(0).billboard_mode
+		# custom billboard implementation - needed for rotate physics (collision) node,
+		# but also apply to mesh node (instead of using billboard material) to keep it in sync with collision shape
 
-	# Try to match the area with the material's billboard setting, if enabled.
-	if billboard_mode > 0:
```
AThousandShips: All of the code below this needs to be fixed for indentation

viewport/gui_in_3d/gui_3d.gd
```diff
@@ -1,29 +1,78 @@
...
+@export var billboard_mode : BaseMaterial3D.BillboardMode :
+	set(new_mode):
+		billboard_mode = new_mode
+		if billboard_mode != BaseMaterial3D.BillboardMode.BILLBOARD_DISABLED:
+			# run billboard specific code only if billboard mode is set
+			set_process(true)
+		else:
+			set_process(false)
+	get():
+		return billboard_mode
```
zodiepupper, line 45: the setter and getter here are really not necessary. `set_process(enabled: bool)` disables the process loop on this node and has nothing to do with billboarding. 

i'll add another suggestion below for the process method

```suggestion

## this is used to hold the rotation just before we started billboarding 
## that way we can revert the rotation changes if the user changes this 
## setting at runtime
var _billboard_revert_rotation : Vector3 

## sets the billboard mode used in the manual billboarding logic in 
## rotate_area_to_billboard
@export var billboard_mode : BaseMaterial3D.BillboardMode:
	set(val):
		# if the new mode disables billboarding, then revert the rotation to what it was before
		# we started billboarding
		if val == BaseMaterial3D.BillboardMode.BILLBOARD_DISABLED:
			global_rotation = _billboard_revert_rotation
		# if we are initiating billboarding (we only update this when coming from 
		# disabled to prevent updating the revert rotation while billboarding is applied)
		elif billboard_mode == BaseMaterial3D.BillboardMode.BILLBOARD_DISABLED:
			# if it was not disabled, then update the revert rotation
			_billboard_revert_rotation = global_rotation
		billboard_mode = val
```

viewport/gui_in_3d/gui_3d.gd
zodiepupper, line 79: you should put the check from the setter here instead to prevent confusing behavior for someone copying this for use elsewhere


```suggestion
func _process(_delta: float) -> void:
	# NOTE: Remove this function if you don't plan on using billboard settings.
	# if billboard is enabled, then call a method to rotate appropriately
	if billboard_mode != BaseMaterial3D.BillboardMode.BILLBOARD_DISABLED:
		rotate_area_to_billboard()
```

viewport/gui_in_3d/gui_3d.gd
```diff
@@ -118,21 +165,20 @@ func _mouse_input_event(_camera: Camera3D, input_event: InputEvent, event_positi
...
 		# Y-Billboard: Lock Y rotation, but gives bad results if the camera is tilted.
 		if billboard_mode == 2:
 			look = Vector3(look.x, 0, look.z)
 
-		node_area.look_at(look, Vector3.UP)
+		node_quad.look_at(look, Vector3.UP)
 
 		# Rotate in the Z axis to compensate camera tilt.
-		node_area.rotate_object_local(Vector3.BACK, camera.rotation.z)
+		node_quad.rotate_object_local(Vector3.BACK, camera.rotation.z)
```
zodiepupper, line 184: this code doesn't do what you were attempting, i wrote a version that implements the billboarding modes properly the way you wanted with detailed comments

```suggestion
## rotates the UI panel manually to replicate billboarding while retaining 
## physics interactions on the final visual
func rotate_area_to_billboard() -> void:
		# custom billboard implementation - needed for rotate physics (collision) node,
		# but also apply to mesh node (instead of using billboard marevert_rotationterial) to keep it in sync with collision shape

		# Get the camera.
		var camera := get_viewport().get_camera_3d()
		
		match billboard_mode:
			# if billboard is set to enabled, then we replicate that behavior
			# with a simple look_at (default up is used and use_model_front
			# is required because of the orientation of the ui panel)
			BaseMaterial3D.BillboardMode.BILLBOARD_ENABLED:
				look_at(camera.global_position, Vector3.UP, true)
			# if billboard is set to fixed y, then we replicate that lazily
			# by just using a holder for what the rotation was before applying
			# the billboad and then resetting the global y rotation.
			BaseMaterial3D.BillboardMode.BILLBOARD_FIXED_Y:
				# grab the rotation before applying
				var revert_rotation : Vector3 = global_rotation
				look_at(camera.global_position, Vector3.UP, true)
				# revert the appropriate axis to replicate the fixed y behavior
				global_rotation.x = revert_rotation.x
				#global_rotation.y = revert_rotation.y
				global_rotation.z = revert_rotation.z
			BaseMaterial3D.BillboardMode.BILLBOARD_PARTICLES:
				pass # this cannot be implemented in this way
```

viewport/gui_in_3d/gui_3d.gd
zodiepupper: i updated it just now with a setter method that enables reverting the rotation to what it was before the billboard mode was enabled

PR #1280, opened 2025-11-16
Multiple fixes and improvements to Mono projects

asyncbtd:
- The mono/2.5d project has had its scripts corrected to work in godot 4.5.0
- In the mono/dodge_the_creeps project, the StartGame() and Hit() signal listeners have been added.
- In the mono/multiplayer_pong project, the collision area of ​​the paddle object was fixed.
- In the mono/pong project, the collision area of ​​the paddle, wall, and ball objects has been fixed.
- In the mono/squash_the_creeps project, added a listener for the OnPlayerHit signal. Added audio pause when a player loses a game

Is it compatible with engine version 3x? I don't know, no testing has been done.

**Don't treat this as a pull request**

In the future, I plan to maintain and create new demos in C#. Do I need to follow Microsoft's formatting guidelines? Or can I format the code however I want?

### DIFF
--- a/mono/squash_the_creeps/Main.cs
+++ b/mono/squash_the_creeps/Main.cs
@@ -15,6 +15,7 @@ public partial class Main : Node
         if (inputEvent.IsActionPressed("ui_accept") && GetNode<Control>("UserInterface/Retry").Visible)
         {
             // This restarts the current scene.
+            GetNode<AudioStreamPlayer>("/root/MusicPlayer").Play();
             GetTree().ReloadCurrentScene();
         }
     }
@@ -46,5 +47,6 @@ public partial class Main : Node
     {
         GetNode<Timer>("MobTimer").Stop();
         GetNode<Control>("UserInterface/Retry").Show();
+        GetNode<AudioStreamPlayer>("/root/MusicPlayer").Stop();
     }
 }
--- a/mono/squash_the_creeps/Main.tscn
+++ b/mono/squash_the_creeps/Main.tscn
@@ -151,4 +151,5 @@ grow_horizontal = 2
 grow_vertical = 2
 text = "Press Enter to retry"
 
+[connection signal="Hit" from="Player" to="." method="OnPlayerHit"]
 [connection signal="timeout" from="MobTimer" to="." method="OnMobTimerTimeout"]


mono/2.5d/addons/node25d-cs/Basis25D.cs
```diff
@@ -106,10 +100,10 @@ public Vector2 this[int columnIndex]
     /// Try to keep this number away from a multiple of Tau/4 (or Pi/2) radians.
     /// </summary>
     /// <param name="angle">The angle, in radians, between the Y axis and the X/Z axes.</param>
-    public static Basis25D Dimetric(real_t angle)
+    public static Basis25D Dimetric(float angle)
     {
-        real_t sin = Mathf.Sin(angle);
-        real_t cos = Mathf.Cos(angle);
+        float sin = Mathf.Sin(angle);
+        float cos = Mathf.Cos(angle);
```
aaronfranke, line 106: `real_t` is intended to provide support for compiling Godot with double-precision floats with the same codebase, so it should not be replaced with `float` which would break that ability.

aaronfranke:
> Is it compatible with engine version 3x? I don't know, no testing has been done.

It is not, but also, it cannot, and it does not need to be. Demos for Godot 3.x are maintained [on a separate branch](https://github.com/godotengine/godot-demo-projects/tree/3.x).

> Do I need to follow Microsoft's formatting guidelines? Or can I format the code however I want?

Follow Godot's guidelines: https://docs.godotengine.org/en/latest/tutorials/scripting/c_sharp/c_sharp_style_guide.html

mono/2.5d/addons/node25d-cs/Basis25D.cs
asyncbtd: I would have kept ``real_t`` if it worked out of the box. The problem is that in the official version of Godot, the ``_Process`` method accepts delta as a ``double``.

Also, this example is for beginners, and I believe that if a beginner compiles Godot themselves with double precision, they can change the float themselves.

I could fix the current code to use ``real_t`` everywhere, but is that really necessary? Other C# examples don't use ``real_t`` at all.

aaronfranke: Then change `delta` and leave the rest as-is?

asyncbtd: It's your repository, you decide. Should I fix it?

PR #1296, opened 2026-02-11
Soft body demo: Lowered impulses because of lowered default mass

jrouwe:
Since #116041, the default mass of a soft body is 1 kg for Jolt Physics (just like Godot Physics). The demos were tweaked with the previous default mass, which is 1 kg per vertex for a total mass of >4000 kg. This change reduces the applied impulses so that the demos behave sensibly again.


aaronfranke (approved):
Approved. This will be good to merge once the demos are targeting Godot 4.7, but that time is not quite here yet.

jrouwe:
@Calinou I've fixed the merge conflict. Could this be integrated? The soft body demo looks quite broken without it on 4.7.

jrouwe:
Thanks!

PR #1299, opened 2026-02-19
Add a custom splash screen demo

### LINKED ISSUE #14166: Option to use a scene as the boot splash screen
### Describe the project you are working on

I am working on a web game. The boot phase is more pronounced especially on web builds, making the initial splash screen a meaningful part of the user experience.

### Describe the problem or limitation you are having in your project

Godot’s current boot splash screen is very limited in customization. I can only change background color and set a static image, which causes several issues:
- The boot splash does not match the game’s visual style.
- No way to show tips and status text.
- No animations.

### Describe the feature / enhancement and how it helps to overcome the problem or limitation

**Option A: Custom scene as boot splash**
Allow developers to define a minimal Godot scene that is rendered during the boot phase. This scene would be subject to constraints (no heavy scripting, limited node types), but would enable:
- Custom layout and branding
- Animated elements
- Loading indicators tied to engine loading progress
- Text elements (tips, status text)

**Option B: Expanded configuration for existing boot splash**
If a custom scene is not feasible, extend the current boot splash with configurable options:
- Customizable loading bar: Position, size, color, style
- Optional text fields
- Optional spritesheet instead of image for simple animation

### Describe how your proposal will work, with code, pseudo-code, mock-ups, and/or diagrams

**Option A – Custom scene as boot splash**
```
Project Settings > Application > Boot Splash
  Splash Mode: Scene or Static Image
  Splash Scene: path/to/scene.tscn
```
For the splash scene:
- Limited node set (Label, TextureRect, ColorRect, ProgressBar. TextureProgressBar)
- No external signals or autoload access

**Option B – Extended configuration**
In addition to current settings, we can add:
```
Project Settings > Application > Boot Splash
  Spritesheet: set a spritesheet to animate
  Show Loading Bar: bool
  Loading Bar Color: color picker
  Loading Text: String
  Text Alignment: Top Left, Top Center, Top Right, Middle Left, Center, Middle Right, Bottom Left, Bottom Center or Bottom Right
```

### If this enhancement will not be used often, can it be worked around with a few lines of script?

No. The boot splash occurs before user scripts and scenes are initialized. Any workaround requires showing a fake loading screen after the splash boot screen.

### Is there a reason why this should be core and not an add-on in the asset library?

I believe everyone will benefit if it's in core system.

Calinou:
This effectively reimplements a demo we had in the Godot 2 days, albeit with different visuals.

I'd like to add a video playback example as well, but since a project can only have one splash screen, I guess it would have to be a separate (non-main) scene that acts as a second splash screen.

- This closes https://github.com/godotengine/godot-proposals/issues/14166.

## Preview

[custom_splash_screen.webm](https://github.com/user-attachments/assets/a15a691a-8dd7-4837-8288-66963d336ac2)


aaronfranke (approved):

BastiaanOlij:
Does this do any sort of background loading of the main scene? It would sort of make more sense to have the main scene loading while the animation plays. That maybe is an entirely different demo but this is especially important in XR games where we want to go to a stereoscopic splash/loading screen as soon as possible, and where any hiccup in framerate is immediately noticeable.

Even with my own current implementation of using the resource loader to do background loading of the main scene, I still get that hiccup due to shader compilation, so a good demo that both shows the correct implementation, and allows us to showcase the issue to solve, would kill two birds with one stone :)

(note here that my hiccup is likely solveable with some of the additions we did in 4.6, still on my list of things to test, but same goes, having an demo project that shows the correct setup to loading your game hiccup free, would be golden)

Calinou:
> Does this do any sort of background loading of the main scene?

Not yet. This could be added, but I'm wary of this kind of background loading introducing stutter while the animation is loading. It would also make the demo much more complex, especially if we implement a precompiling solution for shaders that aren't "caught" by the ubershader system (or when using Compatibility).

I've rarely seen background loading solutions be completely smooth in terms of frametime, which leads to an unpleasant experience if they are loading stuff while a video is playing (on top of [VRR flicker](https://www.rtings.com/monitor/learn/research/vrr-flicker) when playing on a VRR display).

Ashhuby:
Great upgrade imo

zodiepupper:
i think it's pretty great! 

i think it would be slightly improved by also adding a fade in on the main scene so it's not so sudden.

[custom_splash_screen.webm](https://github.com/user-attachments/assets/3c6f0fc5-9955-4171-970e-f472fafb42e6)

Calinou:
Rebased and tested again on 4.7.stable, it works as expected.

> i think it would be slightly improved by also adding a fade in on the main scene so it's not so sudden.

I added a fade-in transition to the main scene.

zodiepupper:
i just pulled and tested it again. looks perfect to me!

PR #1300, opened 2026-02-23
Update and fix Glow for 2D demo for Godot 4.6

Calinou:
- Switch to Mobile renderer for consistent visuals across platforms.
- Enable HDR 2D and debanding for better visuals.
- Tweak glow parameters to make the effect more visible.
- Use AgX tonemapping.
- Use a CanvasLayer to show the help label, so that it is not affected by glow. Give it a outline to be readable on any background.

### DIFF
--- a/2d/glow/README.md
+++ b/2d/glow/README.md
@@ -2,11 +2,14 @@
 
 This showcases how to use glow in a 2D game via the WorldEnvironment node.
 
-Slide the cave image left and right to observe the glow effect at work.
+Slide the cave image left and right to observe the glow effect at work. Press <kbd>G</kbd>
+to toggle the glow map (lens dirt) effect.
+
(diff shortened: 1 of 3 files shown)


allenwp:
Just from looking at the Files Changed, I see that `tonemap_agx_contrast` might be set to `1.0` -- This reduces contrast compared to no tonemapping... is this intentional to push this parameter to its extreme?

Calinou:
> Just from looking at the Files Changed, I see that `tonemap_agx_contrast` might be set to `1.0` -- This reduces contrast compared to no tonemapping... is this intentional to push this parameter to its extreme?

I tried again with contrast `1.25`, but I see a lot of black crush occurring on the lefthand side of the demo. It's probably why I reduced contrast to `1.0` originally.

aaronfranke (approved):
Looks good to me.

PR #1307, opened 2026-03-26
Fix missing signal connections in C# demos

raulsntos:
Looks like these signal connections were accidentally removed in https://github.com/godotengine/godot-demo-projects/pull/1258. It likely happened because the projects were saved without compiling the C# code first, and Godot removed the connections to _unknown_ signals.


Calinou (approved):
Looks good to me.

aaronfranke (approved):

PR #1309, opened 2026-03-28
Fix indentation in note_manager.gd causing Rhythm Game to fail at run.

JordanChwalik:
Godot 4.6.1

In the current Rhythm Game Demo, the script note_manager.gd contained an indentation error that prevented the demo project from running. This fork fixes that indentation problem and allows the NoteManager class to be referenced in other files.


Calinou (approved):
Tested locally, it works as expected. Thanks!

PR #1310, opened 2026-04-01
Support the Mobile renderer in the Decals demo

Calinou:
The default renderer is still Forward+, but this allows the demo to display correctly when using the Mobile renderer by splitting each decal receiver into its own MeshInstance3D node. This ensures each mesh does not display more than 8 decals at once, which is the limit in the Mobile renderer.

This paves the way for supporting the demo with the Compatibility renderer as well once https://github.com/godotengine/godot/pull/118070 is merged, therefore making the demo friendly to web exports.


aaronfranke (approved):

BastiaanOlij:
This works well for me, the only sample that failed on mobile was the distance fade, but that might be a bug in the decal code we need to look at (or a limitation).

Calinou:
> This works well for me, the only sample that failed on mobile was the distance fade, but that might be a bug in the decal code we need to look at (or a limitation).

The leftmost decal on the distance fade example is meant to suddenly disappear when you get far away enough from it. This is why it also occurs on Forward+. Maybe we should change it to have a small fade transition instead, so it looks less like a bug.

PR #1311, opened 2026-04-08
Update vehicle.gd in 3d/truck_town

FlashAndromeda:
Fixed the double acceleration bug when reversing. Breaking needs to happen through reducing speed to 0 instead of applying reverse acceleration.


Calinou (approved):

I can confirm this fixes the issue where going in reverse was faster than going forwards.

PR #1316, opened 2026-05-01
Add custom_resource plugin demo

### LINKED ISSUE #1061: Update plugins demo to use custom Resources

adamcodeworks:
This adds a new `custom_resource` plugin to the existing plugins demo project, mirroring the `custom_node` plugin but for Resources. It uses `add_custom_type` to register a small `Stats` Resource subclass (with `max_health`, `strength`, and `speed` @export properties) so that the type appears in the "New Resource…" dialog and as an option for `@export var foo: Stats` properties in the inspector.

Why additive rather than replacing `custom_node`: the heart-drawing demo is genuinely a Node use case (it draws via `_draw()` and lives in the scene tree), and the two demos now sit side by side as the canonical pair for `add_custom_type`. Happy to consolidate if you'd prefer a different shape — let me know.

Closes #1061

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -16,10 +16,13 @@ Check out this demo on the asset library: https://godotengine.org/asset-library/
 
 # How does it work?
 
-This project contains 4 plugins:
+This project contains 5 plugins:
 
 * The custom node plugin shows how to create a custom node type
   using `add_custom_type`. [More info](addons/custom_node).
+
+* The custom resource plugin shows how to create a custom Resource type
+  using `add_custom_type`. [More info](addons/custom_resource).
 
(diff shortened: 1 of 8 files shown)


AThousandShips:

AThousandShips:

PR #1319, opened 2026-05-02
Interpolate brush samples in gd_paint to fill stroke gaps

### LINKED ISSUE #846: Spaces between circles in GDPaint
Hi. Why in your demo I have spaces between circles? How can i fix it?

https://user-images.githubusercontent.com/5132385/219780927-0dbb08e1-91b8-44ab-b30e-0ef6fa254265.mov

adamcodeworks:
Fixes #846.

## Problem

The GDPaint demo adds one brush sample per `_process` call (one per frame). When the mouse moves faster than one brush per frame the stroke shows visible gaps between brushes — most obvious with the default 32-pixel brush and a quick swipe of the mouse.

## Fix

When a stroke continues from one frame to the next, interpolate brush samples along the line from the previous mouse position to the current one, stepping roughly one pixel at a time. This is the same idea as Bresenham's line algorithm (as suggested by @Calinou in the issue), expressed parametrically because brush positions are stored as floats. A new helper `add_brush_line()` centralizes the loop.

Two corner cases keep the existing behavior intact:

1. The first sample of a stroke still places a single brush at the mouse position — there is no previous in-stroke sample to interpolate from.
2. If the previous sample's position is outside the canvas (the user briefly left and re-entered the drawing area mid-stroke), we also place a single brush instead of drawing a line through the off-canvas area. The `PaintControl` node does not clip its drawing, so without this guard a re-entry could draw brushes on top of the tools panel.

Undo continues to work unchanged: `undo_element_list_num` is captured before any sample is added in a stroke, and `undo_stroke()` already pops every sample added since then, regardless of how many were appended per frame.

## Testing

I haven't run this in the editor yet, so a quick visual check before merging would be appreciated. Algorithmically: gapless strokes for any mouse speed, undo semantics preserved, no drawing outside the canvas.

### DIFF
--- a/gui/gd_paint/paint_control.gd
+++ b/gui/gd_paint/paint_control.gd
@@ -56,7 +56,8 @@ func _process(_delta: float) -> void:
 		# If we do not have a position for when the mouse was first clicked, then this must
 		# be the first time is_mouse_button_pressed has been called since the mouse button was
 		# released, so we need to store the position.
-		if mouse_click_start_pos.is_equal_approx(Vector2.INF):
+		var stroke_just_started := mouse_click_start_pos.is_equal_approx(Vector2.INF)
+		if stroke_just_started:
 			mouse_click_start_pos = mouse_pos
 
 		# If the mouse is inside the canvas and the mouse is 1px away from the position of the mouse last _process call.
@@ -69,8 +70,16 @@ func _process(_delta: float) -> void:
 					if undo_set == false:
 						undo_set = true
 						undo_element_list_num = brush_data_list.size()
-					# Add the brush object to draw_elements_array.
-					add_brush(mouse_pos, brush_mode)
+					# On the first sample of a new stroke we just place a single brush at the mouse
+					# position. On subsequent samples we interpolate from the previous mouse
+					# position so that fast mouse motion does not leave gaps between brushes.
+					# When last_mouse_pos was outside the canvas (e.g. the cursor briefly left
+					# the drawing area mid-stroke) we also fall back to a single brush so we
+					# don't draw a line through the area outside the canvas.
+					if stroke_just_started or not drawing_area_rect.has_point(last_mouse_pos):
+						add_brush(mouse_pos, brush_mode)
+					else:
+						add_brush_line(last_mouse_pos, mouse_pos, brush_mode)
 
 	else:
 		# We've finished our stroke, so we can set a new undo (if a new stroke is made).
@@ -186,6 +195,19 @@ func add_brush(mouse_pos: Vector2, type: BrushMode) -> void:
 	# Add the brush and update/draw all of the brushes.
 	brush_data_list.append(new_brush)
 	queue_redraw()
+
+
+# Adds a series of brush samples along the line from `from` to `to` so that fast
+# mouse motion does not leave visible gaps between brushes. We step roughly once
+# per pixel along the line; this is the same idea as Bresenham's line algorithm,
+# expressed parametrically because brush positions are stored as floats.
+func add_brush_line(from: Vector2, to: Vector2, type: BrushMode) -> void:
+	var steps := int(ceil(from.distance_to(to)))
+	if steps <= 1:
+		add_brush(to, type)
+		return
+	for i in range(1, steps + 1):
+		add_brush(from.lerp(to, float(i) / float(steps)), type)
 
 
 func _draw() -> void:
--- a/gui/gd_paint/paint_control.gd
+++ b/gui/gd_paint/paint_control.gd
@@ -197,15 +197,18 @@ func add_brush(mouse_pos: Vector2, type: BrushMode) -> void:
 	queue_redraw()
 
 
-# Adds a series of brush samples along the line from `from` to `to` so that fast
-# mouse motion does not leave visible gaps between brushes. We step roughly once
-# per pixel along the line; this is the same idea as Bresenham's line algorithm,
-# expressed parametrically because brush positions are stored as floats.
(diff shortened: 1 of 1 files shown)


emmyjemmyjammy:
Hi! I've been trying to play around with this demo and was getting frustrated because I was unsure how to add the interpolation needed to close the gaps between brush instances so I was so excited when I googled my issue and saw this pull request! I immediately went to test the code and just wanted to let you know that it seems to have some issues relating to performance where it starts out smooth but as the amount of brush objects grows it becomes jagged

screenshot-2026-05-12-20-31-03

screenshot-2026-05-12-20-31-53

I also got this error when trying to clear the picture or undo strokes at a high brush object count:
E 0:01:35:164   button_pressed: Invalid access to property or key 'BrushModes' on a base object of type 'Control (paint_control.gd)'.
  <GDScript Source>tools_panel.gd:50 @ button_pressed()
  <Stack Trace> tools_panel.gd:50 @ button_pressed()

From my research on how digital painting programs work, I'm not sure if this might be an issue related to needing to combine all the objects into a single one when the stroke is completed.

adamcodeworks:
Thanks for actually trying this out @emmyjemmyjammy — that's super helpful.

You're right that the performance gets worse as the brush list grows. The demo's `_draw` walks the entire `brush_data_list` on every redraw, so stroke cost is roughly O(N²) for a stroke of N samples. My first pass stepped once per pixel, which made the list grow about four times faster than the previous behaviour and made the stutter hit much sooner.

Just pushed `cc07fc68`: step at `max(1, brush_size / 4)` instead of every pixel. With the default 32px brush that's one sample every 8px — still ~75% overlap between adjacent brushes, so the gaps stay closed, but the list grows roughly a quarter as fast as before. This matches the trade-off Calinou suggested back when the issue was filed ("every pixel … or perhaps every 2 pixels to improve performance"). It won't make the underlying O(N²) demo architecture fast, but the stutter should kick in much later.

On the `'BrushModes'` error: that one's confusing me, because the script uses `BrushMode` (singular) everywhere — the only `s` near there is on `brush_data_list`. Could you double-check that `tools_panel.gd:50` line on your end, in case something got hand-edited locally? A copy of the surrounding 3-4 lines from your file would help me track it down. The line number doesn't match `paint_control.BrushMode.ERASER` in the current source either, so I suspect a local edit or a different branch.

AThousandShips:

emmyjemmyjammy:
> Thanks for actually trying this out @emmyjemmyjammy — that's super helpful.
> 
> You're right that the performance gets worse as the brush list grows. The demo's `_draw` walks the entire `brush_data_list` on every redraw, so stroke cost is roughly O(N²) for a stroke of N samples. My first pass stepped once per pixel, which made the list grow about four times faster than the previous behaviour and made the stutter hit much sooner.
> 
> Just pushed `cc07fc68`: step at `max(1, brush_size / 4)` instead of every pixel. With the default 32px brush that's one sample every 8px — still ~75% overlap between adjacent brushes, so the gaps stay closed, but the list grows roughly a quarter as fast as before. This matches the trade-off Calinou suggested back when the issue was filed ("every pixel … or perhaps every 2 pixels to improve performance"). It won't make the underlying O(N²) demo architecture fast, but the stutter should kick in much later.
> 
> On the `'BrushModes'` error: that one's confusing me, because the script uses `BrushMode` (singular) everywhere — the only `s` near there is on `brush_data_list`. Could you double-check that `tools_panel.gd:50` line on your end, in case something got hand-edited locally? A copy of the surrounding 3-4 lines from your file would help me track it down. The line number doesn't match `paint_control.BrushMode.ERASER` in the current source either, so I suspect a local edit or a different branch.


Here's what I have, I can try redownloading the entire demo, I had just copied the code from your pull request into my project so it's possible I had edited something but didn't remember:

```
func button_pressed(button_name):
	# If a brush mode button is pressed.
	var tool_name = null
	var shape_name = null

	if button_name == "mode_pencil":
		paint_control.brush_mode = paint_control.BrushModes.PENCIL
		brush_settings.modulate = Color(1, 1, 1, 1)
		tool_name = "Pencil"
	elif button_name == "mode_eraser":
		paint_control.brush_mode = paint_control.BrushModes.ERASER
		brush_settings.modulate = Color(1, 1, 1, 1)
		tool_name = "Eraser"
	elif button_name == "mode_rectangle":
		paint_control.brush_mode = paint_control.BrushModes.RECTANGLE_SHAPE
		brush_settings.modulate = Color(1, 1, 1, 0.5)
		tool_name = "Rectangle shape"
	elif button_name == "mode_circle":
		paint_control.brush_mode = paint_control.BrushModes.CIRCLE_SHAPE
		brush_settings.modulate = Color(1, 1, 1, 0.5)
		tool_name = "Circle shape"

	# If a brush shape button is pressed
	elif button_name == "shape_rectangle":
		paint_control.brush_shape = paint_control.BrushShapes.RECTANGLE
		shape_name = "Rectangle"
	elif button_name == "shape_circle":
		paint_control.brush_shape = paint_control.BrushShapes.CIRCLE
		shape_name = "Circle";

	# If a opperation button is pressed
	elif button_name == "clear_picture":
		paint_control.brush_data_list = []
		paint_control.queue_redraw()
	elif button_name == "save_picture":
		save_dialog.popup_centered()
	elif button_name == "undo_stroke":
		paint_control.undo_stroke()

	# Update the labels (in case the brush mode or brush shape has changed).
	if tool_name != null:
		label_tools.text = "Selected tool: " + tool_name
	if shape_name != null:
		label_brush_shape.text = "Brush shape: " + shape_name
```

AThousandShips:

PR #1321, opened 2026-05-04
Reinitialize water_plane compute resources on scene re-entry

### LINKED ISSUE #1152: Compute Texture Demo is Brittle + Proposed Fix
water_plane

Godot v4.4.dev.mono (e9504b662) - Windows 10.0.19045 - Multi-window, 1 monitor - Vulkan (Forward+) - dedicated NVIDIA GeForce RTX 3070 Laptop GPU (NVIDIA; 31.0.15.3742) - 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz (16 threads)

**Issue description:**
The demo code is too brittle to use as reference material. A scene change or scene reload cleans up the texture but does not rebind it. If you add another blank scene to the demo, open the main scene and look at the pretty texture, change tab to the new scene then reopen the main scene the texture will be blank and you wont see any ripples.

I traced the issue to the Texture2DRD not surviving the scene change/reload event. You can fix it by doing the following:

```
func _ready():
	next_texture = 0;
	t = 0.0
... Replace the get texture with:
			# Set our texture object.
			material.set_shader_parameter("effect_texture", texture);
```

At the end of the exit tree request ready refire so that the scene will rebuild the freed textures:
```
func _exit_tree():
...
	request_ready()
```

Finally make the new Texture2DRD object in _initialize_compute_code:
```
func _initialize_compute_code(init_with_texture_size):
	# As this becomes part of our normal frame rendering,
	# we use our main rendering device here.
	rd = RenderingServer.get_rendering_device()
	
	# Create a new texture
	texture = Texture2DRD.new()
```

With these changes I can smoothly reload a scene and not have the compute shader fail.

Another QoL side note, include RenderingDevice.TEXTURE_USAGE_CAN_COPY_FROM_BIT on the tf usage bits so you can preview the texture in the editor.

adamcodeworks:
Fixes #1152.

## What's wrong

Switching the scene tab away from the water plane and back (or reloading the scene at runtime) left the plane blank and printed errors. Three things were going wrong:

1. **Init only ran once.** The compute resources and the `Texture2DRD` binding were set up in `_ready`, which only fires the first time a node enters the tree. `_exit_tree` was already clearing `texture_rd_rid` and freeing the compute resources, but on re-entry nothing rebuilt them.
2. **The scene's `Texture2DRD` SubResource doesn't survive an unload reliably.** Picking it back up via `material.get_shader_parameter(&"effect_texture")` on re-entry returned a `Texture2DRD` whose underlying RD texture had been torn down.
3. **Main-thread / render-thread race.** `_initialize_compute_code` is queued onto the render thread via `call_on_render_thread`. Between `_enter_tree` returning and the render thread actually running, `_process` kept running on the main thread and bound `texture_rds[next_texture]` — still holding the *previous* lifecycle's now-freed RIDs — onto the texture. With 9 uniform sets backed by those RIDs you get exactly the 9 errors `@Calinou` reported on 4.7.beta2.

## Fix

In `water_plane.gd`:

* Move per-instance setup from `_ready` into `_enter_tree` so it runs every tree entry (matches what `_exit_tree` was already pairing with, and the approach `@BastiaanOlij` suggested in the issue).
* Create a fresh `Texture2DRD` in `_enter_tree` and bind it onto the material with `set_shader_parameter`, instead of reading whatever the scene hands back. The script owns the `Texture2DRD`'s lifetime now, so reloads can't leave it stranded.
* Add a `ready_for_render` flag set on the render thread at the end of `_initialize_compute_code` and cleared on the main thread in `_exit_tree`. `_process` bails out before binding/dispatching while it's false, so we no longer bind freed RIDs during the init window.
* Reset `texture_rds` / `texture_sets` / `shader` / `pipeline` at the end of `_free_compute_resources` so a stale read from the main thread can never observe freed RIDs.

No changes to the GLSL, `.tscn`, or any other file.

## Verification

I wrote a small standalone scene mirroring the lifecycle pattern (init in `_ready` vs. init in `_enter_tree`) and exercised it through `add_child` / `remove_child`:

```
=== buggy variant (init in _ready) ===
after first add: initialized=1
after re-add:    initialized=1   <-- bug

=== fixed variant (init in _enter_tree) ===
after first add: initialized=1
after re-add:    initialized=2   <-- fixed
```

I haven't loaded the patched scene in the editor to confirm the 9 errors are gone and the ripples come back visually after a tab switch on macOS — please re-test before merging.

### DIFF
--- a/compute/texture/water_plane/water_plane.gd
+++ b/compute/texture/water_plane/water_plane.gd
@@ -32,8 +32,10 @@ var add_wave_point: Vector4
 var mouse_pos: Vector2
 var mouse_pressed: bool = false
 
-# Called when the node enters the scene tree for the first time.
-func _ready() -> void:
+# Use _enter_tree so the compute resources and texture binding are recreated
+# whenever the node re-enters the scene tree (e.g. switching scene tabs in the
+# editor or a runtime scene reload). _ready only fires on the first entry.
+func _enter_tree() -> void:
 	# In case we're running stuff on the rendering thread
 	# we need to do our initialisation on that thread.
 	RenderingServer.call_on_render_thread(_initialize_compute_code.bind(texture_size))
--- a/compute/texture/water_plane/water_plane.gd
+++ b/compute/texture/water_plane/water_plane.gd
@@ -32,27 +32,41 @@ var add_wave_point: Vector4
 var mouse_pos: Vector2
 var mouse_pressed: bool = false
 
+# Set on the render thread once compute resources have been (re)created. The
+# main thread checks this in _process so it doesn't bind freed RIDs in the
+# window between scene re-entry and the render thread finishing init.
+var ready_for_render: bool = false
+
 # Use _enter_tree so the compute resources and texture binding are recreated
 # whenever the node re-enters the scene tree (e.g. switching scene tabs in the
 # editor or a runtime scene reload). _ready only fires on the first entry.
 func _enter_tree() -> void:
+	next_texture = 0
+
+	# Create our own Texture2DRD and bind it to the material rather than
+	# reading the SubResource declared in the scene. The SubResource doesn't
+	# survive a scene unload reliably, so owning the lifetime here keeps the
+	# binding consistent across reloads.
+	texture = Texture2DRD.new()
+
+	var material: ShaderMaterial = $MeshInstance3D.material_override
+	if material:
+		material.set_shader_parameter(&"effect_texture", texture)
+		material.set_shader_parameter(&"effect_texture_size", texture_size)
+
 	# In case we're running stuff on the rendering thread
 	# we need to do our initialisation on that thread.
 	RenderingServer.call_on_render_thread(_initialize_compute_code.bind(texture_size))
 
-	# Get our texture from our material so we set our RID.
-	var material: ShaderMaterial = $MeshInstance3D.material_override
-	if material:
-		material.set_shader_parameter(&"effect_texture_size", texture_size)
-
-		# Get our texture object.
-		texture = material.get_shader_parameter(&"effect_texture")
-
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Haven't tested this but the fix seems sound. I'm a little surprised by the problem with switching tabs, I guess there was a recent change where we cache the node tree of each tab loaded? Before I'm pretty sure the tree was reconstructed.

All in all, this feels like a safe fix to do.

Calinou (approved):
.

Calinou:
Tested locally on macOS (both Metal and Vulkan), this doesn't appear to fix the issue on 4.7.beta2. Switching scenes back and forth will still cause 9 errors to be printed and the water texture will no longer update in the editor.

adamcodeworks:
Thanks for re-testing @Calinou — pushed a follow-up that should address the 9 errors and the frozen texture on 4.7.beta2.

Two extra things on top of the original `_enter_tree` rename:

* The `Texture2DRD` SubResource from the scene file wasn't surviving the unload reliably, so the script now creates a fresh `Texture2DRD` in `_enter_tree` and binds it onto the material itself instead of reading whatever the scene hands back.
* `_initialize_compute_code` runs on the render thread, but `_process` keeps running on the main thread in the meantime — so it was binding `texture_rds[next_texture]` from the previous lifecycle (already-freed RIDs) for a few frames after re-entry. Added a `ready_for_render` flag that the render thread sets after init and cleared on `_exit_tree`; `_process` returns early until then. Also reset the RID arrays at the end of `_free_compute_resources` for good measure.

PR body updated with the full description. I haven't been able to retest on macOS — could you give it another spin when you have a moment?

AThousandShips:

AThousandShips:

PR #1324, opened 2026-05-15
Fix relative motion for paired touch/mouse events in GUI in 3D

### LINKED ISSUE #1108: Fix motion & drag event in 'GUI in 3D'
GUI in 3D

**Issue description:**
The variables `last_event_pos2D` and `last_event_time` will be assigned repeatedly in same frame, if there are multiple events triggered in one frame. 
It leads to wrong event data for `InputEventMouseMotion` and `InputEventScreenDrag` which need last frame's info. 
These variables should be assigned at most once per frame. 

**Step to reproduce:**
1. Set project settings `input_devices/pointing/emulate_touch_from_mouse` to `true`
2. Add a ScrollContainer and some scrollable Nodes to the main scene. 
3. The ScrollContainer does not response to mouse dragging in 3D scene. 

**Modification should fix it:**
``` diff
...

+ var record_frame = 0
+ var record_pos2D : Vector2
+ var record_event_time : float

func _mouse_input_event(_camera: Camera3D, event: InputEvent, event_position: Vector3, _normal: Vector3, _shape_idx: int) -> void:
	# Get mesh size to detect edges and make conversions. This code only supports PlaneMesh and QuadMesh.
	var quad_mesh_size: Vector2 = node_quad.mesh.size

	# Event position in Area3D in world coordinate space.
	var event_pos3D := event_position

	# Current time in seconds since engine start.
	var now := Time.get_ticks_msec() / 1000.0

	# Convert position to a coordinate space relative to the Area3D node.
	# NOTE: `affine_inverse()` accounts for the Area3D node's scale, rotation, and position in the scene!
	event_pos3D = node_quad.global_transform.affine_inverse() * event_pos3D

	# TODO: Adapt to bilboard mode or avoid completely.

	var event_pos2D := Vector2()

	if is_mouse_inside:
		# Convert the relative event position from 3D to 2D.
		event_pos2D = Vector2(event_pos3D.x, -event_pos3D.y)

		# Right now the event position's range is the following: (-quad_size/2) -> (quad_size/2)
		# We need to convert it into the following range: -0.5 -> 0.5
		event_pos2D.x = event_pos2D.x / quad_mesh_size.x
		event_pos2D.y = event_pos2D.y / quad_mesh_size.y
		# Then we need to convert it into the following range: 0 -> 1
		event_pos2D.x += 0.5
		event_pos2D.y += 0.5

		# Finally, we convert the position to the following range: 0 -> viewport.size
		event_pos2D.x *= node_viewport.size.x
		event_pos2D.y *= node_viewport.size.y
		# We need to do these conversions so the event's position is in the viewport's coordinate system.

	elif last_event_pos2D != null:
		# Fall back to the last known event position.
		event_pos2D = last_event_pos2D

+	# Update last frame state (only update once per frame)
+	if record_frame != Engine.get_process_frames():
+		record_frame = Engine.get_process_frames()
+		last_event_pos2D = record_pos2D
+		last_event_time = record_event_time
+		
+		record_pos2D = event_pos2D
+		record_event_time = now
	
	# Set the event's position and global position.
	event.position = event_pos2D
	if event is InputEventMouse:
		event.global_position = event_pos2D

	# Calculate the relative event distance.
	if event is InputEventMouseMotion or event is InputEventScreenDrag:
		# If there is not a stored previous position, then we'll assume there is no relative motion.
		if last_event_pos2D == null:
			event.relative = Vector2(0, 0)
		# If there is a stored previous position, then we'll calculate the relative position by subtracting
		# the previous position from the new position. This will give us the distance the event traveled from prev_pos.
		else:
			event.relative = event_pos2D - last_event_pos2D
			event.velocity = event.relative / (now - last_event_time)

-	## Update last_event_pos2D with the position we just calculated.
-	#last_event_pos2D = event_pos2D
-
-	## Update last_event_time to current time.
-	#last_event_time = now

	# Finally, send the processed input event to the viewport.
	node_viewport.push_input(event)

...
```

adamcodeworks:
Closes #1108.

## What's wrong

When `input_devices/pointing/emulate_touch_from_mouse` is enabled, a single drag produces both an `InputEventMouseMotion` and an `InputEventScreenDrag` in the same physics frame at the same position. The current `_mouse_input_event` refreshes `last_event_pos2D` / `last_event_time` at the bottom of every call, so by the time the second event is processed, `last_event_pos2D` already equals this event's position. The relative motion comes out as `Vector2.ZERO`, which is what stops a `ScrollContainer` inside the SubViewport from receiving scroll deltas. Reproduction is in the issue.

## Fix

Stage every event's position and time into new `record_pos2D` / `record_event_time` members, and only promote them into `last_event_pos2D` / `last_event_time` on the first event of each frame (tracked via `Engine.get_process_frames()` and a new `record_frame` member). Every event in a frame now computes its relative motion against the previous frame's final position, which:

* fixes the paired-event case above (both events compute the same correct delta against the previous frame), and
* gives accurate relatives whenever multiple input events share a frame.

This is the structure the issue reporter proposed, with one small tweak: `record_pos2D` is updated on every event (not just the first of the frame), so the next frame's "previous-frame position" is the *last* event from this frame, not the first.

Drive-by: the existing `if last_event_pos2D == null` guard was always true — `last_event_pos2D` is a `Vector2` and can never be null. Replaced it with `if last_event_time < 0.0`, matching the existing sentinel value, so the first event after node spawn correctly reports zero relative motion.

## Verification

I haven't booted the SubViewport scene in the editor to confirm scroll wheel deltas, please double-check there before merging. The logic change is small and self-contained.

### DIFF
--- a/viewport/gui_in_3d/gui_3d.gd
+++ b/viewport/gui_in_3d/gui_3d.gd
@@ -4,11 +4,31 @@ extends Node3D
 ## Used for checking if the mouse is inside the Area3D.
 var is_mouse_inside: bool = false
 
-## The last processed input touch/mouse event. Used to calculate relative movement.
+## The position of the last input event from the *previous* frame, used to
+## calculate relative movement for [InputEventMouseMotion] and
+## [InputEventScreenDrag] events. See [member record_pos2D] for why this is
+## not updated inside the same frame.
 var last_event_pos2D := Vector2()
 
-## The time of the last event in seconds since engine start.
+## The time of the last input event from the previous frame, in seconds since
+## engine start.
 var last_event_time := -1.0
+
+## The frame in which the last update of [member last_event_pos2D] happened.
+## Used to ensure [member last_event_pos2D] / [member last_event_time] are only
+## refreshed once per frame, so events sharing a frame (e.g. a paired
+## [InputEventMouseMotion] / [InputEventScreenDrag] when
+## `input_devices/pointing/emulate_touch_from_mouse` is enabled) compute their
+## relative motion against the previous frame, not against each other.
+var record_frame: int = -1
+
+## The position of the most recent input event in the current frame. Promoted
+## into [member last_event_pos2D] on the next frame's first event.
+var record_pos2D := Vector2()
+
+## The time of the most recent input event in the current frame. Promoted
+## into [member last_event_time] on the next frame's first event.
+var record_event_time: float = -1.0
 
 @onready var node_viewport: SubViewport = $SubViewport
 @onready var node_quad: MeshInstance3D = $Quad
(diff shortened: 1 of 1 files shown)


AThousandShips:
> I haven't booted the SubViewport scene in the editor to confirm scroll wheel deltas, please double-check there before merging

What is preventing you from doing this when verifying your changes?

Also are you using any AI tools in any part of this PR? Please see our [contribution guidelines](https://contributing.godotengine.org/en/latest/organization/general_rules_and_guidelines.html#ai-assisted-contributions)

AThousandShips:

PR #1327, opened 2026-06-10
Add signal connections for note events

hvdyinv4688hbv:


Calinou (approved):

Thanks!

PR #1329, opened 2026-06-20
Squash the Creeps: Update Mob.gd, explicit floats

samw-26:
Explicitly use floats for min and max speed for clarity.

Documentation pr: https://github.com/godotengine/godot-docs/pull/12075

### DIFF
--- a/3d/squash_the_creeps/Mob.gd
+++ b/3d/squash_the_creeps/Mob.gd
@@ -4,9 +4,9 @@ extends CharacterBody3D
 signal squashed
 
 ## Minimum speed of the mob in meters per second.
-@export var min_speed = 10
+@export var min_speed = 10.0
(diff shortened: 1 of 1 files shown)


aaronfranke:
In the demos, this is already floating-point division, because it gets assigned from `randf_range`:

```gd
var random_speed = randf_range(min_speed, max_speed)
```

We should probably change `randi_range` to `randf_range` in the docs. Then in both the docs and the demos, change the export variables to floats by adding a `.0` suffix:

```gd
## Minimum speed of the mob in meters per second.
@export var min_speed = 10.0
## Maximum speed of the mob in meters per second.
@export var max_speed = 18.0
```

aaronfranke (approved):

Calinou (approved):
Looks good to me.

PR #1339, opened 2026-07-24
[WebSocket Multiplayer] Use PROTO_NAME constant in _init.

### LINKED ISSUE #652: WebSocket Multiplayer Demo Mini-"Bug"

WebSocket Multiplayer Demo
https://godotengine.org/asset-library/asset/540

Irrelevant - it's an easy fix

**Issue description:**

In Script main.gd in this project, the constant for the Protocol Name is defined in line 04 thus:
`	const PROTO_NAME = "ludus"`

This is properly used in the code later down, but not in line 73. where instead the content of PROTO_NAME is typed out again. So, Line 73 should read:
`	host.listen(DEF_PORT, PoolStringArray([PROTO_NAME]), true)`
instead of 
`	host.listen(DEF_PORT, PoolStringArray(["ludus"]), true) `

The reason why this matters is that when users presume the code to be fine they might just change the constant at the top (which I did) only to find that if you rename it the connections no longer work for apparently no reason.

That is all.

TAB5432:
Replaces hardcoded "ludus" string with PROTO_NAME constant in main.gd line 17 so changing protocol definition at top of script works properly. Fixes #652.


aaronfranke (approved):

Calinou (approved):
Thanks!

PR #1341, opened 2026-07-28
Link to Asset Store in Custom Splash Screen demo README

Calinou:
The demo is being reviewed on the asset store currently, so wait until it's approved before merging this.


Calinou:
The demo is approved now: https://store.godotengine.org/asset/godot-foundation/custom-splash-screen-demo/

PR #1347, opened 2026-08-25
Fix truck_town wheel radius

Schnorg:
The wheel radius in truck town is inconsistent amongst the trucks, even though sharing the same wheel model.

This needs to be fixed in order to test and confirm fixes like this: [Fix VehicleWheel3D sinking - #112440](https://github.com/godotengine/godot/pull/112440)

I looked at the wheel model in blender. I am no expert with the measuring tool using my track-pad, but the radius looks like 0.23, so putting that everywhere seemed to have better results (some wheels were set to 0.25, some 0.18, and some just broken at 0.1).


Calinou (commented):
Tested locally with all vehicles:

Wheels still sink slightly into the ground, maybe a radius of 0.24 or 0.25 would be more suitable:

Image

I do think we should keep the collision size ever so slightly smaller than the wheel's appearance, as otherwise the wheels seem "solid" and not like actual tires with air inside of them.

The truck's back wheels are still sinking a fair bit into the ground too:

Image

The same goes for the tow truck, to a less extreme extent: 

Image

Schnorg:
That looks like the existing bug in VehicleBody3D.cpp. Did you test it with PR 112440: [Fix VehicleWheel3D sinking - #112440](https://github.com/godotengine/godot/pull/112440)

The idea of this PR is to make all the wheels relatively correct so that existing bugs (and their fixes) can be more clearly seen, to evaluate #112440.

This is what I am seeing with @skorpnok's godot engine PR and this PR (with 0.23):

image
imageimage

PR #1351, opened 2026-09-08
Fix main scene in custom_logging demo

jmdejong:
In the custom_logging demo use main.tscn as the main scene instead of custom_logger_ui.tscn

custom_logger_ui.tscn is not meant to be ran as main scene and will not demo the functionality by itself.


Calinou (approved):

Thanks!

PR #1354, opened 2026-09-19
Add gyro, LED and touchpad support to the Joypads demo

Calinou:
- Improve axis value display.

Controller model source: [controller.blend.zip](https://github.com/user-attachments/files/32404780/controller.blend.zip) (made by yours truly :slightly_smiling_face:)

There seems to be an issue with the rotation appearing to drift when you wiggle back and forth on different axes; I'm not sure if it's gyroscope drifting. It may be related to the rotation order, but I've tried to change it on the Controller node to no avail.

@Nintorch ~~On 4.8.dev a46e3a103 (Linux, built-in SDL), I noticed the reported number of touchpads is wrong if you disconnect a controller and switch IDs later on. The lone number is the reported number of touchpads here:~~

Screenshot_20260919_015130

~~This causes the touchpads panel to remain visible if you disconnect a controller that supports touchpads, then connect a controller that doesn't support touchpads until you restart the project.~~

**Edit:** Fixed by this engine PR:

- https://github.com/godotengine/godot/pull/123672

### Controller with gyro/LED/touchpad support

joypads_with_gyro

### Controller without gyro/LED/touchpad support

joypads_without_gyro


Nintorch (approved):
Tested on Windows 11 with my DualShock 4 connected either via a cable or an 8BitDo adapter in XInput and Nintendo Switch modes, everything works as expected (excluding the bug you described), thanks!
> On 4.8.dev a46e3a103 (Linux, built-in SDL), I noticed the reported number of touchpads is wrong if you disconnect a controller and switch IDs later on.

I think this is happening because I forgot to add a call to `joy_touch.erase` here when a controller gets disconnected, therefore the information about the previous joypad touchpads doesn't get cleared:
https://github.com/godotengine/godot/blob/57277407e77e61b161f35dbd7aeb510f7a9e26a6/core/input/input.cpp#L743-L757
```c++
	} else {
		js.connected = false;
		for (int i = 0; i < (int)JoyButton::MAX; i++) {
			JoyButton c = _combine_device((JoyButton)i, p_idx);
			joy_buttons_pressed.erase(c);
		}
		for (int i = 0; i < (int)JoyAxis::MAX; i++) {
			set_joy_axis(p_idx, (JoyAxis)i, 0.0f);
		}
		MotionInfo *motion = joy_motion.getptr(p_idx);
		if (motion != nullptr && motion->gamepad_motion != nullptr) {
			delete motion->gamepad_motion;
		}
		joy_motion.erase(p_idx);
	}
```
I think just adding `joy_touch.erase(p_idx);` should work, I will try to do that today!

misc/joypads/joypads.gd
```diff
@@ -87,6 +95,48 @@ func _process(_delta: float) -> void:
...
+		# Only show fingers that are currently being pressed.
+		finger_sprite.visible = false
+
+	for finger: int in Input.get_joy_touchpad_fingers(cur_joy):
+		var finger_sprite: Sprite2D = $Touchpad/Fingers.get_node(str(finger))
+		finger_sprite.visible = true
+		finger_sprite.position = Vector2(
+				remap(Input.get_joy_touchpad_finger_position(cur_joy, finger).x, 0.0, 1.0, 0.0, $Touchpad.size.x),
+				remap(Input.get_joy_touchpad_finger_position(cur_joy, finger).y, 0.0, 1.0, 0.0, $Touchpad.size.y),
+		)
```
AThousandShips:
```suggestion
			)
```

Calinou:
https://github.com/godotengine/godot/pull/123672 fixes the issue reported in OP:

image

